PR #4, opened 2024-05-22
Reorganize files into a flat structure

shainaraskas:
This PR reorganizes our files into a flat structure, and updates xrefs so all images and partials load as intended.

The flat structure is intended to:
- allow us to play with information architecture in future while minimizing file moves in the backend
- allow us to share certain resources that require relative paths easily
- adhere to the versioning structure we expect we will use (backporting TODO)

If we decide that this structure isn’t granular enough to navigate for our contributors, we can add additional subfolders later.

```
serverless      # the deployment type
├── pages       # mdx pages
├── images      # you get it
├── nav         # nav files for serverless
└── partials    # mdx included in other pages    
```

nb: the partials we currently have are only used in a single place. in my opinion, these could be integrated into the parent page.


docs/shared-features/project-settings/images/es-query-rule-action-summary.png
shainaraskas, line 1: duplicate

docs/shared-features/project-settings/images/es-query-rule-recovery-action.png
shainaraskas, line 1: duplicate

serverless/partials/generate-embeddings-widget.mdx
shainaraskas, line 1: not sure why this is detecting as a new file, but it's not letting me fix it

PR #51, opened 2024-08-22
Fix API feedback template

lcawl:
This PR fixes the following error that was occurring for the API feedback template: 

```
There is a problem with this template
body[1]: options must not include booleans. Please wrap values such as 'yes', and 'true' in quotes.
```
[Learn more about this error.](https://docs.github.com/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms#bodyi-options-must-not-include-booleans-please-wrap-values-such-as-yes-and-true-in-quotes)


kilfoyle (approved):
LGTM! 🐇

Issue #146, opened 2024-10-25
[Serverless]: Serverless Developer Tools -> Links result in error


maggieghamry:
### Serverless Docs

Welcome to Elastic Serverless

### Description

On the Serverless / Developer tools page, there are links under the `Available in` column, all which link to an empty page with an error:![Image](https://github.com/user-attachments/assets/4c0dc963-3918-4eb7-916d-c9724925759e)

Clicking on any of these links results in an error:
![Image](https://github.com/user-attachments/assets/eea7d584-f13a-4b4f-b287-79d296c58acd)

Additionally, clicking on the same icons in all subsequent `Dev tools` links results in the same errors
[Console](https://www.elastic.co/docs/current/serverless/devtools/run-api-requests-in-the-console)
[Search Profiler](https://www.elastic.co/docs/current/serverless/devtools/profile-queries-and-aggregations)
[Grok Debugger](https://www.elastic.co/docs/current/serverless/devtools/debug-grok-expressions)
[Painless Lab](https://www.elastic.co/docs/current/serverless/devtools/debug-painless-scripts)
[Troubleshooting](https://www.elastic.co/docs/current/serverless/devtools/dev-tools-troubleshooting)
![Image](https://github.com/user-attachments/assets/696d5772-f3ae-4592-83c7-87919e7a2186)

### Resources and additional context

Each of these links shows as active and hyperlinked, and should either be disabled or resolved to navigate to the corresponding information ![Image](https://github.com/user-attachments/assets/8ab42bff-78aa-4700-b36e-ff98f3b69f8b)
each of these icons leads to:
![Image](https://github.com/user-attachments/assets/d8caf5c9-e265-4345-bfc9-06d5b17d92b9)

leemthompo:
I believe they're all missing the `current/` prefix in the URL

maggieghamry:
Yes, it looks like it's removing `current/` in front of `serverless/`

georgewallace:
after migration to asciidoc these badges now work as expected and no longer return a 404. cc: @leemthompo @KOTungseth

Issue #88, opened 2024-10-21
[Serverless]: Project settings: Broken links, typos, and editorial stuff on project management docs main page


timto-elastic:
### Serverless Docs

Project and management settings

### Description

The solution links for observability, search, and security badged items on https://www.elastic.co/docs/current/serverless/project-and-management-settings are broken. 

Broken links: 

In main text:
https://www.elastic.co/docs/serverless/elasticsearch/what-is-elasticsearch-serverless
https://www.elastic.co/docs/serverless/observability/what-is-observability-serverless
https://www.elastic.co/docs/serverless/security/what-is-security-serverless

Typo: the link text for [serverless Elasticsearch docs](https://www.elastic.co/docs/current/serverless/elasticsearch/what-is-elasticsearch-serverless) should be Elasticsearch serverless docs

Editorial:
Why does there need to be an admonition that has links that go to the same content in the main body text? Also I can't tell that those target pages (which are the what is pages) tell me about how project settings are different across solutions or if I am using a solution where I would look. Should we be just showing a list here?

Also Integrations and fleet/agent are also in this bucket. Should we be talking about what those are here? We are inferring project management here I think but not the broader category of things in this bucket and what they are for. Also connector clients are in the search docs and  but not listed on the integrations page (https://www.elastic.co/docs/current/serverless/integrations). I would assume we want to link to 
https://www.elastic.co/docs/current/serverless/elasticsearch/ingest-data-through-integrations-connector-client since we call the connectors "integrations" in search.

### Resources and additional context

https://www.elastic.co/docs/current/serverless/project-and-management-settings
![Image](https://github.com/user-attachments/assets/06507458-af22-4b26-82b7-b51c368c4e5e)

leemthompo:
https://github.com/elastic/docs-content/pull/149 addresses:

- Broken links
- Typo
- Editorial (by deleting unnecessary admonition)

bmorelli25:
Fixed by migration.

PR #149, opened 2024-10-28
Delete solution badges

leemthompo:
Addresses https://github.com/elastic/docs-content/issues/88
Closes https://github.com/elastic/docs-content/issues/146

We don't have link checking on these badges and the URLs might change again in the future, silently failing, so best to just remove solution badges entirely.

It's less pretty but `¯\_(ツ)_/¯`

### **Screenshot**:

Screenshot 2024-10-29 at 12 03 14

### DIFF
--- a/serverless/pages/action-connectors.mdx
+++ b/serverless/pages/action-connectors.mdx
@@ -6,7 +6,7 @@ tags: [ 'serverless' ]
 ---
 
 <DocBadge template="technical preview" />
-This content applies to: <DocBadge template="elasticsearch" slug="/serverless/elasticsearch/what-is-elasticsearch-serverless" /> <DocBadge template="observability" slug="/serverless/observability/what-is-observability-serverless" /> <DocBadge template="security" slug="/serverless/security/what-is-security-serverless" />
+This content applies to: <DocBadge template="elasticsearch" /> <DocBadge template="observability" /> <DocBadge template="security" />
 
 The list of available connectors varies by project type.
 
--- a/serverless/pages/action-connectors.mdx
+++ b/serverless/pages/action-connectors.mdx
@@ -6,8 +6,11 @@ tags: [ 'serverless' ]
 ---
 
 <DocBadge template="technical preview" />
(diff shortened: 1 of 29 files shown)


elasticdocs:
:rocket: Built _elastic-dot-co-docs-preview-docs_ successfully!

* https://elastic-dot-co-docs-production-9ju4o8c8k-elastic-dev.vercel.app

* Built with commit b18414da0d4c76c7df21931ea93f9cc2ba13801d

> Issues? Visit #next-docs in Slack

bmorelli25:
It looks like the badges are still trying to link out. Possibly because of [the template](https://github.com/elastic/docsmobile/blob/main/docsmobile/template/src/components/docBadge/config/templates.ts)?

leemthompo:
Yeah need to remove linking entirely from badges methinks.

Related PR: https://github.com/elastic/docsmobile/pull/946

timto-elastic:
Do we need badges at all?

leemthompo:
> Do we need badges at all?

Good question, they're a little neater than plain text. But in terms of opportunity cost of our time here, @bmorelli25 if removing the logic from the docsmobile code is too heavy a lift, we could just find and replace all badges with plaintext I guess.

bmorelli25:
I agree they're neater than plaintext. With that being said, I looked at https://github.com/elastic/docsmobile/pull/946 and this seems like more work than it's worth, especially for a system that we're hoping to migrate away from soon.

Sorry to have you change this PR, but I think it makes more sense to do away with our use of these entirely. That has the added benefit of simplifying migration.

leemthompo:
I've removed 'em entirely and updated the PR description to represent this new tack

bmorelli25 (approved):
Preview LGTM!

florent-leborgne (approved):

leemthompo:
@bmorelli25 @colleenmcginnis I don't know if you need this PR merged in light of re-migration, so I'll let y'all merge if needed. :)

bmorelli25:
Thanks for your work on this! In light of re-migration, I will close this pull request. Your effort is nonetheless appreciated.

PR #182, opened 2024-11-07
Fix stateful vs serverless attribute conflicts

colleenmcginnis:
Removes attribute `include`s in `index*.asciidoc` files that are subsections of the new Serverless book. These files will inherit attributes from [`serverless/index.asciidoc`](https://github.com/elastic/docs-content/blob/main/serverless/index.asciidoc?plain=1).

h/t @leemthompo for pointing out incorrect attribute values on rendered pages!


colleenmcginnis:
run docs-build

colleenmcginnis:
For the reviewer: You can see the result of this change by going to [this diff](https://github.com/elastic/built-docs/compare/master...docs-content_bk_182) in the built-docs repo or looking at [diff page](https://docs-content_bk_182.docs-preview.app.elstc.co/diff) and comparing the preview link to the production link:

https://github.com/user-attachments/assets/3580ac0d-ebb9-4e71-a275-98f6c66dba6f

serverless/index-serverless-devtools.asciidoc
```diff
@@ -1,8 +1,3 @@
-:doctype: book
-
-include::{asciidoc-dir}/../../shared/versions/stack/master.asciidoc[]
-include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
-
 [[devtools-developer-tools]]
```
colleenmcginnis, line 1:
```suggestion
// Attributes used in files included here come from
// https://github.com/elastic/docs-content/blob/main/serverless/index.asciidoc?plain=1#L3-L37

[[devtools-developer-tools]]
```

serverless/index-serverless-elasticsearch.asciidoc
```diff
@@ -1,8 +1,3 @@
-:doctype: book
-
-include::{asciidoc-dir}/../../shared/versions/stack/master.asciidoc[]
-include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
-
 [[what-is-elasticsearch-serverless]]
```
colleenmcginnis, line 1:
```suggestion
// Attributes used in files included here come from
// https://github.com/elastic/docs-content/blob/main/serverless/index.asciidoc?plain=1#L3-L37

[[what-is-elasticsearch-serverless]]
```

serverless/index-serverless-general.asciidoc
```diff
@@ -1,8 +1,3 @@
-:doctype: book
-
-include::{asciidoc-dir}/../../shared/versions/stack/master.asciidoc[]
-include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
-
 [[intro]]
```
colleenmcginnis, line 1:
```suggestion
// Attributes used in files included here come from
// https://github.com/elastic/docs-content/blob/main/serverless/index.asciidoc?plain=1#L3-L37

[[intro]]
```

serverless/index-serverless-project-settings.asciidoc
```diff
@@ -1,8 +1,3 @@
-:doctype: book
-
-include::{asciidoc-dir}/../../shared/versions/stack/master.asciidoc[]
-include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
-
 [[project-and-management-settings]]
```
colleenmcginnis, line 1:
```suggestion
// Attributes used in files included here come from
// https://github.com/elastic/docs-content/blob/main/serverless/index.asciidoc?plain=1#L3-L37

[[project-and-management-settings]]
```

colleenmcginnis (commented):
If the comment below would be a helpful reminder, we can add it to each index file.

leemthompo:
@colleenmcginnis there'll be just a couple of smelly ones with this change, see [this grep](https://github.com/search?q=repo%3Aelastic%2Fdocs-content+%22serverless+%7Bes%7D%22&type=code) I wonder if we need to still have a bare attribute for simply "Elasticsearch"?

leemthompo:
But could also just be cleaned up in follow up by ES writers!

colleenmcginnis:
It's up to you and your team. If you want separate attributes for serverless and stateful, you could create two unique keys with unique values and update throughout the docs as you see fit. 

Note: This is not unique to the move to AsciiDoc. There was only [one value for the `es` variable in the docs-site repo](https://github.com/elastic/docs-site/blob/main/config/variables/terms.js#L13), too, so it was probably rendering strangely in the Docsmobile docs, too. 🤷

leemthompo:
I need to look at this again tomorrow because too late for my feeble brain 😵‍💫. But I fear it might be a cognitive ouchy to have the same attribute do different things whether we're working in the serverless book or not, but yeah it's not a huge deal anyways. We had `es-serverless` but I don't think we were consistent one way or another.

lcawl:
With respect to the attributes that we're overloading in https://github.com/elastic/docs-content/blob/7602ddbdb3de8fe99b0850423c308f85e2952d6d/serverless/index.asciidoc?plain=1#L31

1) `serverless-full` and `serverless-short` is just a capitalization difference, so I think we should just synch on one and define it only in the docs repo
2) Likewise I think we should find out which of the "Elasticsearch Serverless" or "Elasticsearch on serverless" is preferred and sync on that for `es-serverless`. If there's really a necessary difference here, IMO we'd add a new attribute for it in the docs repo.
3) For the overloaded `es` attribute, of which there are 202 occurrences in the docs-content repo, I think we should review those occurrences and change them to `es-serverless` if appropriate then remove this override.
3) For the overloaded `observability` and `security` attributes, I'd recommend creating `obs-serverless` and `sec-serverless` (or some such variation) in the shared attributes in the docs repo, then using them appropriately. There were 6 occurrences of `{observability}` and 0 occurrences of `{security}` in docs-content, but I didn't dig into the other repos.

colleenmcginnis:
Do we want the ability to use the same source content for both stateful and serverless docs when the only difference between the content is the product names? If yes, I think it makes sense to have one attribute key that has two values depending on the context.

(An aside: If we were to consolidate serverless and stateful docs into a single file, which product name would we use?)

lcawl:
> Do we want the ability to use the same source content for both stateful and serverless docs when the only difference between the content is the product names? 

I would hope we don't have any pages where that's true, since if there aren't any differences other than the product names it's not worth having two pages IMO. Maybe this exists in the short-term book-based context but ideally not in the long term.

> (An aside: If we were to consolidate serverless and stateful docs into a single page, which product name would we use?)

I think this is the kind of exercise we need to try with real examples, since IMO you could go for the most generic names or most specific names depending on the context.

For something very high-level and conceptual like https://www.elastic.co/guide/en/kibana/current/alerting-getting-started.html and https://www.elastic.co/guide/en/serverless/current/rules.html, there's no real difference in its behaviour across these contexts, so the most general possible references would work IMO.

For something lower-level like https://www.elastic.co/guide/en/cloud/current/ec-invite-users.html and https://www.elastic.co/guide/en/serverless/current/general-manage-access-to-organization.html, I think we'd use the most specific product/solution/deployment names to call out any differences (though in this case I think they're pretty much identical too).

colleenmcginnis:
>Maybe this exists in the short-term book-based context but ideally not in the long term.

Yea, I guess I was thinking in the short term as we try to prepare for what's next, but are still waiting for v3 to be ready. But, if we don't want to consolidate any source files while we wait, I guess it's worth the effort to look at each instance of `es`, `observability`, and `security` to determine which should be changed to a `*-serverless` alternative. I'll leave it to each team to do that on their section of the docs.

Proposed next steps:

* [ ] I'll open a PR in /docs to add `es-serverless`, `observability-serverless`, and `security-serverless` attributes.
* [ ] @elastic/platform-docs: You can use this PR if you'd like to determine which `es` / `observability` / `security` instances should be changed to the `*-serverless` version throughout all docs-content content.
* [ ] @elastic/obs-docs: I'll use https://github.com/elastic/observability-docs/pull/4487 to determine which `es` / `observability` / `security` instances should be changed to the `*-serverless` version throughout all observability-docs content.
* [ ] @elastic/security-docs: You can open a fresh PR to determine which `es` / `observability` / `security` instances should be changed to the `*-serverless` version throughout all security-docs content.

lcawl:
I've added a commit to remove the redundant variables after https://github.com/elastic/docs/pull/3109 was merged.

leemthompo:
Apologies if I haven't caught up fully here haven't been able to dig in but here are some thoughts

> 1. serverless-full and serverless-short is just a capitalization difference, so I think we should just synch on one and define it only in the docs repo

This isn't a capitalization issues AFAICT it's `Elastic Cloud Serverless` versus `Serverless`.
I don't think we need to capitalize serverless when it's isolated so I think we can potentially remove both of these.

> Likewise I think we should find out which of the "Elasticsearch Serverless" or "Elasticsearch on serverless" is preferred and sync on that for es-serverless. If there's really a necessary difference here, IMO we'd add a new attribute for it in the docs repo.

I think it's "Elasticsearch Serverless" so we should stick to that one IMO.

> For the overloaded es attribute, of which there are 202 occurrences in the docs-content repo, I think we should review those occurrences and change them to es-serverless if appropriate then remove this override.

There are few cases where we just need bare `Elasticsearch` so I'd vote for replacing all `{es}`s with `{es-serverless}`s removing the override, and ES writers can fix up any stragglers over time. We're going to be doing a lot of cleanup there in the next week or so anyways.

leemthompo:
> I'd vote for replacing all {es}s with {es-serverless}s and removing the override, and ES writers can fix up any stragglers over time

I can do this on Monday :)

colleenmcginnis:
Oh gosh this is getting so confusing 😵‍💫 

>>serverless-full and serverless-short is just a capitalization difference, so I think we should just synch on one and define it only in the docs repo
>
>This isn't a capitalization issues AFAICT it's Elastic Cloud Serverless versus Serverless.

I think Lisa was referring to:
* `serverless-full`: 
  * ✅  [`serverless-full` = `Elastic Cloud Serverless`](https://github.com/elastic/docs-site/blob/main/config/variables/terms.js#L7) (Docsmobile pre-migration)
  * ❌  [`serverless-full` = `Elastic Cloud serverless`](https://github.com/elastic/docs/blob/6a83930d88594a79816acef1fb43dcb94d069d35/shared/attributes.asciidoc#L184) (AsciiDoc pre-migration)
* `serverless-short`:
  * ✅  [`serverless-short` = `Serverless`](https://github.com/elastic/docs-site/blob/main/config/variables/terms.js#L8) (Docsmobile pre-migration)
  * ❌  [`serverless-short` = `serverless`](https://github.com/elastic/docs/blob/6a83930d88594a79816acef1fb43dcb94d069d35/shared/attributes.asciidoc#L186) (AsciiDoc pre-migration)

In https://github.com/elastic/docs/pull/3109/files we updated the AsciiDoc values to match the Docsmobile values. 

>I don't think we need to capitalize serverless when it's isolated

It looks like you (@leemthompo) updated the `serverless-short` value to be capitalized a couple weeks ago in https://github.com/elastic/docs-site/pull/49. 😝

leemthompo:
😵‍💫 Here's some more words to add to this salad:

- I don't think we need a variable for the term `serverless` and I also don't think we ever need to capitalize `Serverless` on its own.
- I think "Elastic Cloud Serverless" is the correct capitalization, but I don't like the `serverless-full` variable because it's not immediately clear to an arbitrary contributor what this refers to. Maybe `elastic-serverless` is better.

leemthompo:
> It looks like you  updated the serverless-short value to be capitalized a couple weeks ago

That's water under the bridge. Those were simpler times.

leemthompo (approved):
I think we can all scatter and revisit our use of attributes in our content sets separately?

colleenmcginnis:
Sure. If you don't want to make any other updates here, feel free to merge.

PR #241, opened 2024-12-12
Update Java client docs (there's no serverless version)

swallez:
This PR updates the Java client docs to reflect the fact that there is no dedicated Java client for ES Serverless.

PR #242, opened 2024-12-12
Recommend stack Elasticsearch Python client

pquentin:
Relates https://github.com/elastic/docs-content/pull/241. There's a Python Elasticsearch serverless client (https://github.com/elastic/elasticsearch-serverless-python), but we're planning to deprecate it.

### DIFF
--- a/serverless/pages/clients-python-getting-started.asciidoc
+++ b/serverless/pages/clients-python-getting-started.asciidoc
@@ -4,20 +4,24 @@
 // :description: Set up and use the Python client for {es3}.
 // :keywords: serverless, elasticsearch, python, how to
 
+This page guides you through the installation process of the Python
+client, shows you how to initialize the client, and how to perform basic
+{es} operations with it.
+
+See the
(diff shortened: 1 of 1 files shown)


leemthompo (approved):
LGTM!

leemthompo:
@pquentin sorry I approved a little hastily, perhaps we should add an IMPORTANT note for users who happen to use the existing serverless client that we recommend using the classic client?

pquentin:
We could, however there's near-zero usage of the serverless client today anyway. I was thinking more about adding a deprecation warning in the client itself.

leemthompo:
> We could, however there's near-zero usage of the serverless client today anyway. I was thinking more about adding a deprecation warning in the client itself.

OK that sounds reasonable! In that case should we update the title of this page to remove serverless mention? 

> Get started with the ~serverless~ Python client

pquentin:
Oops, definitely! Thanks, fixed it.

leemthompo (approved):
Thanks!

PR #256, opened 2025-01-09
Fix issues on client subpages before consolidating

marciw:
I'm working on a PR to consolidate the client subpages in the serverless docs and point to the standard/stack clients instead. In the meantime, quick-fixing some issues that crept in with other early PRs in this effort:
- Remove (now inaccurate) serverless banner on the Go subpage
- Remove repository links to avoid conflicts with pointers to standard clients
- Match Java subpage title, mostly for brevity in the left nav


shainaraskas (approved):
LGTM!

PR #298, opened 2025-02-03
Fix incorrectly mapped troubleshooting page

marciw:
Removing a page that shouldn't have been mapped to a new v3 page (instead, it will be used in `index.md` in a many-to-one way, as part of refining/writing)

part of (but does not close) https://github.com/elastic/docs-projects/issues/402


shainaraskas (dismissed):
good to go, but

pls rebase in like 30 min after [this is merged](https://github.com/elastic/docs-content/pull/315) to make sure this isn't breaking links

shainaraskas:
image

shainaraskas (approved):

PR #302, opened 2025-02-04
Test preview

reakaleek:


reakaleek:
@Mpdreamz for some reason the upload-artifact step shows 0s.

But when I watched it "live", the upload took ~40s and the download took ~20s.

Hence, this adds roughly 1 minute. Which is fine IMO for now. We can definitely revisit this if the time grows.

I think we should maybe monitor it somehow.

Mpdreamz:
Aye at least we still fail fast on the build itself.

PR #344, opened 2025-02-05
Update name for 'Elasticsearch Service'

kilfoyle:
According to the last naming update I've seen, `Elasticsearch Service` is now `Elastic Cloud Hosted`. 

The `{{ess}}` attribute has been carried over into the migrated docs, I think it makes sense to continue using this attribute with the new name.

CC @eedugon, @kosabogi, @shainaraskas  since I know you've been working on what were the Cloud docs.

### DIFF
--- a/docset.yml
+++ b/docset.yml
@@ -153,7 +153,7 @@ subs:
   apm-php-ref-v:   "https://www.elastic.co/guide/en/apm/agent/php/{{apm-php-branch}}"
   ecloud:   "Elastic Cloud"
   esf:   "Elastic Serverless Forwarder"
-  ess:   "Elasticsearch Service"
+  ess:   "Elastic Cloud Hosted"
   ece:   "Elastic Cloud Enterprise"
   eck:   "Elastic Cloud on Kubernetes"
   serverless-full:   "Elastic Cloud Serverless"


shainaraskas (approved):
lgtm. we might want to add another one that matches the new acronym (`ech`) for ease of use for contributors as well but can do that separately

shainaraskas (changes requested):
actually, sorry. thinking for one more second about this. we can't do this because ESS used to have only one meaning, but now it's represented by two concepts - Elastic Cloud and Elastic Cloud Hosted

shainaraskas:
I suggest that we add a new dedicated variable and start using it instead to avoid confusion

kilfoyle:
> I suggest that we add a new dedicated variable and start using it instead to avoid confusion

@shainaraskas  Okay. I've added a new variable instead. I guess the existing "ESS" attributes will all need to be reviewed or replaced en masse.

florent-leborgne:
+1 with adding a new variable and "deprecating" the old one. {{ess}} will need to be replaced by either {{ecloud}} or the new {{ech}} in the sources. I think the content refinement we're doing now as part of the migration is a good opportunity to do that accurately. 

With the recent addition of serverless, it might even be wise to check that existing {{ecloud}} occurrences are correct and aren't just meant to designate {{ech}}

florent-leborgne (approved):
Nice to have this as an option now!

shainaraskas (approved):
ty!!

PR #386, opened 2025-02-10
[troubleshoot] Link clean up

colleenmcginnis:
Cleans up links including:

* Fixing links that used `.md` instead of `.html` in the migration process.
* Fixing links that lost prefixes during the sorting process.
* Replacing links to AsciiDoc pages that redirect on the current docs site.


marciw (approved):
whoops, missed this 🐌  LGTM!

PR #393, opened 2025-02-11
update email doc changed

eedugon:
Per conversation with @shainaraskas we are splitting the content into multiple sub-sections as we had multiple procedures all together.

Closes https://github.com/elastic/docs-projects/issues/291


shainaraskas (approved):
perf

PR #426, opened 2025-02-12
[manage data] Add intro to the Elasticsearch data store section

colleenmcginnis:
Closes https://github.com/elastic/docs-projects/issues/368

Adds intro to the Elasticsearch data store section. @kilfoyle @wajihaparvez let me know what you think about this approach!


manage-data/data-store.md
```diff
@@ -3,56 +3,12 @@ mapped_pages:
...
-* **Geospatial search**: Search for locations and calculate spatial relationships using geospatial queries.
-
-**Security**
-
-* **Security information and event management (SIEM)**: Collect, store, and analyze security data from applications, systems, and services.
-* **Endpoint security**: Monitor and analyze endpoint security data.
-* **Threat hunting**: Search and analyze data to detect and respond to security threats.
-
-This is just a sample of search, observability, and security use cases enabled by {{es}}. Refer to Elastic [customer success stories](https://www.elastic.co/customers/success-stories) for concrete examples across a range of industries.
+You can also read more about working with {{es}} as a data store including how to use [aliases](/manage-data/data-store/aliases.md) to point to multiple indices, how to use [index templates](/manage-data/data-store/templates.md) to tell {{es}} how to configure an index when it is created, and how to use the [command line to manage data](/manage-data/data-store/manage-data-from-the-command-line.md) stored in {{es}}.
```
wajihaparvez, line 14: Aliases come after index templates in the side nav right now, so maybe you could switch the two in your sentence here? (Unless we're going to change it in the side nav)

wajihaparvez:
@colleenmcginnis I really like this idea! It's concise and gives the reader all the info they need. I just have one small suggestion but LGTM overall 😄

wajihaparvez (approved):

manage-data/data-store.md
colleenmcginnis: Nice catch! 

```suggestion
You can also read more about working with {{es}} as a data store including how to use [index templates](/manage-data/data-store/templates.md) to tell {{es}} how to configure an index when it is created, how to use [aliases](/manage-data/data-store/aliases.md) to point to multiple indices, and how to use the [command line to manage data](/manage-data/data-store/manage-data-from-the-command-line.md) stored in {{es}}.
```

wajihaparvez (approved):

PR #428, opened 2025-02-12
[manage data] Add an intro to the Data lifecycle section

colleenmcginnis:
Closes https://github.com/elastic/docs-projects/issues/374

This PR adds an intro to the Data lifecycle section. It also moves "Rollup" to the bottom of this section of the table of contents since it is deprecated.

Two questions for @kilfoyle @wajihaparvez:

* Right now https://www.elastic.co/guide/en/serverless/current/project-setting-data.html is mapped to this page, but I'm not sure that content fits here. What do you think?
* What do you think about collapsing the "Why" part of the note about ILM not being available on serverless? I was thinking that the full note could be distracting especially for non-serverless users, but it has helpful info for serverless users.


kilfoyle (dismissed):
LGTM! 🐇 

> Right now https://www.elastic.co/guide/en/serverless/current/project-setting-data.html is mapped to this page, but I'm not sure that content fits here. What do you think?

Yeah, I agree. That content is super general and doesn't belong here in the ILM docs.

> What do you think about collapsing the "Why" part of the note about ILM not being available on serverless? I was thinking that the full note could be distracting especially for non-serverless users, but it has helpful info for serverless users.

I like it! ⚡

colleenmcginnis:
>Right now https://www.elastic.co/guide/en/serverless/current/project-setting-data.html is mapped to this page, but I'm not sure that content fits here. What do you think?

I wasn't sure exactly where to map https://www.elastic.co/guide/en/serverless/current/project-setting-data.html if not here. I looked at how the pages linked from the page in question and it looks like it is mostly linking out to ingestion tools so I thought maybe [Ingest tools overview](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/428/manage-data/ingest/tools) could make sense for a redirect-only mapping.

kilfoyle:
> I wasn't sure exactly where to map https://www.elastic.co/guide/en/serverless/current/project-setting-data.html if not here. I looked at how the pages linked from the page in question and it looks like it is mostly linking out to ingestion tools so I thought maybe [Ingest tools overview](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/428/manage-data/ingest/tools) could make sense for a redirect-only mapping.

Yup! That makes sense to me. Some items aren't related to "Ingest" (the Transforms section, for example), but Ingest seems to capture more than half, so that would be the closest fit.

kilfoyle (approved):
LGTM! 🌵

PR #444, opened 2025-02-13
Add link to the APM OTel guide in the ingest overview

kilfoyle:
@colleenmcginnis In the `otel-documentation` Slack channel we have a request to mention the OTel APM guide somewhere in the ingest overview content. There are many Observability guides so it might not be ideal to link directly to just this one, but then we don't really have mention of OTel elsewhere in the top-level ingestion content, so it does seem worth highlighting.

This would add a link to the "[Ingesting data for Elastic solutions](https://github.com/elastic/docs-content/blob/main/manage-data/ingest/ingesting-data-for-elastic-solutions.md)" page.

Please let me know what you think.

---

Screenshot 2025-02-13 at 11 03 16 AM


colleenmcginnis (approved):
Makes sense to me!

PR #496, opened 2025-02-18
[Docs] Add Ignore missing component templates page

wajihaparvez:
Closes [#372](https://github.com/elastic/docs-projects/issues/372) 
Preview [here]()

### Summary

What I changed:

- Combined content from the two mapped pages and refined it
- Renamed the page to a more user friendly title. Let me know what you think!
- Deleted the linked raw migrated files

I'll delete code comments after the review


manage-data/data-store/templates/ignore-missing-component-templates.md
```diff
@@ -0,0 +1,99 @@
...
+% GitHub issue: docs-projects#372
+
+% Scope notes: Combine with "Usage example" subpage.
+
+% Use migrated content from existing pages that map to this page:
+
+% - [x] ./raw-migrated-files/elasticsearch/elasticsearch-reference/ignore_missing_component_templates.md
+% - [x] ./raw-migrated-files/elasticsearch/elasticsearch-reference/_usage_example.md
+
+When an index template references a component template that might not exist, the `ignore_missing_component_templates` configuration option can be used. With this option, every time a data stream is created based on the index template, the existence of the component template will be checked. If it exists, it will used to form the index’s composite settings. If it is missing, it will be ignored.
```
kilfoyle, line 20:
```suggestion
When an index template references a component template that might not exist, the `ignore_missing_component_templates` configuration option can be used. With this option, every time a data stream is created based on the index template, the existence of the component template will be checked. If it exists, it will used to form the index’s composite settings. If it is missing, it will be ignored.
```

Perfectly good. Just one suggestion: We often aim to address the user directly with active sentences. e.g., "You can use the abc configuration option..." I wouldn't bother changing it in this case, but I noticed it a couple of times in your PRs so just wanted to mention it.

kilfoyle (approved):
LGTM! 🏎️

manage-data/data-store/templates/ignore-missing-component-templates.md
wajihaparvez: Thank you, @kilfoyle! For this sentence specifically, the old page phrased it this way and I wasn't 100% sure that this configuration option is something a user can directly implement. But I'll remember for next time!

PR #490, opened 2025-02-18
Reorganize 'data streams' section

kilfoyle:
As I proposed [here](https://github.com/elastic/docs-projects/issues/324#issuecomment-2625484410), this moves the "Data streams" section out from underneath "Index types" (since I don't think they belong there), and puts it on the same level as "Mapping", "Text analysis", "Templates", etc.

This also removes the "Vector DB" page since that feature isn't yet ready to document.

**Still to do:** 
 - Remove the "Index types" page.
 - Ensure that all parts of the old "Manage indexes" content is still covered somewhere.
 - Add info about Logs DB and TSDB to the main "Data streams" page and/or the "Index basics" page.

See: [docs preview](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/490/manage-data/data-store/data-streams)
Rel: https://github.com/elastic/docs-projects/issues/324

---

screen2

PR #508, opened 2025-02-19
Small fixes to 'data streams' section

kilfoyle:
This is just a small follow-up PR to the earlier https://github.com/elastic/docs-content/pull/490

It:
 - Removes the "Index types" page (all content in that page is already captured in the "index basics" page that Wajiha assembled
 - Moves the "Logs data stream" page to be parallel with "Time series data stream", rather than nested under it (a mistake from my earlier PR).

I also planned to add a note about "data streams" to the "Index basics" page, but I see Wajiha already has that one covered. :-) 

Closes: https://github.com/elastic/docs-projects/issues/324


colleenmcginnis:
Should we add something about data streams to https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/508/manage-data/data-store now that it is at the same level as the other links referenced on this page?

kilfoyle:
> Should we add something about data streams to https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/508/manage-data/data-store now that it is at the same level as the other links referenced on this page?

Good idea @colleenmcginnis 
I've added as shown:

---

Screenshot 2025-02-19 at 12 47 27 PM

wajihaparvez (approved):

PR #488, opened 2025-02-18
[deploy and manage] Add warnings to unfinished deploy and manage pages

colleenmcginnis:
Adds warnings and links to source content to unfinished deploy and manage pages.

PR #512, opened 2025-02-19
More temporary deploy and manage content

colleenmcginnis:
The logic I used to identify the pages in https://github.com/elastic/docs-content/pull/488 didn't account for pages that include code comments AND IDs (like this `$$$some-id$$$`).


shainaraskas (approved):
tyty

PR #495, opened 2025-02-18
Add IDs needed for cross-repo links

colleenmcginnis:
Adds IDs needed for cross-repo links.

PR #533, opened 2025-02-20
[cross-repo links] Add ID that cross-repo links depends on

colleenmcginnis:
Add another ID that cross-repo links depends on like we did in https://github.com/elastic/docs-content/pull/495.


shainaraskas (approved):
sure!

PR #433, opened 2025-02-12
[solutions] Replace AsciiDoc API links

colleenmcginnis:
Replace AsciiDoc API links that should redirect to the new API docs. For more details, refer to https://github.com/elastic/docs-projects/issues/414#issuecomment-2654939177.

PR #431, opened 2025-02-12
[explore and analyze] Replace AsciiDoc API links

colleenmcginnis:
Replace AsciiDoc API links that should redirect to the new API docs. For more details, refer to https://github.com/elastic/docs-projects/issues/414#issuecomment-2654939177.

PR #544, opened 2025-02-20
Rerun API redirect script

colleenmcginnis:
Rerun API redirect script to catch links added/changed since #430 #431 #432 #433 #434 #435.


bmorelli25 (approved):
links tested and working 👍

PR #545, opened 2025-02-20
Add version labels to most 'Manage data' topics

kilfoyle:
For the bulk of the 'Manage data' pages this adds the following labels:

```
applies_to:
  stack: ga
  serverless: ga
```

For the Cloud 'Migrating data' docs I've adjusted to the best of my understanding.

I haven't touched the 'Ingest reference architectures' nor the 'Ingesting data from applications' sections because those might need to be assessed on a case by case basis (I'm really not sure).

Note that these are the same labels that I see @florent-leborgne applying in https://github.com/elastic/docs-content/pull/543/files so I trust it's a really smart and correct thing to do. :-)


wajihaparvez:
Thank you for taking care of this, @kilfoyle! LGTM 🤩

PR #613, opened 2025-02-27
Lock Podman major version

matt-elastic:
## Summary

To prevent automatic Podman major version updates, configure the Podman version to be locked while still allowing minor and patch updates.

### DIFF
--- a/deploy-manage/deploy/cloud-enterprise/configure-host-rhel-cloud.md
+++ b/deploy-manage/deploy/cloud-enterprise/configure-host-rhel-cloud.md
@@ -72,16 +72,24 @@ Verify that required traffic is allowed. Check the [Networking prerequisites](ec
 
 4. Install podman:
 
-    * For RHEL 8 and Rocky Linux, install version `4.*`.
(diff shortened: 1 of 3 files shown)


kilfoyle (approved):
LGTM! 🚢 
@florent-leborgne  It'd be nice to have a second set of eyes since I haven't viewed many PRs in the new system yet.

florent-leborgne (approved):
LGTM as well, thanks for making those changes @matt-elastic! 
Just to be sure, this is just for ECE 4.0 and later? Or should we also port this back to the "old docs" for ECE 3.8 or previous versions?

florent-leborgne:
Merging regardless of my previous question as it would be a different PR anyways in /cloud ^^

matt-elastic:
Thanks a lot for your reviews @kilfoyle @florent-leborgne 
I think these changes need to be back ported to previous versions as well.

florent-leborgne:
> I think these changes need to be back ported to previous versions as well.

Ok, I'll create a PR and tag you there, we can check which versions we want to backport to over there as well.

PR #627, opened 2025-02-28
Add missing Fleet page

colleenmcginnis:
I revisited the IA inventory and found [one page](https://github.com/elastic/ingest-docs/blob/do-not-delete_legacy-docs/docs/en/ingest-management/elastic-agent/otel-agent-transform.asciidoc) that existed on `main` in elastic/ingest-docs, but never made it to `current`.


lcawl (approved):
LGTM

PR #649, opened 2025-03-04
[kibana] Remove default OpenSSL legacy provider

jbudz:
Related to https://github.com/elastic/kibana/pull/213123

This default configuration is being removed in 9.0+.  For production, preferably no changes should be made and this section is no longer needed.


eedugon (approved):
LGTM!

PR #659, opened 2025-03-05
ECE: map Universal Profiling domains to proxy certificates

inge4pres:
We haven't yet included the Universal Profiling endpoints in the ECE documentation, despite Universal Profiling being GA since a while.

This PR adds documentation to include the domains.


inge4pres:
@bmorelli25 it would be nice to backport this to ECE version 3.7, which is the version where Universal Profiling became avialable

eedugon (approved):
Let's merge this! I have a huge PR that also touches this file, so I'll make sure the content doesn't get removed in case of a conflict :)

Thanks!!

PR #669, opened 2025-03-05
[Observability metrics reference] Remove duplicated file

alaudazzi:
This PR removes https://www.elastic.co/guide/en/serverless/current/observability-metrics-reference.html as it duplicates https://github.com/elastic/docs-content/tree/main/reference/data-analysis/observability/index.md

Relates to https://github.com/elastic/docs-projects/issues/432.


lcawl:
I wonder if this file shouldn't just be removed since it duplicates https://github.com/elastic/docs-content/tree/main/reference/data-analysis/observability/index.md?

Then just update the frontmatter in index.md to make it clear it replaces two pages:

```
---
mapped_pages:
  - https://www.elastic.co/guide/en/observability/current/metrics-reference.html
  - https://www.elastic.co/guide/en/serverless/current/observability-metrics-reference.html
 applies_to:
  stack: all
  serverless: all
---
```

Actually that `applies_to` is probably unnecessary, since it's asserting what should generally be the default (i.e. the page applies to everything).

lcawl (approved):
My assumption is we'll also do the same for the child pages (which currently have "-serverless" in the URL), but this page LGTM

PR #678, opened 2025-03-05
Deploy + manage: self-managed ES

shainaraskas:


deploy-manage/deploy/elastic-cloud/restrictions-known-problems.md
```diff
@@ -55,7 +55,7 @@ Elasticsearch APIs
 $$$ec-restrictions-apis-kibana$$$
 
 Kibana APIs
-:   There are no rate limits restricting your use of the Kibana APIs. However, Kibana features are affected by the [Kibana configuration settings](../self-managed/configure.md), not all of which are supported in {{ecloud}}. For a list of what settings are currently supported, check [Add Kibana user settings](edit-stack-settings.md). For all details about using the Kibana APIs, check the [Kibana API reference documentation](https://www.elastic.co/guide/en/kibana/current/api.html).
+:   There are no rate limits restricting your use of the Kibana APIs. However, Kibana features are affected by the [Kibana configuration settings](/deploy-manage/deploy/self-managed/configure-kibana.md), not all of which are supported in {{ecloud}}. For a list of what settings are currently supported, check [Add Kibana user settings](edit-stack-settings.md). For all details about using the Kibana APIs, check the [Kibana API reference documentation](https://www.elastic.co/guide/en/kibana/current/api.html).
```
shainaraskas, line 58:
```suggestion
:   There are no rate limits restricting your use of the Kibana APIs. However, Kibana features are affected by the [Kibana configuration settings](kibana://reference/configuration-reference.md), not all of which are supported in {{ecloud}}. For a list of what settings are currently supported, check [Add Kibana user settings](edit-stack-settings.md). For all details about using the Kibana APIs, check the [Kibana API reference documentation](https://www.elastic.co/docs/api/doc/kibana/).
```

deploy-manage/deploy/self-managed/_snippets/enable-auto-indices.md
```diff
@@ -0,0 +1,9 @@
+Some features automatically create indices within {{es}}. By default, {{es}} is configured to allow automatic index creation, and no additional steps are required. However, if you have disabled automatic index creation in {{es}}, you must configure [`action.auto_create_index`](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-create) in `elasticsearch.yml` to allow features to create the following indices:
```
eedugon, line 1: The link of `action.auto_create_index` doesn't feel correct. It should point to wherever https://www.elastic.co/guide/en/elasticsearch/reference/current/index-management-settings.html is.

Also in the text `if you have disabled automatic index creation` we should offer a link, as the reader may think `how that could be done`? as it's not obvious. The best page is probably https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#index-creation (wherever that is :) ).

eedugon: It's also important to clarify that all that section is only needed if the user has explicitly changed `action.auto_create_index` from its default value.

I would suggest to add in a banner that statement (that section is only needed if...) instead of the banner for logstash and beats.

deploy-manage/deploy/self-managed/_snippets/cmd-line-config.md
```diff
@@ -0,0 +1,21 @@
+{{es}} loads its configuration from the following location by default:
+
+```
+{{es-conf}}{{slash}}elasticsearch.yml
+```
```
eedugon, line 5:
```suggestion
```sh subs=true
{{es-conf}}{{slash}}elasticsearch.yml
```
```

File: deploy-manage/deploy/self-managed/_snippets/cmd-line-config.md [hunk 4]
```diff
@@ -0,0 +1,21 @@
...
+
+The format of this config file is explained in [](/deploy-manage/deploy/self-managed/configure-elasticsearch.md).
+
+Any settings that can be specified in the config file can also be specified on the command line, using the `-E` syntax as follows:
+
+```sh
+.\bin\elasticsearch.bat -Ecluster.name=my_cluster -Enode.name=node_1
+```
```
eedugon, line 13: elasticsearch.bat is only useful for windows systems, and this snippet is used also in the linux doc.
Not sure if there's any way to pass specific substitutions to snippets when calling them.

deploy-manage/deploy/self-managed/_snippets/enroll-nodes.md
```diff
@@ -0,0 +1,31 @@
...
+    ```sh
+    bin{{slash}}elasticsearch-create-enrollment-token -s node
+    ```
+
+    Copy the enrollment token, which you’ll use to enroll new nodes with your {{es}} cluster.
+
+2. From the installation directory of your new node, start {{es}} and pass the enrollment token with the `--enrollment-token` parameter.
+
+    ```sh
+    bin{{slash}}elasticsearch --enrollment-token <enrollment-token>
```
eedugon, line 20: why do you have {{slash}} in multiple places? why don't you just use `/`? 

Is it because of windows vs linux?

In such case we need `subs=true` in the code block to enable the substitution. There are more occurrences out there.

deploy-manage/deploy/self-managed/install-elasticsearch-docker-prod.md
```diff
@@ -0,0 +1,203 @@
...
+Pin your deployments to a specific version of the {{es}} Docker image. For example:
+
+```sh
+docker.elastic.co/elasticsearch/elasticsearch:{{stack-version}}
+```
+
+
+## Always bind data volumes [_always_bind_data_volumes]
+
+You should use a volume bound on `/usr/share/elasticsearch/data` for the following reasons:
```
eedugon, line 189:
```suggestion
You should use a volume bound to `/usr/share/elasticsearch/data` for the following reasons:
```

deploy-manage/deploy/self-managed/install-elasticsearch-docker-prod.md
```diff
@@ -0,0 +1,203 @@
...
+
+1. The data of your {{es}} node won’t be lost if the container is killed.
```
eedugon, line 191:
```suggestion
1. The data of your {{es}} node won’t be lost if the container is deleted.
```

deploy-manage/deploy/self-managed/_snippets/enroll-nodes.md
shainaraskas: Yes. The subs=true was still in development when I was getting this pr finalized but I do have to go back and annotate the relevant code blocks

PR #680, opened 2025-03-05
[security] Add missing mapped pages

colleenmcginnis:
Does it make sense to redirect the AsciiDoc URLs I added in this PR to the v3 files they were added to? I _think_ it might be that there are some AsciiDoc pages that are near duplicates, but I could be wrong about that.


benironside (approved):
Verified these redirects look good! Thank you Colleen

PR #686, opened 2025-03-06
[reference] Fix external links

colleenmcginnis:
In the migration tool, external URLs that contained `.html` were incorrectly modified to use `.md`.  

To (attempt) to fix these links:

1. I ran a script to find URLs that started with `http` (i.e. is not an internal or cross-repo link) and included `.md` somewhere in the URL.
2. Then I ran all the resulting URLs through an external link-checker.
    i. If I got a [`200` status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200) back, I didn't do anything to the link.
    ii. If I got a [`404` status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404) back, I swapped out the `.md` with `.html`.
3. I reran the link-checker to make sure all the updated links were now returning `200`.


leemthompo (approved):
lgtm

PR #709, opened 2025-03-07
change the supported stack versions

kvalliyurnatt:
Update supported stack version in the doc


kvalliyurnatt:
@eedugon I see that `supporteed-versions.asciidoc` was removed from the eck repo, but I am nto able to find that or an equivalent file here. Also I am able to find a file that matches the content in https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-operator-config.html

I am able to find a file with k8s-operator-config name but the contents seems a little different

naemono (approved):
from what I see it looks good, pending your questions.

eedugon:
thanks a lot for raising your concerns @kvalliyurnatt 

I'll check on Monday about the supported versions information.

Regarding the ECK configuration, we have moved the list of comfig flags to cloud-on-k8s/docs/reference and kept the rest here in docs-content.

The new content you see about ECK configuration feels a bit different because we have explained that the most common way to change the ECK configuration flags is through the ConfigMap provided in the standard installation, or through helm values if the helm installation was used.

The original (current for 2.16) doc lacks that information and it shows advanced methods (command line arguments or environment variables) that normal ECK users won't probably need (we have kept that part of the content at the end of the new doc). I hope you like the change.

As I also mentioned to Peter and Michael we can arrange a call to show and present all doc changes and of course if you want to revert anything (like bringing the config flags back to the config doc) we can also evaluate it, of course.

eedugon:
@kvalliyurnatt , I now see and understand what you mean about supported-versions.asciidoc original document.

The content of that document has been merged into the document you are editing in this PR, so that's the only copy of the information for 3.0 onwards.

During the migration from asciidoc to markdown and the adoption of the new Information Architecture (IA), we haven't kept a 1-1 map of original to new docs in all cases.

Some of the original docs / urls have been left almost intact, others have been refined in different ways to fit them into the new IA and narrative, and some other have been merged with others or into new docs.

What we do have and we can provide easily is a complete map of all original URLS (from current/2.16 docs) and what we have done with them (where they are now).

Also, if you feel that supported versions should be an individual document as before and not a section of the ECK overview document, we can analyze it.

Your feedback in all this is important and it will be more than welcome.

cc: @bmorelli25 , FYI

eedugon (approved):

kvalliyurnatt:
thanks @eedugon, the changes you mentioned make sense. I will merge this as it is

PR #714, opened 2025-03-10
Routing optimization for logsdb is gated by licensing

kkrik-es:
The subscription page is getting updated to contain an entry for `Custom routing on sort fields for logsdb`, limited to enterprise licenses.


marciw (approved):
thanks!

PR #717, opened 2025-03-10
typo: samling->sampling

carsonip:
Correct typo samling->sampling

### DIFF
--- a/solutions/observability/apps/tail-based-sampling.md
+++ b/solutions/observability/apps/tail-based-sampling.md
@@ -1,11 +1,11 @@
 ---
 mapped_pages:
-  - https://www.elastic.co/guide/en/observability/current/apm-tail-based-samling-config.html
+  - https://www.elastic.co/guide/en/observability/current/apm-tail-based-sampling-config.html
 applies_to:
   stack: all
 ---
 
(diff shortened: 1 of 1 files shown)


solutions/observability/apps/tail-based-sampling.md
```diff
@@ -1,11 +1,11 @@
 ---
 mapped_pages:
-  - https://www.elastic.co/guide/en/observability/current/apm-tail-based-samling-config.html
+  - https://www.elastic.co/guide/en/observability/current/apm-tail-based-sampling-config.html
```
carsonip, line 3: not sure how it affects ascii->md mapping. Does it require a corresponding change in old ascii observability docs too?

colleenmcginnis: Let's leave this the same. We're going to use `mapped_pages` to establish redirects from the URLs listed to the URL where this new content is published for the 9.0 release. Since the old URL uses the incorrect spelling, we'll still want to redirect from that misspelled URL to this new content. 

```suggestion
  - https://www.elastic.co/guide/en/observability/current/apm-tail-based-samling-config.html
```

colleenmcginnis (approved):

colleenmcginnis (approved):

PR #766, opened 2025-03-13
Add note to change-hardware.md docs

claudia-correia:
This PR adds a note to the change-hardware.md docs, to remind customers that the target instance configuration must be compatible with the tier/resource they're updating.


florent-leborgne (approved):
👍 Thanks Cláudia!

claudia-correia:
@florent-leborgne Thanks for the review and the additional info. I don't think this is urgent. @yuvielastic Would it be OK to release this docs the same day as the 9.0 release? 🙂 
(Also, I can't find the related asciidoc docs in the cloud repo. Maybe I'm missing something?)

florent-leborgne:
> Also, I can't find the related asciidoc docs in the cloud repo. Maybe I'm missing something?

Asciidoc files will remain in the cloud repo until up to the ms-122 branch. From ms-123 (and master), they're removed in favor of this repo (only some reference content remains in the cloud repo in branches ms-123+, in markdown).

PR #773, opened 2025-03-13
Update Fleet & Agent 'Upgrade' page

kilfoyle:
This makes just a few super small updates the Elastic Agent "Upgrade" steps for 9.0.

See: [preview page](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/773/reference/ingestion-tools/fleet/upgrade-elastic-agent)

I don't think there's anything more that we need for this page for 9.0, but @cmacknz please correct me if you think that's not right.

Note:
 - The screen captures all show 8.x versions, but I think that makes sense for now since we don't have 9.x versions to upgrade to.
 - The version numbers are hard-coded in the code blocks because we haven't settled on how version variables will be handled in the new system, but once we're decided on that I can fix these.

Closes: https://github.com/elastic/ingest-docs/issues/1730


kilfoyle:
Closing in favour of https://github.com/elastic/docs-content/pull/808 because we now have a variable for the stack versions.

PR #770, opened 2025-03-13
apm: Document tail-based sampling performance

carsonip:
Describe TBS requirements and give some numbers to demonstrate perf overhead.

Numbers are from benchmarks done in https://github.com/elastic/apm-server/issues/11346

Fixes https://github.com/elastic/apm-server/issues/11346

PR #776, opened 2025-03-13
Clean APM anchors

bmorelli25:
Removes attributes from anchor tags in `transaction-sampling.md`.

For #770.


colleenmcginnis (approved):
Attributes, be gone! ✨

PR #778, opened 2025-03-13
Clean up raw migrated files used in Troubleshoot section

marciw:
Double-check raw migrated files used in the Troubleshoot section and delete the ones we don't need anymore.

@shainaraskas one fyi for you: I confirmed that `ece-find-md`
is now represented in both Troubleshoot and Deploy & Manage, according to our divine mappings 🔮


natasha-moore-elastic (approved):

marciw:
> thanks @marciw !
> 
> @shainaraskas , I see the page is mapped in both places:
> 
> * deploy-manage/deploy/cloud-enterprise/search-filter-deployments.md
> * troubleshoot/deployments/cloud-enterprise/cloud-enterprise.md
> 
> In my opinion the page belongs to deploy and manage, as it's about filtering deployments (can be useful in troubleshooting to filter unhealthy deployments) but it's not directly related with troubleshooting at all. The title is misleading so we can also update the title and remove the `finding problems` from it.
> 
> Where should we fix this? If we agree I can create a quick PR and remove the duplicated mapping.

Shaina and I already discussed this during the mapping phase, and it's in the spreadsheet with both mappings... in the troubleshooting section, the page is called "Finding deployments" and we have plans to tailor it.

Happy for you to do whatever you like in D&M though of course :)

shainaraskas:
we need to remove the duplicated `mapped_pages` value if it exists (I'd probably prefer linking to the deploy and manage one) but I'm ok with the content being duplicated (prob need to use a snippet if we don't see it diverging soon)

marciw:
> we need to remove the duplicated `mapped_pages` value if it exists (I'd probably prefer linking to the deploy and manage one) but I'm ok with the content being duplicated (prob need to use a snippet if we don't see it diverging soon)

gotcha. 

@eedugon the `mapped_pages` value is correct [in the D&M file](https://github.com/elastic/docs-content/blob/main/deploy-manage/deploy/cloud-enterprise/search-filter-deployments.md)

stay tuned for a PR removing the mapping from the troubleshooting file

Issue #571, opened 2025-02-24
[REQUEST][AutoOps] Various doc updates


alaudazzi:
### Description

- [x] Replace the word `endpoint` with the word `connector` as we no longer use endpoint in the UI. 
https://github.com/elastic/docs-content/pull/786   
We should replace in all places except in:
  - [REST Endpoint Integration Guide](https://help.victorops.com/knowledge-base/rest-endpoint-integration-guide/)
  - Webhook integration
     - section 3 - endpoint_type
     - URL: This is the endpoint to which HTTP POST requests will be sent when events occur.
- [x] Add a new page on Email connector, as per this [gdoc](https://docs.google.com/document/d/1RffxHySe9Oh_1zG2S72OuK4FEjkKM79boRhY0RteI4Y/edit?tab=t.0) https://github.com/elastic/docs-content/pull/574
- [x] Update the [AutoOps regions page](https://docs-v3-preview.elastic.dev/elastic/docs-content/tree/main/deploy-manage/monitor/autoops/ec-autoops-regions) with `AWS | ap-southeast-1 |  Singapore` https://github.com/elastic/docs-content/pull/572

### Resources

Slack thread with PM and docs

### Which documentation set does this change impact?

Elastic On-Prem and Cloud (all)

### Feature differences

N/A

### What release is this request related to?

9.0

### Collaboration model

The documentation team

### Point of contact.

@alaudazzi 
@arnonst 

CC @eedugon

alaudazzi:
Relates to https://github.com/elastic/docs-content/issues/303

PR #786, opened 2025-03-14
[AutoOps] Replace endpoint with connector

alaudazzi:
This PR replaces `endpoint` with `connector` where needed, as we no longer use endpoint in the UI.

Closes https://github.com/elastic/docs-content/issues/571


arnonst (approved):
LGTM

PR #787, opened 2025-03-14
Removes unused Security raw migrated files

natasha-moore-elastic:
Removes Security ESS and serverless raw migrated files that are no longer needed for cleanup.


nastasha-solomon (approved):
Thank you!

PR #845, opened 2025-03-20
Fix links for Console

lcawl:
The Kibana doc link service currently links to https://www.elastic.co/guide/en/serverless/current/devtools-run-api-requests-in-the-console.html, which seems to be identical to https://www.elastic.co/guide/en/kibana/current/console-kibana.html.

This PR therefore adds the former URL to the list of mapped pages for https://docs-v3-preview.elastic.dev/elastic/docs-content/tree/main/explore-analyze/query-filter/tools/console and updates some other links to that page.


florent-leborgne (approved):
Thanks!

PR #860, opened 2025-03-20
Clean up raw files, fix malformed tables, rm unused subs

shainaraskas:
This PR: 

- Removes raw files I believe are used where they need to be used + their toc entries. generally:
   - 1:1 mapping with a page in the IA, populated with all of the content (and related reorg PRs are closed)
     OR
   - I know the area and reviewed the PRs that merged/split the content, and there are no stray mappings to the page
- Fixes tables that had an empty heading row and then a clear actual heading data row
- Removes unused subs (still throwing a warning due to https://github.com/elastic/docs-builder/issues/779)


lcawl (approved):
Skimmed only, but LGTM

PR #873, opened 2025-03-21
[Elasticsearch] The histogram field type does not support time_series_metric attribute

martijnvg:
This change also applies to 9.0.

The 8.x side is here: https://github.com/elastic/elasticsearch/pull/125366


leemthompo (approved):
Thanks @martijnvg!

PR #875, opened 2025-03-21
Adds `applies_to tags` to Troubleshoot > Ingest tools

natasha-moore-elastic:
Adds applies_to tags to the **Troubleshoot > Ingest tools** section.

Preview: [Ingest tools](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/875/troubleshoot/ingest)


marciw (approved):
Thanks! LGTM but I defer to the experts who are also reviewing 😁

karenzone (approved):
I was hoping that we could avoid versioning Ingest tools because they are vehicles for getting data into Elasticsearch and not _features_.   Troubleshooting makes for an interesting overlap, and I think it makes sense.

PR #881, opened 2025-03-21
ECE Networking Port Requirements Update

mmahacek:
Move note on Zookeeper ports from side note to be included in the table of ports.


eedugon (approved):
LGTM, thanks!

PR #882, opened 2025-03-21
Update configure-service-level-objective-slo-access.md

bmorelli25:
> The **correct** index priviledges are:
> ".slo-observability.*"

> We found out the correct ones through the SLO view in Kibana, where the necessary index privileges were stated if a > user was missing the correct access.
> We tested both versions and those that Kibana showed (with the dot) were the correct ones.

Closes https://github.com/elastic/observability-docs/issues/4720.


eedugon (approved):
LGTM! This should probably close https://github.com/elastic/observability-docs/issues/4720 then, right? :)

PR #889, opened 2025-03-21
clean up security-related raw files

shainaraskas:
This content is already used, so the raw files can be removed. one exception, and I moved over the content and added it to the ia/fixed the mapping.


bmorelli25 (approved):
🧹

PR #944, opened 2025-03-26
Clean up additional unused pages

shainaraskas:
removing observability-related pages that I confirmed were covered by one or more pages in the new solutions > observability section

+ related images
+ empty index pages we used to build the nav we removed


mdbirnstiehl (approved):
🦭 👍

PR #945, opened 2025-03-26
Fix hard-coded versions in Fleet & Agent docs

kilfoyle:
We have around 70 instances of `9.0.0-beta1` that need to be replaced in the F&A docs.


karenzone (approved):
LGTM! 🎸

PR #949, opened 2025-03-27
Update verify-zookeeper-sync-status.md

rseldner:
Revisiting this PR that fell through the cracks on my end https://github.com/elastic/cloud/pull/108690

Proposing an alternative zk sync status command that will provide a cleaner output and not rely on the use of additional system packages like `nc` or `telnet`.   We can achieve the same thing using `curl` which I think is just about ubiquitously available on linux hosts.

Current commit is written as an inline shell script.  So technically not a `one-liner`, but still easily copy/pasteable and much more readable. But here are the equivalent one-liners if we wish to stick to that format:

**container level**
```
docker exec frc-zookeeper-servers-zookeeper sh -c 'for i in $(seq 2191 2199); do output=$(echo mntr | curl -s telnet://localhost:$i | grep -E "server_state|leader|follower|not currently serving|zk_znode_count"); [ -n "$output" ] && echo "ZK mntr Response from port $i:" && echo "$output" && break; done'
```
**host level**
```
for i in $(seq 2191 2199); do output=$(echo mntr | curl -s telnet://localhost:$i | grep -E "server_state|leader|follower|not currently serving|zk_znode_count"); [ -n "$output" ] && echo "ZK mntr Response from port $i:" && echo "$output" && break; done
```

I've used a grep that will closely match the one-liner output without the noisey `trying port` lines...
**Leader output:**
```
ZK mntr Response from port 2191:
zk_server_state leader
zk_znode_count  783
zk_synced_followers     0
zk_synced_non_voting_followers  0
zk_leader_uptime        795608083
zk_avg_leader_unavailable_time  293.0
zk_min_leader_unavailable_time  293
zk_max_leader_unavailable_time  293
zk_cnt_leader_unavailable_time  1
zk_sum_leader_unavailable_time  293
zk_avg_follower_sync_time       0.0
zk_min_follower_sync_time       0
zk_max_follower_sync_time       0
zk_cnt_follower_sync_time       0
zk_sum_follower_sync_time       0
```

But we could consider this grep for an even cleaner output
```
grep -E "^zk_server_state|^zk_followers|^zk_synced_followers|not currently serving|^zk_znode_count"
```
**Leader Output:**
```
ZK mntr Response from port 2191:
zk_server_state leader
zk_znode_count  783
zk_synced_followers     0
```

### DIFF
--- a/troubleshoot/deployments/cloud-enterprise/verify-zookeeper-sync-status.md
+++ b/troubleshoot/deployments/cloud-enterprise/verify-zookeeper-sync-status.md
@@ -16,10 +16,18 @@ It is recommended to check the ZooKeeper sync status before starting any mainten
 
 To check that ZooKeeper is in sync with the correct number of followers, run the following steps:
 
-1. Run the one-line command on each Director node:
+1. Run the inline shell script command on each Director node:
 
     ```sh
-    docker exec frc-zookeeper-servers-zookeeper sh -c 'for i in $(seq 2191 2199); do echo trying port: $i;echo mntr | nc localhost ${i} 2>/dev/null | grep "not currently serving";echo mntr | nc localhost ${i} 2>/dev/null| grep leader; echo mntr | $(which nc) localhost ${i} 2>/dev/null | grep follower ; done'
+    docker exec frc-zookeeper-servers-zookeeper sh -c '
+    for i in $(seq 2191 2199); do 
+      output=$(echo mntr | curl -s telnet://localhost:$i | grep -E "server_state|leader|follower|not currently serving|zk_znode_count"); 
+      if [ -n "$output" ]; then 
+        echo "ZK mntr Response from port $i:"; 
+        echo "$output"; 
+        break; 
+      fi 
+    done'
     ```
 
     ::::{note} 
@@ -32,53 +40,32 @@ To check that ZooKeeper is in sync with the correct number of followers, run the
     * All followers are listed as synced
 
 
-The one-line command can return the following types of output:
+The inline shell script command can return the following types of output:
 
 * If the host is the current ZooKeeper Leader, the command returns the Leader’s info including follower count and follower sync status.
 
     ```
-    trying port: 2191
+    ZK mntr Response from port 2191:
     zk_server_state leader
-    zk_followers    2
+    zk_znode_count  783
     zk_synced_followers     2
-    trying port: 2192
-    trying port: 2193
-    trying port: 2194
-    trying port: 2195
-    trying port: 2196
-    trying port: 2197
-    trying port: 2198
-    trying port: 2199
+    ...
     ```
 
 * If the host is a follower, the command returns only the follower state, and continues until it finds the Leader:
 
     ```
-    trying port: 2191
-    trying port: 2192
-    trying port: 2193
+    ZK mntr Response from port 2193:
     zk_server_state follower
-    trying port: 2194
-    trying port: 2195
-    trying port: 2196
-    trying port: 2197
-    trying port: 2198
-    trying port: 2199
+    zk_znode_count  777
+    ...
     ```
 
 * If the ZooKeeper container is up and listening, but the current node doesn’t have the quorum, the command returns the message `This ZooKeeper instance is not currently serving requests`:
 
     ```
-    trying port: 2191
-    trying port: 2192
+    ZK mntr Response from port 2192:
     This ZooKeeper instance is not currently serving requests
-    trying port: 2193
-    trying port: 2194
-    trying port: 2195
-    trying port: 2196
-    trying port: 2197
-    trying port: 2198
-    trying port: 2199
     ```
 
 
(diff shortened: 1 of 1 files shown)


kunisen (commented):
@rseldner thanks for the doc PR. 

I have some thoughts:

### [1] 
I checked chatgpt and found `curl telnet://` isn't a standard approach for interacting with raw TCP services. Instead, use `nc (netcat)` or `telnet` directly was recommended. 

WDYT?

### [2]
I understand it might be in the old days we didn't have nc bundled in container (e.g. https://github.com/elastic/support-dev-help/issues/7458#issuecomment-513885696 - `nc: command not found`) but nowadays we already have `nc` there by default. 

I don't have strong opinion but IMHO, there's no absolute need for us to replace `nc` with `curl` nowadays. 
If you strongly suggest this, it's fine. 

### [3]

This PR only works for ECE 4+ but not for ECE 3.x.
Do we also want to backport this to ECE 3.x docs? 
(If so, then filing a pairing doc PR in https://github.com/elastic/cloud/issues but using `asciidoc` format is required)

troubleshoot/deployments/cloud-enterprise/verify-zookeeper-sync-status.md
```diff
@@ -32,86 +40,53 @@ To check that ZooKeeper is in sync with the correct number of followers, run the
...
-    * for example `telnet localhost 2191`
-
-3. Look for the following output lines:
-
-    * `zk_server_state leader` or `zk_server_state follower` to indicate node leadership
-    * Lines indicating the follower count and sync status when run against a Leader node
-
-        * `zk_followers    2`
-        * `zk_synced_followers     2`
+If the inline shell script command doesn’t work (e.g. your user lacks permissions to access docker), you can run the check directly from the director host. This approach avoids entering the container and doesn't require installing additional tools like `telnet` or `nc`, relying instead on `curl`, which is typically available by default on most Linux systems.
```
florent-leborgne, line 77:
```suggestion
If the inline shell script command doesn’t work, you can run the check directly from the director host. This can happen for example when your user lacks permissions to access Docker. This approach avoids entering the container and doesn't require installing additional tools like `telnet` or `nc`, relying instead on `curl`, which is typically available by default on most Linux systems.
```
Small tweak to avoid parenthesis and phrasing like "e.g."

troubleshoot/deployments/cloud-enterprise/verify-zookeeper-sync-status.md
```diff
@@ -32,86 +40,53 @@ To check that ZooKeeper is in sync with the correct number of followers, run the
...
 
+1. Run the equivalent inline shell script directly on the host terminal (outside of the zookeeper container)
```
florent-leborgne, line 79:
```suggestion
1. Run the equivalent inline shell script directly on the host terminal, outside of the zookeeper container
```

troubleshoot/deployments/cloud-enterprise/verify-zookeeper-sync-status.md
```diff
@@ -32,86 +40,53 @@ To check that ZooKeeper is in sync with the correct number of followers, run the
...
+    for i in $(seq 2191 2199); do 
+      output=$(echo mntr | curl -s telnet://localhost:$i | grep -E "server_state|leader|follower|not currently serving|zk_znode_count"); 
+      if [ -n "$output" ]; then 
+        echo "ZK mntr Response from port $i:"; 
+        echo "$output"; 
+        break; 
+      fi 
+    done
+    ```
+2. Look for the following lines in the output (just as noted above)
```
florent-leborgne, line 90:
```suggestion
2. Look for the following lines in the output
```

florent-leborgne (commented):
LGTM from a docs standpoint, with minor styling suggestions.

Not explicitly approving until Kuni & team validate this technically. Thanks!

rseldner:
Thanks for the feedback @kunisen

**[1]**

In more security-constrained environments, running commands inside containers or installing new packages (nc, telnet) isn’t always feasible (even for an ECE admin).  Requesting to install telnet might get laughed out of the room by some security teams 😅 . On the other hand, curl is almost always pre-installed on ECE hosts, making it reliable for ad-hoc troubleshooting.  The chatgpt recommendation you received likely didn’t account for these sorts of operational realities.  Security constraints aside, using `curl telnet://` works perfectly fine for issuing a simple mntr command to Zookeeper.  I can give chatgpt a prompt myself that acknowledges this. (but why even rely on chatgpt?)

**[2]**

My main goal here was to offer a host-level curl alternative to replace the "use telnet instead" section.  I also updated the container-level command as well for the sake of consistency, but we could absolutely leave that version using `nc` if preferred.  We could alternatively document both: nc for the container and curl for the host. That said, switching to host-level curl makes the dilemma a moot point.  It's portable and practical.

Also, the change I proposed also returns a slightly cleaner output and output lines sometimes request from support.

**[3]**
Understood.  I started with repo given the upcoming doc changes (at the time)

troubleshoot/deployments/cloud-enterprise/verify-zookeeper-sync-status.md
```diff
@@ -16,10 +16,18 @@ It is recommended to check the ZooKeeper sync status before starting any mainten
 
 To check that ZooKeeper is in sync with the correct number of followers, run the following steps:
```
kunisen, line 17:
```suggestion

### Check at container level

To check that ZooKeeper is in sync with the correct number of followers, run the following steps:
```

troubleshoot/deployments/cloud-enterprise/verify-zookeeper-sync-status.md
kunisen, line 77:
```suggestion

### Check at host level

If the inline shell script command doesn’t work (e.g. your user lacks permissions to access docker), you can run the check directly from the director host. This approach avoids entering the container and doesn't require installing additional tools like `telnet` or `nc`, relying instead on `curl`, which is typically available by default on most Linux systems.
```

kunisen:
```suggestion

### Check at host level

If the inline shell script command doesn’t work (e.g. your user lacks permissions to access docker), you can run the check directly from the director host. This can happen for example when your user lacks permissions to access Docker. This approach avoids entering the container and doesn't require installing additional tools like `telnet` or `nc`, relying instead on `curl`, which is typically available by default on most Linux systems.
```

kunisen (approved):
Thanks @rseldner for the detailed explanation.
LGTM overall, I left a small suggestion which is to add the section headings to make the structure more logically visible.

**:: Side note - 1**
I wasn't aware that doing in `curl telnet://` doesn't need to install telnet package. TIL. Thanks!

Also I had a test and found we still have the same issue - some OS doesn't have `telnet` and `nc` package installed by default, but the error was hidden by `2>/dev/null` which was 🤦 very unfortunate, it doesn't tell you the reason even there's no `nc` in container image (which I tested and confirmed the behavior).

```bash
# No error
elastic@sl-kuniyasusen-d66e44db-host1:~$ docker exec frc-zookeeper-servers-zookeeper sh -c 'for i in $(seq 2191 2199); do echo trying port: $i;echo mntr | nc localhost ${i} 2>/dev/null | grep "not currently serving";echo mntr | nc localhost ${i} 2>/dev/null| grep leader; echo mntr | $(which nc) localhost ${i} 2>/dev/null | grep follower ; done'
trying port: 2191
trying port: 2192
trying port: 2193
...

# But actually there's no nc package
elastic@sl-kuniyasusen-d66e44db-host1:~$ docker exec -it frc-zookeeper-servers-zookeeper sh -c 'for i in $(seq 2191 2199); do echo trying port: $i;echo mntr | nc localhost ${i} ; done'
trying port: 2191
sh: nc: not found
trying port: 2192
sh: nc: not found
trying port: 2193
...
``` 

**:: Side note 2**

> In more security-constrained environments, running commands inside containers or installing new packages (nc, telnet) isn’t always feasible (even for an ECE admin).

I think I would disagree. 

In lots of situations we need customer's understanding and cooperation to take the information that is essential for troubleshooting. 

In many cases, we argue with them about getting ECE diags, or ES diags for us, which I understand the security concern, however, this is crucial to us. 
Also, one more example, `sar` is another thing we sometimes need, but this is not shipped by most OS packages by default. 

I get your point - we should use the libraries and packages that are commonly bundled by most OS by default, as much as we can, but that doesn't mean we have to close the door due to i.e. security reason. 

(One recent thing that hit this is we deprecated the built-in ECE diag and now we need to download it from Internet. For most ECE platform related cases, we have to let customer understand we need this.)

leemthompo:
@florent-leborgne @kunisen @rseldner is this good to merge?

florent-leborgne (approved):
LGTM

PR #955, opened 2025-03-27
[Solutions] Adds info about preconfigured LLM

szabosteve:
## Overview

This PR adds info about the preconfigured LLM to the Playground and AI Assistant docs.


leemthompo (approved):
LGTM with very minor optional suggestion for parallelism :)

shubhaat (changes requested):
We want to not use the term "Elastic LLM" for now and use a more general the preconfigured connector or preconfigured LLM where possible to be accurate.

shubhaat (approved):

PR #1007, opened 2025-04-02
Add mapping and redirect for hidden security tutorial

shainaraskas:
[This tutorial](https://github.com/elastic/docs-content/blob/main/raw-migrated-files/stack-docs/elastic-stack/install-stack-demo-secure.md) is hidden while we figure out what content is still needed from it (most of it is a duplicate, and it's only relevant to self-managed)

so I have mapped the page to the explanation of how to set up security in a self-managed environment and added a redirect.


colleenmcginnis (approved):
🙏

PR #1022, opened 2025-04-03
instructions for es multinode + kib external access

shainaraskas:
Harvested all of the content from the old [install stack tutorial](https://github.com/elastic/docs-content/blob/main/raw-migrated-files/stack-docs/elastic-stack/installing-stack-demo-self.md) that was missing from the other docs. 

namely: 

1. set up your first node so other nodes can connect to it (example in [RPM](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1022/deploy-manage/deploy/self-managed/install-elasticsearch-with-rpm#step-3-set-up-the-node-for-connectivity), [Win](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1022/deploy-manage/deploy/self-managed/install-elasticsearch-with-zip-on-windows#step-3-optional-set-up-the-node-for-connectivity) - the steps have a different order across install types because of the availability of the `elasticsearch-reconfigure-node` tool for RPM and debian)
2. expose kibana so remote users can connect ([example](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1022/deploy-manage/deploy/self-managed/install-kibana-on-windows#step-3-optional-make-kib-externally-accessible))

^ both of these were added as optional steps to the es and kibana install tutorials. I also simplified and deduplicated initial TCP config info across the install guides

Wonder if the kibana instructions should instead be part of the config step, but not sure if the correct `server.host` value needs to be present when kibana is enrolled. this preserves the order from the old tutorial (config then start).

I also rounded up the [ports](https://www.elastic.co/guide/en/elastic-stack/current/installing-stack-demo-self.html#install-stack-self-overview) from the old image and added them in the installation overviews - not necessarily the best spot but maybe an adequate one? took the content from a [knowledge article](https://support.elastic.co/knowledge/CfT70cu9) and added the relevant settings

open to feedback about whether this is a good addition or not

closes https://github.com/elastic/docs-projects/issues/462


deploy-manage/deploy/self-managed/_snippets/first-node.md
```diff
@@ -0,0 +1,34 @@
+Update the {{es}} configuration on this first node so that other hosts are able to connect to it by editing the settings in [`elasticsearch.yml`](/deploy-manage/deploy/self-managed/configure-elasticsearch.md):
+
+1. Retrieve the external IP address of your host. You’ll need this value later.
```
shainaraskas, line 3: I made this really generic so it could be used across tutorials - let me know whether you think I should be more specific. old tutorial specified:

>In a terminal, run the ifconfig command and copy the value for the host inet IP address (for example, 10.128.0.84). You’ll need this value later.

^ but that means I'd have to get the proper command + variable for windows/linux/macos as well

shainaraskas:
found [something I was missing](https://www.elastic.co/guide/en/elastic-stack/current/installing-elastic-stack.html#network-requirements) .... todo

shainaraskas:
found [something I was missing](https://www.elastic.co/guide/en/elastic-stack/current/installing-elastic-stack.html#network-requirements) .... todo

deploy-manage/deploy/self-managed/_snippets/ports.md
```diff
@@ -0,0 +1,9 @@
...
+
+These settings can be overridden in the relevant configuration file.
+
+| Port | Access type | Purpose | Setting |
+| --- | --- | --- | --- |
+| 9200-9300	| HTTP (REST) | REST API for Elasticsearch. This is the primary interface used for access to the cluster from external sources, including {{kib}} and {{ls}}. | Elasticsearch [`http.port`](elasticsearch://reference/elasticsearch/configuration-reference/networking-settings.md#common-network-settings) |
+| 9300-9400	| TCP |	Transport API. Used for intra-cluster communications and client access via the transport API (Java client). | Elasticsearch [`transport.port`](elasticsearch://reference/elasticsearch/configuration-reference/networking-settings.md#common-network-settings) |
+| 5601 | HTTP |	{{kib}} default access port. | Kibana [`server.port`](kibana://reference/configuration-reference/general-settings.md#server-port) |
```
shainaraskas, line 9:
```suggestion
| 5601 | HTTP |	{{kib}} default access port. | Kibana [`server.port`](kibana://reference/configuration-reference/general-settings.md#server-port) |

Additional ports might be required for [optional {{stack}} components](/get-started/the-stack.md). Refer to the installation guide for the component that you want to install.
```

deploy-manage/deploy/self-managed/_snippets/first-node.md
eedugon: My advice is to suggest always `0.0.0.0` as the best practice and explain that that means `listening in all interfaces of the host`. Then if the user prefers to bind to a specific IP address or network interface they can do it (and sharing a link to the network reference documentation where all possible syntaxes are explained).

Binding to a specific IP address makes for example connections to `localhost:9200` or `127.0.0.1:9200` from the same host to not work, and causes confusion to some users.

Also, suggesting 0.0.0.0 and explaining what that means gives more help than the other way around (specifying a fake static IP address, explaining that it's an example and then explaining that 0.0.0.0 can also be used).

I'd suggest 0.0.0.0 in both `elasticsearch` transport and http configuration, and also for kibana `server.host`. The majority of production systems everywhere use that approach, as binding into specific IP address doesn't really add much security benefits and make for example IP changes a nightmare.

deploy-manage/deploy/self-managed/_snippets/join-existing-cluster.md
```diff
@@ -25,9 +25,9 @@ When you install {{es}}, the installation process configures a single-node clust
    The `elasticsearch-reconfigure-node` tool has updated several settings. For example:
 
    * The `transport.host: 0.0.0.0` setting is already uncommented.
-   * The `discovery_seed.hosts` setting has the value that you added for `network_host` on the first {{es}} node. As you add each new {{es}} node to the cluster, the `discovery_seed.hosts` setting will contain an array of the IP addresses and port numbers to connect to each {{es}} node that was previously added to the cluster.
+   * The `discovery_seed.hosts` setting has the value has the IP address and port of the other {{es}} nodes added the cluster so far. As you add each new {{es}} node to the cluster, the `discovery_seed.hosts` setting will contain an array of the IP addresses and port numbers to connect to each {{es}} node that was previously added to the cluster.
```
shainaraskas, line 28:
```suggestion
   * The `discovery_seed.hosts` setting has the IP address and port of the other {{es}} nodes added the cluster so far. As you add each new {{es}} node to the cluster, the `discovery_seed.hosts` setting will contain an array of the IP addresses and port numbers to connect to each {{es}} node that was previously added to the cluster.
```

eedugon (commented):
I think I forgot to apply my review comments last week before your day started and I started to run tests in my afternoon :)

Sharing them now, as I think they are still relevant. Please ignore otherwise.

I'll start a new review now checking the changes we talked about the procedure itself.

deploy-manage/deploy/self-managed/_snippets/clean-up-multinode.md
```diff
@@ -0,0 +1,9 @@
+Perform the following steps on each node in the cluster:
+
+1. Open `elasticsearch.yml` in a text editor.
+2. Comment out or remove the `cluster.initial_master_nodes` setting.
```
eedugon, line 4: > 2. Comment out or remove the `cluster.initial_master_nodes` setting.

Would it make sense to indicate that that setting will only appear in the first node?

deploy-manage/deploy/self-managed/_snippets/clean-up-multinode.md
```diff
@@ -0,0 +1,9 @@
...
+3. Update the `discovery.seed_hosts` value so it contains the IP address and port of each of the master-eligible {{es}} nodes in the cluster.
```
eedugon, line 5: > 3. Update the `discovery.seed_hosts` value so it contains the IP address and port of each of the master-eligible {{es}} nodes in the cluster.


Would it make sense to indicate that first node won't even have this setting? And the user has to add it.

The most important part of this section is the first node, although I like how it's explained as all nodes need to be reviewed.

deploy-manage/deploy/self-managed/_snippets/reset-superuser-rpm-deb.md
eedugon, line 1: Does it make sense to explain that this step is not needed for all the nodes? Just once after the cluster is up for example after starting the initial node only?

deploy-manage/deploy/self-managed/install-elasticsearch-with-debian-package.md
```diff
@@ -135,17 +145,29 @@ sudo dpkg -i elasticsearch-{{stack-version}}-amd64.deb
...
 :::{include} _snippets/pw-env-var.md
 :::
 
-## Step 6: Check that {{es}} is running [deb-check-running]
+## Step 7: Check that {{es}} is running [deb-check-running]
 
 :::{include} _snippets/check-es-running.md
 :::
 
+## Step 8 (Multi-node clusters only): Clean up the config files
```
eedugon, line 164: Not my expertise area, but would `consolidate configuration files` sound better than cleaning up?  For your review and consideration.

deploy-manage/deploy/self-managed/install-elasticsearch-with-debian-package.md
```diff
@@ -135,17 +145,29 @@ sudo dpkg -i elasticsearch-{{stack-version}}-amd64.deb
...
+
+If you are deploying a multi-node cluster, then the `elasticsearch-reconfigure-node` tool adds all existing nodes to each newly enrolled node's `discovery.seed_hosts` setting. However, you need to go back to all of the nodes in the cluster and edit them so each node in the cluster can restart and rejoin the cluster as expected.
+
```
eedugon, line 167: I would also remark that the initial node won't even have seed_hosts configured (because the cluster was bootstrapped as a single-node cluster), which is a mandatory setting for multi-node clusters.

eedugon (approved):
Looks really great. I've left a few comments for your consideration.
If you need another review let me know!

deploy-manage/deploy/self-managed/_snippets/clean-up-multinode.md
shainaraskas: added `if present` - just want to avoid sending the message that it should be appearing anywhere and it's a bit of a wording challenge.

deploy-manage/deploy/self-managed/install-elasticsearch-with-debian-package.md
shainaraskas: `consolidate` to me means that you are merging multiple config files into fewer config files, instead of editing them

shainaraskas: changed to "update"

PR #1049, opened 2025-04-07
Update the-stack.md

georgewallace:
removed link as ES and Kibana are referenced in this page. The link no longer made sense with the changing of the other page.


marciw (approved):
🚗

PR #1062, opened 2025-04-09
adding enterprise search comments and links to ECK 2.16 docs

eedugon:
ECK 3.0 still supports Enterprise Search for deployments running 8.x versions of Elastic Stack (and maybe even 7.17).

We had removed almost all references to Enterprise Search during the migration and that will look odd considering that some users might still run ECK-managed deployments with Enterprise Search.

This PR adds mentions to Enterprise Search in 3 different docs:

- Stack helm chart utilization
- Orchestrating other elastic stack applications
- Network policies examples

Another approach for this would be to bring back all content related to Enterprise Search from older docs, but maybe it's better this way.

@shainaraskas , @bmorelli25 : I'd need to know if the links I'm using for the older documents are correct in terms of destination.

### DIFF
--- a/deploy-manage/deploy/cloud-on-k8s/managing-deployments-using-helm-chart.md
+++ b/deploy-manage/deploy/cloud-on-k8s/managing-deployments-using-helm-chart.md
@@ -80,6 +80,10 @@ helm install eck-stack-with-apm-server elastic/eck-stack \
     --values https://raw.githubusercontent.com/elastic/cloud-on-k8s/2.16/deploy/eck-stack/examples/apm-server/basic.yaml -n elastic-stack
 ```
 
+## Enterprise Search server along with Elasticsearch and Kibana [k8s-install-enterprise-search-elasticsearch-kibana-helm]
+
+Enterprise Search is not available in {{stack}} versions 9.0 and later. For an example deployment of {{es}}, {{kib}}, and an Enterprise Search server using the Helm chart, refer to the [previous ECK documentation](https://www.elastic.co/guide/en/cloud-on-k8s/2.16/k8s-stack-helm-chart.html).
+
 ## Install individual components of the Elastic Stack [k8s-eck-stack-individual-components] 
 
 You can install individual components in one of two ways using the provided Helm Charts.
--- a/deploy-manage/security/k8s-network-policies.md
+++ b/deploy-manage/security/k8s-network-policies.md
@@ -433,3 +433,6 @@ spec:
       common.k8s.elastic.co/type: logstash
 ```
 
+## Isolating Enterprise Search [k8s-network-policies-enterprise-search-isolation]
+
+Enterprise Search is not available in {{stack}} versions 9.0 and later. For an example of Enterprise Search isolation using network policies, refer to the [previous ECK documentation](https://www.elastic.co/guide/en/cloud-on-k8s/2.16/k8s_prerequisites.html#k8s-network-policies-enterprise-search-isolation).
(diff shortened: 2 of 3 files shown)


deploy-manage/deploy/cloud-on-k8s/managing-deployments-using-helm-chart.md
```diff
@@ -80,6 +80,10 @@ helm install eck-stack-with-apm-server elastic/eck-stack \
     --values https://raw.githubusercontent.com/elastic/cloud-on-k8s/2.16/deploy/eck-stack/examples/apm-server/basic.yaml -n elastic-stack
 ```
 
+## Enterprise Search server along with Elasticsearch and Kibana [k8s-install-enterprise-search-elasticsearch-kibana-helm]
```
shainaraskas, line 83:
```suggestion
## Enterprise Search server along with {{es}} and {{kib}} [k8s-install-enterprise-search-elasticsearch-kibana-helm]
```

deploy-manage/deploy/cloud-on-k8s/managing-deployments-using-helm-chart.md
```diff
@@ -80,6 +80,10 @@ helm install eck-stack-with-apm-server elastic/eck-stack \
...
+
+Enterprise Search is not available in {{stack}} versions 9.0 and later. For an example deployment of {{es}}, {{kib}}, and an Enterprise Search server using the Helm chart, refer to the [previous ECK documentation](https://www.elastic.co/guide/en/cloud-on-k8s/2.16/k8s-stack-helm-chart.html).
```
shainaraskas, line 85:
```suggestion
Enterprise Search is not available in {{stack}} versions 9.0 and later. For an example deployment of {{es}} version 8.x, {{kib}} 8.x, and an 8.x Enterprise Search server using the Helm chart, refer to the [previous ECK documentation](https://www.elastic.co/guide/en/cloud-on-k8s/2.16/k8s-stack-helm-chart.html).
```

deploy-manage/security/k8s-network-policies.md
```diff
@@ -433,3 +433,6 @@ spec:
       common.k8s.elastic.co/type: logstash
 ```
 
+## Isolating Enterprise Search [k8s-network-policies-enterprise-search-isolation]
+
+Enterprise Search is not available in {{stack}} versions 9.0 and later. For an example of Enterprise Search isolation using network policies, refer to the [previous ECK documentation](https://www.elastic.co/guide/en/cloud-on-k8s/2.16/k8s_prerequisites.html#k8s-network-policies-enterprise-search-isolation).
```
shainaraskas, line 438:
```suggestion
Enterprise Search is not available in {{stack}} versions 9.0 and later. For an example of Enterprise Search isolation using network policies in previous {{stack}} versions, refer to the [previous ECK documentation](https://www.elastic.co/guide/en/cloud-on-k8s/2.16/k8s_prerequisites.html#k8s-network-policies-enterprise-search-isolation).
```

shainaraskas (approved):
couple small edits

shainaraskas:
>Another approach for this would be to bring back all content related to Enterprise Search from older docs, but maybe it's better this way.

@leemthompo who owns these docs has been asked not to do this so we'll leave it out for now

>@shainaraskas , @bmorelli25 : I'd need to know if the links I'm using for the older documents are correct in terms of destination.

These look fine - explicitly specifying the ECK version will make sure these links are live as long as they're relevant AFAIK

> What should we do with URL mappings of those components? For example [elastic.co/guide/en/cloud-on-k8s/current/k8s-enterprise-search.html](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-enterprise-search.html) and all childs have been removed in the new docs system. Should we create redirects or something to point them to the 2.16 version of the docs? or should they all be mapped to the new version of the docs that tells users that Enterprise Search is not supported in 9.x and link to the 2.16 docs?

These are currently redirecting to 2.16 and I think that's the right choice. you can see the mappings in [this spreadsheet](https://docs.google.com/spreadsheets/d/1bQQVc-QwwXDSd24Vw6Q62dnn_9bxsBeAzeWJAdyxdqI/edit?gid=1108496812#gid=1108496812).

PR #1092, opened 2025-04-11
Add note about keyword ignore_above logsdb setting

limotova:
Specifies that the setting applies to all members of the keyword family.


martijnvg (approved):
LGTM

PR #1093, opened 2025-04-11
Clarify ECH node bootlooping documentation

kunisen:
## Description

Per an internal discussion (`sdh-cp #9409`), we would like to update the ECH doc to clarify the node bootlooping related doc. 

### Background

A setting is configured in Elasticsearch but it reports error

```bash
org.elasticsearch.common.settings.SettingsException: 
The configuration setting [xpack.security.authc.realms.foobar.foobar1.foobar2.client_secret] is required
```

And during the troubleshooting, we found the setting is configured via client tooling and it's not properly configured (zero byte). 

Based on that, we logically thought to solve the issue, users should 
- understand this may happen if the keystore setting (`[xpack.security.authc.realms.foobar.foobar1.foobar2.client_secret]` to be configured on ECH side) is invalid, missing or improperly configured
- understand the way to solve is to check if the keystore is valid or not and make sure it's configured properly
- additionally if users configured this via client tooling, they should try configuring this in cloud UI directly as a way to isolate the cause

### After PR being merged

The following orange section will be added

image

image


gbocchini (approved):

maggieghamry:
Thanks so much @kunisen, is there a doc with the text outside of the image to make suggestions to? I would recommend having an example next to `via a client tool or API` (for example `terraform-ec`) - I have seen 4 different issues related to ECH and `terraform-ec` this week, often with threads that are working towards preflight. If we can help establish the documentation baseline where we clearly have a reference with a mention of tools we see often, I believe that can help on both ends.

AlexP-Elastic (approved):
LGTM

maggieghamry (approved):
LGTM! small note: perhaps having an example next to "via a client tool or API" (for example terraform-ec) or perhaps utilize a knowledge article separately.

kunisen:
Thanks @maggieghamry sure, I added terraform ec provider as an example as suggested.
image

PR #1107, opened 2025-04-11
[Fleet] Update procedure to change logging level

kilfoyle:
In the Fleet & Agent docs, this updates the steps to "Change the logging level":

screen

---

@Supplementing, Thanks for the nice and clear docs issue! Just FYI: with our new docs system we don't yet have a "future" branch, so we call out version differences are directly in the text, hence the note about "Prior to Version 9.1". It's not ideal, but I think it's the best approach we have for right now.

Closes: https://github.com/elastic/ingest-docs/issues/1752


Supplementing (approved):
LGTM 🚀

PR #1109, opened 2025-04-11
move restrictions doc up a level

shainaraskas:
found a cloud restrictions doc, but all of the content only applies to cloud hosted, so I nested it in the cloud hosted section

also did some spot cleanup of the ecloud overview while I was checking for an x-link to the page


marciw (approved):
pesky build error? otherwise LGTM

PR #1121, opened 2025-04-14
eck_version, eck_release_branch and eck_helm_minimum_version subs

eedugon:
ECK substitutions applied, using current asciidoc (2.16) substitutions as reference.

`eck_helm_minimum_version` is an improvement as the minimum supported helm version appeared hardcoded in multiple docs.

Note that in the current asciidoc we also have a substitution like `eck_github: https://github.com/elastic/cloud-on-k8s` not used here. If that's considered useful I can apply it everywhere also.

Pending to determine if we should add also `eck_crd_version` the same way it was used in original asciidoc files.


docset.yml
```diff
@@ -272,6 +272,8 @@ subs:
   kib-pull:   "https://github.com/elastic/kibana/pull/"
   stack-version: "9.0.0"
   eck_version: "3.0.0"
+  eck_release_branch: "3.0"
+  eck_helm_minimum_version: "3.2.0"
```
barkbay, line 276: IIUC these fields are going to be maintained by the ECK team? If it is the case we should update our release issue template: https://github.com/elastic/k8s-dev/blob/main/.github/ISSUE_TEMPLATE/release-issue-template.md

deploy-manage/upgrade/orchestrator/upgrade-cloud-on-k8s.md
shainaraskas, line 1: on this page, there's a visible variable (this is likely my fault). could you clean it up while you're here?

image

shainaraskas: image

eedugon: of course!

eedugon: should be solved now.

docset.yml
barkbay: > Do you want me to write a PR or an issue to update your k8s-dev releases template? Although I'm not familiar with your overall releasing procedure.

I update the last step in the release process to not forget to update this template. See the last step in the "Post release" section [here](https://github.com/elastic/k8s-dev/issues/523).

eedugon: perfect! In that template I still see mentions to the old `asciidoc` files and not to the new markdown docs.
Remember that now we don't use the asciidoc system anymore and some of the docs that appear in the template doesn't exist in `main`.

kvalliyurnatt: Thanks @eedugon and @barkbay  I will make sure the references to asciidoc are updated and also will add the other changes to the releasee template

eedugon: perfect @kvalliyurnatt, thanks! If you need any help or information from my side about new docs system or when to touch the old `asciidoc` files (available only for 2.x branches), ping me in private!

PR #1122, opened 2025-04-14
eck resources list substitution added

eedugon:
Based on the original substitutions configured on ECK original asciidoc documentation, which were lost during the migration.


shainaraskas (approved):
handy! we should keep this in mind if we ever get [variable nesting](https://github.com/elastic/docs-builder/issues/152), because we need to keep the official product names correct inside of the variables

PR #1128, opened 2025-04-14
ECE log into cloud ui toc update

eedugon:
Updating the TOC for ECE Log into the Cloud UI entry.

This was suggested some time ago by @shainaraskas. I believe this minor change might make sense to give better visibility to the page.

I've reviewed installation docs and landing pages and they properly link this doc, so the page doesn't really need to be inside `install ECE` in the TOC, and IMO it feels right between `deploy an orchestrator` and `manage deployments`

Part of: https://github.com/elastic/docs-projects/issues/466


shainaraskas (approved):
beaut!

Issue #1134, opened 2025-04-14
[REQUEST]: Rename "Elastic instance" to "Elastic Cloud resource"


shainaraskas:
### Description

renaming the generic "instances" umbrella term for ECH deployments + serverless projects to "Elastic Cloud resources" 

excludes instances that comprise a deployment

todo: 
* [x] update narrative docs in docs-content https://github.com/elastic/docs-content/pull/1136
* [x] check cloud tooling docs in cloud (done: no instances found)
* [x] check API docs
* [ ] update API doc strings

### Resources

see https://github.com/elastic/cloud/issues/139405

### Which documentation set does this change impact?

Elastic Cloud (Hosted and Serverless) only

### What release is this request related to?

N/A

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @alxchalkias 

## Action
added by (@georgewallace)
- Update the cloud api doc strings

shainaraskas:
https://www.elastic.co/docs/api/doc/cloud-billing/

^ some instances here

eedugon:
I've been recently working on billing docs, I'll give this a try soon.

eedugon:
Created https://github.com/elastic/docs-content/pull/5900 for docs-content updates.
I'll continue next week with the API docs per https://github.com/elastic/docs-content/issues/1134#issuecomment-2852215049

Update: I'm leaving this issue as this last step involves updating cloud API docs, and I think we don't have vault permissions for it.

PR #1136, opened 2025-04-14
Replace "Instance" with  "Elastic Cloud resource"

shainaraskas:
part of https://github.com/elastic/docs-content/issues/1134

the only place I see the term being used is in the context of instance access roles. this PR is pending the final strings being set

### DIFF
--- a/deploy-manage/users-roles.md
+++ b/deploy-manage/users-roles.md
@@ -84,7 +84,7 @@ You can't manage users and roles for {{eck}} clusters at the orchestrator level.
 serverless: all
 ```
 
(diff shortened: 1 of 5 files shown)


florent-leborgne:
👍 to my knowledge we've avoided using this term both in the UI and the docs so that should be it.
It's surfacing in RBAC and Billing UIs, but the billing (monitoring usage/consumption) docs don't go into that level of details.

wajihaparvez (approved):
lgtm!

bmureseanu (approved):

PR #1184, opened 2025-04-16
Add issue automation

bmorelli25:
Screenshot 2025-04-16 at 3 31 34 PM

### DIFF
--- a/.github/workflows/add-to-board.yml
+++ b/.github/workflows/add-to-board.yml
@@ -29,8 +29,8 @@ jobs:
           "team:Obs":        "649",
           "team:Security":   "1034",
           "team:Platform":   "1232",
-          "team:DocsEng":   "1625",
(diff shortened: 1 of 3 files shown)


.github/workflows/add-to-board.yml
```diff
@@ -0,0 +1,47 @@
+name: Triage & route issues
+
+on:
+  issues:
+    types: [labeled]
+
+permissions:
+  issues: write
+  contents: read
+  projects: write
```
reakaleek, line 10: unfortunately, `projects` doesn't exist. See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

We will need a PAT with org/projects write permissions.

KOTungseth (approved):

PR #1221, opened 2025-04-22
Call out explicitly that we don't provide option to disable MFA

kunisen:
## Description

As mentioned in an internal ticket - SDH CP 9470, disabling MFA poses a significant security risk, thus we don't provide this option. This doc PR is to call this part out explicitly clear. 

~After PR is merged, the orange part will be added:~ (updated - see below 1st image)

image

Updated as follows after discussion (stronger intro sentence instead of yet another callout): 

image


cloud-account/multifactor-authentication.md
```diff
@@ -20,6 +20,12 @@ Elastic recommends that you enable multiple methods so that you can still access
 
 If you use only a Google or Microsoft account to log in, then you can’t configure MFA in {{ecloud}}. You can check and manage your multifactor authentication options in your Google or Microsoft account security settings.
 
+::::{warning}
+
+Disabling MFA poses a significant security risk, and as such, we do not provide the option to disable this feature.
```
leemthompo, line 25:
```suggestion
Disabling MFA poses a significant security risk. Therefore, we do not provide the option to disable this feature.
```

leemthompo: wording nit

leemthompo: Maybe we should put this right at the top of the page? This will help anyone seeking to disable MFA eject from the doc immediately with this info and ideally we shouldn't stack admonitions because the page starts to get messy fast :)

florent-leborgne: I'd argue that we should make it **not** a warning and just tweak the 1st sentence of the page to make it even more explicit that MFA is mandatory

leemthompo: Also a valid approach @florent-leborgne 👍

florent-leborgne: i.e. something like:

~`If you use a password-based login, Elastic requires you to enable multifactor authentication (MFA) for added security on your {{ecloud}} account. You can choose from the following methods:`~ -> `Multifactor authentication (MFA) is mandatory when you log in to {{ecloud}} using a standard email/password combination. It helps protecting your account by adding an additional identity verification step when you log in. You can choose and define the MFA method to use based on your preference:`

maggieghamry: agreed (not sure if we may also want to note that Cloud MFA is enforced by default)

florent-leborgne: @kunisen when you get to see this, let me know if the above suggestion would work for you and I'll be happy to make the change. Thanks for calling out that this isn't explicit enough!

kunisen: @brunofarache since we are now actively discussing this in SDH, could you kindly comment on this? 
cc @delvedor since you also shared your opinion in the SDH comment. 
(Given this is the public repo I won't share any real link, but this is for SDH Control Plane 9470).

Based on our internal discussion, "SDH link + `/#issuecomment-2823720051`", we shouldn't offer this option (even we technically could). Also I think logically this can prevent users from having confusion on this by simply saying "it's not possible". 

Please kindly review and thanks! 🙏 

---

also thanks Liam, Florent and Maggie for the kind advice! 🙇

brunofarache: @kunisen I'm good with whatever you prefer, as long as Support is aware there are alternatives and won't be pointing to this doc as the final say about it.

florent-leborgne: Done in https://github.com/elastic/docs-content/pull/1221/commits/0e672d19492464190d2734e199a40b4b9bebe360

(The note about SMS was also outdated and a little bit misplaced so there's that change in there too)

florent-leborgne (approved):
LGTM after latest discussed changes

PR #1124, opened 2025-04-14
update JWT realm docs to include at+jwt support

richard-dennehy:
Update JWT documentation to mention `at+jwt` support introduced [here](https://github.com/elastic/elasticsearch/pull/126687)

Rendered [here](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1124/deploy-manage/users-roles/cluster-or-deployment-auth/jwt#jwt-validation-header)

PR #1229, opened 2025-04-22
Revert "update JWT realm docs to include at+jwt support"

shainaraskas:
Reverts elastic/docs-content#1124

need to wait for 9.1


leemthompo (approved):
Thanks for catching this and sorry for the noise

PR #1242, opened 2025-04-23
Add link to shared FS troubleshooting from system config page

kilfoyle:
This adds a tip on the [Important system configuration](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/important-system-configuration) page to the shared file system page, since users may encounter some difficulty with that specific scenario.

Requested by @tailorzed 

---

screen

### DIFF
--- a/deploy-manage/deploy/self-managed/important-system-configuration.md
+++ b/deploy-manage/deploy/self-managed/important-system-configuration.md
@@ -25,6 +25,10 @@ The following settings **must** be considered before going to production:
 For examples of applying the relevant settings in a Docker environment, refer to [](/deploy-manage/deploy/self-managed/install-elasticsearch-docker-prod.md).
 :::
 
+:::{tip}
+If you plan to use shared filesystem repositories, there are a few things to be aware of. Refer to [](/deploy-manage/tools/snapshot-and-restore/shared-file-system-repository.md) for guidance and troubleshooting tips.
+:::
+
 ::::{admonition} Use dedicated hosts
 :::{include} _snippets/dedicated-hosts.md
 :::


shainaraskas:
I don't think this note is targeted enough. Why do we need it on this page? What settings does it impact? The link doesn't really clarify this for me.

tailorzed:
hey @shainaraskas,

I could have used a more exact link, sorry.

I was trying to point out the note [here](https://www.elastic.co/guide/en/elasticsearch/reference/8.17/snapshots-filesystem-repository.html#_troubleshooting_a_shared_file_system_repository):

>Many NFS implementations match accounts across nodes using their numeric user IDs (UIDs) and group IDs (GIDs) rather than their names. It is possible for Elasticsearch to run under an account with the same name (often elasticsearch) on each node, but for these accounts to have different numeric user or group IDs. If your shared file system uses NFS then ensure that every node is running with the same numeric UID and GID, or else update your NFS configuration to account for the variance in numeric IDs across nodes.

Maybe this needs a permalink to be linked easier.

This issue could happen any time Customer uses shared filesystem (yes, this is not on us, but neither like setting swap :) ).

I hope this clear the intention!

kilfoyle:
> Why do we need it on this page? What settings does it impact?

@shainaraskas I gather that this page is about hardening for production and not just about specific settings, so it seems to me that users in that flow who are using NFS should be guided toward the caveats for that. However, I'm not very familiar with these docs. Please let me know if there's somewhere more appropriate for the tip and I'll move it over, and of course I can rephrase it as well. Btw, I tried linking to the anchor for the troubleshooting section but that didn't seem to work, so I just linked to the main page.

kilfoyle:
Ah, I see there's a Production guidance section, so maybe that would be a better fit.

shainaraskas:
I think this info needs to be as close to the info about the user that needs to run the node as possible, but we don't really have guidance about this in a central place (and very few pieces of info at all other than discouraging `root`). 

I agree that it probably belongs in this section (prod guidance is generally related to resilience, scaling, and performance), but we might need a page that talks about all of the requirements for the user that will run the service in the section. we could swap this to an issue to see if we can build out an mvp of that.

if we want to hack this in now, we need to specify that there are requirements specific to the NFS case for the user running elasticsearch, then link to the relevant section. I wonder if it might be worth making the nfs tip a prereq rather than a note in the troubleshooting section as well.

@eedugon let me know if any of this sounds incorrect

PR #1244, opened 2025-04-23
Creates UI copy request template

benironside:
Creates a UI copy request template for the docs-content repo. 

To preview the template, go to the code tab in this repo, switch to the `create-UI-copy-request-template` branch, then navigate to https://github.com/elastic/docs-content/blob/create-UI-copy-request-template/.github/ISSUE_TEMPLATE/ui-copy-request.yaml


.github/ISSUE_TEMPLATE/ui-copy-request.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+        We need a final draft of the new UI copy no later than December 5, ideally by December 1. No differences between serverless/ESS. 
+
+        This is an urgent request, it would be great to have updated copy for A and B ASAP, since the original copy is already in production. No differences between serverless/ESS. 
+    validations:
+      required: true
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for submitting this issue! Remember to contact a writer directly or to send a message in the #docs Slack channel.
```
jmikell821, line 99: I don't think we need the last part. 
```suggestion
        Thanks for submitting this issue! We'll be in contact shortly. 
```

.github/ISSUE_TEMPLATE/ui-copy-request.yaml
```diff
@@ -0,0 +1,99 @@
...
+    attributes:
+      label: Software version
+      description: Is this work targeted for a particular release version? If so, which one?
+      placeholder: |
+        For example:
+        "8.19.0"; 
+        "This feature has been in production since 8.15.0"; 
+        "N/A"
+    validations:
+      required: true
```
florent-leborgne, line 70: I think we can simplify this for UI copy and just request to provide the area. Because:
- "documentation set" isn't relevant in the context of UI copy
- "doc-set-differences", whether it's serverless or not isn't relevant in the context of UI copy
- "software-version" isn't relevant as we have a Timeline field already later in the form. Design tasks don't always have a release version set.

Instead we could request to specify the product area that is concerned to make sure the issue is routed to the right folks (we can discuss the options :D):

```suggestion
  - type: dropdown
    id: product-area
    attributes:
      label: Which product area does this mainly concern?
      options:
        - Cloud UI
        - Search solution
        - Observability solution
        - Security solution
        - Kibana Analytics (Discover, Dashboards...)
        - Kibana Management
        - Something else
      default: 0
    validations:
      required: true
```

florent-leborgne (commented):
Thanks for creating this!!

jmikell821 (approved):

PR #1246, opened 2025-04-24
Clarify replica shard allocation step in rolling restart procedure

rseldner:
Updates the restart documentation to specify that setting `cluster.routing.allocation.enable` to `primaries` disables **_replica_** shard allocation, not all shard allocation.

The phrase “disable shard allocation” seems to be a colloquial shorthand.  Those experienced with Elasticsearch may implicitly understand it refers to replica shards during restarts, but this can be unclear to newer users.

Historically (and TIL :bulb:), the restart procedures ([doc](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/rolling-upgrades.html)) used the `none` setting to fully disable shard allocation, which matched the “disable shard allocation” phrasing .  In version 6.7, the recommended setting changed to `primaries` to allow primary shard allocation while avoiding unnecessary replica movement.  However, the descriptive text was not updated to reflect this change, leaving behind a vestigial phrase from earlier behavior.

Clarifying this language helps align the documentation with the actual behavior in the example API calls we provide and reduces the risk of misinterpretation (e.g., mistakenly using "none" instead of "primaries").


ywangd (approved):
LGTM

shainaraskas (approved):

PR #1266, opened 2025-04-25
Update frontmatter in ingest-pipelines sibling pages

lcawl:
While fixing URLS in Kibana doc link service, I saw https://www.elastic.co/docs/manage-data/ingest/transform-enrich/ingest-pipelines-serverless and https://www.elastic.co/docs/manage-data/ingest/transform-enrich/ingest-pipelines seem to have incorrect applies_to frontmatter.  This PR fixes the frontmatter to match the titles. If they're later consolidated into a single page, the frontmatter can be updated again.


kilfoyle (approved):
LGTM! 🌵 
Thanks for catching that!

karenzone (approved):
LGTM

PR #1268, opened 2025-04-25
apm: remove TBS high RSS common problem

carsonip:
Removing APM Server TBS high RSS common problem as it was resolved in 8.14.2. Moreover, in 9.0, TBS consumes even less memory and disk IO thanks to a new storage layer.


colleenmcginnis (approved):
🎉

PR #1274, opened 2025-04-26
Add Base64 image loads as example in URL -> Image Format

nicpenning:
This is a common use case for rendering images when base64 encoded images are stored in Elasticsearch documents. Inspiration from a really old discuss post here: https://discuss.elastic.co/t/display-live-image-data-base64-jpeg-in-kibana/181554/2

Why not make it official?

Open to some ideas, but this PR is a POC of what that might look like in it's rawest form.

# Doing this:
url_image_base64_load_format

# Results in this:
elastic_logo_b64_rendered_in_discover


nicpenning:
@florent-leborgne Thoughts on this one?

florent-leborgne (approved):
Nice addition, thank you! I'll merge this one as is for now, and will probably try to reformat the page a little bit in a future update

PR #1283, opened 2025-04-28
Add limit on agentless deployments per project

kilfoyle:
This update the Agentless integrations page to note a limit on the number of agentless deployments per project, implemented by https://github.com/elastic/ingest-dev/issues/5326

Closes: https://github.com/elastic/ingest-docs/issues/1765

---

**For beta integrations:**

screen

---

**For GA integrations:**

screen2


nchaulet (approved):
🚀  Should also add some doc for non beta integration that are limited to 30 deployments per project?

nchaulet (approved):

PR #1290, opened 2025-04-28
[KNN] Adding default value for oversampling in 9.1.0

Samiul-TheSoccerFan:
Adding default value for `oversampling` into the knn documentation. We have added this value from `9.1.0`.


solutions/search/vector/knn.md
```diff
@@ -901,7 +901,7 @@ Approximate kNN search always uses the [`dfs_query_then_fetch`](https://www.elas
 
 When using [quantized vectors](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md#dense-vector-quantization) for kNN search, you can optionally rescore results to balance performance and accuracy, by doing:
 
-* **Oversampling**: Retrieve more candidates per shard.
+* **Oversampling**: Retrieve more candidates per shard. Starting in `9.1.0`, the default value for oversample is `3.0`.
```
jimczi, line 904: The default oversample is just for bbq so we should make it clear. I am not sure we need to mention the version as well, the docs are already versioned so that's fine.

Samiul-TheSoccerFan:
@jimczi, can I get another review please?

solutions/search/vector/knn.md
```diff
@@ -901,7 +901,7 @@ Approximate kNN search always uses the [`dfs_query_then_fetch`](https://www.elas
 
 When using [quantized vectors](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md#dense-vector-quantization) for kNN search, you can optionally rescore results to balance performance and accuracy, by doing:
 
-* **Oversampling**: Retrieve more candidates per shard.
+* **Oversampling**: Retrieve more candidates per shard. The default is `3.0` in `bbq`.
```
jimczi, line 904: It's a bit unclear. Maybe something like:
```
The default oversampling factor is 3, but only for the `bbq` quantization method.
Other quantization methods must explicitly specify an oversample value either in the field mapping or at query time.
```

jimczi (commented):
I left one comment

solutions/search/vector/knn.md
leemthompo: As of 9.0.0 we don't do page-level versioning anymore, so we need to call out changes more explicitly like the original @Samiul-TheSoccerFan's first draft.

I'm also asking myself "what was the default before?" now that we're adding info here :)

Samiul-TheSoccerFan: Before it would set to `0` if no value is provided, which means no oversample.

jimczi (approved):

leemthompo (approved):
Thanks. Let's keep wording the same for consistency.

Can you hold on merging and I'll take care of it? I just want to check with docs team on the latest plans for merging future-looking content.

solutions/search/vector/knn.md
```diff
@@ -913,7 +913,7 @@ All forms of quantization will result in some accuracy loss and as the quantizat
 
 * `int8` requires minimal if any rescoring
 * `int4` requires some rescoring for higher accuracy and larger recall scenarios. Generally, oversampling by 1.5x-2x recovers most of the accuracy loss.
-* `bbq` requires rescoring except on exceptionally large indices or models specifically designed for quantization. We have found that between 3x-5x oversampling is generally sufficient. But for fewer dimensions or vectors that do not quantize well, higher oversampling may be required.
+* `bbq` requires rescoring except on exceptionally large indices or models specifically designed for quantization. We have found that between 3x-5x oversampling is generally sufficient. But for fewer dimensions or vectors that do not quantize well, higher oversampling may be required. As noted above, we default to an oversampling factor of `3.0`.
```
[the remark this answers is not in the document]
Samiul-TheSoccerFan: Sure, that works. Thank you :)

Issue #437, opened 2025-02-13
wrong path: `C:\...\bin>bin\...`


kirisakow:
https://github.com/elastic/docs-content/blob/76aac68d066e2af935c38bca8ce04d3ee67a8dd9/deploy-manage/deploy/self-managed/install-elasticsearch-with-zip-on-windows.md#L190

https://github.com/elastic/docs-content/blob/76aac68d066e2af935c38bca8ce04d3ee67a8dd9/deploy-manage/deploy/self-managed/install-elasticsearch-with-zip-on-windows.md#L200

shainaraskas:
@eedugon are you able to confirm this?

PR #438, opened 2025-02-13
fix issue 437 (wrong path)

kirisakow:
fix #437

PR #1295, opened 2025-04-28
fixes and consolidates elasticsearch deploy on windows commands

eedugon:
Substitutes https://github.com/elastic/docs-content/pull/438

Closes https://github.com/elastic/docs-content/issues/437

It fixes a couple of wrong command execution paths and it also consolidates the command execution examples to `./bin/elasticsearch....` instead of `bin/elasticsearch...` (using ./ is safer in windows power shell and it's totally valid in Linux).

Also `./bin` is already used for a lot linux examples like starting Elasticsearch from the command line.


shainaraskas (approved):
ty, lgtm!

PR #1339, opened 2025-05-01
Create comment-on-removed-files.yml

bmorelli25:
## Summary

Per @colleenmcginnis's suggestion, adding a GH action to remind folks of the appropriate actions to take when files are removed (at least until we have real redirect support).


.github/workflows/comment-on-removed-files.yml
```diff
@@ -0,0 +1,64 @@
...
+            // Get all files in the PR
+            const { data: files } = await github.rest.pulls.listFiles({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: prNumber
+            });
+
+            // Filter for removed .md files
+            const removedMd = files
+              .filter(f => f.status === 'removed' && f.filename.endsWith('.md'))
```
colleenmcginnis, line 29: We might want to warn on both `removed` AND `renamed` since the URL is determined by the file path.

.github/workflows/comment-on-removed-files.yml
```diff
@@ -0,0 +1,64 @@
...
+              .map(f => `- \`${f.filename}\``);
+
+            if (removedMd.length > 0) {
+              // Prepare comment body
+              const body = [
```
colleenmcginnis, line 34: I've had trouble with multiline comments in the past, but this is an interesting approach! 🤞

.github/workflows/comment-on-removed-files.yml
```diff
@@ -0,0 +1,64 @@
...
+                removedMd.length === 1 
+                  ? "The following Markdown file was removed in this PR:" 
+                  : `The following ${removedMd.length} Markdown files were removed in this PR:`,
+                ...removedMd,
+                "",
+                "### Action Required",
+                "",
+                "We currently do not have an easy way to implement redirects for removed files. If possible, please:",
+                "",
+                "- Keep files whenever possible and exclude them from the TOC by using [`exclude`](https://docs-v3-preview.elastic.dev/elastic/docs-builder/tree/main/configure/content-set/navigation#exclude)",
```
colleenmcginnis, line 46: I think we might want to use `hidden` instead of `exclude`. https://elastic.github.io/docs-builder/configure/content-set/navigation/#hidden-files

.github/workflows/comment-on-removed-files.yml
bmorelli25: Good idea 👍

.github/workflows/comment-on-removed-files.yml
bmorelli25: Ah, yes. Definitely 👍

colleenmcginnis (approved):
🐻‍❄️

PR #1341, opened 2025-05-01
Comment specific preview links for changed files

bmorelli25:
### Summary

This PR adds a comment with links from each changed file to the built location in the deployment preview. This comment will be updated every time a new commit is pushed.

Screenshot 2025-05-01 at 3 09 29 PM


bmorelli25:
Unfortunately starting with a comment and updating when the deployment completes seems to need perms beyond the general GITHUB_TOKEN. I'm going to remove this functionality for now:

Screenshot 2025-05-01 at 2 38 14 PM

.github/workflows/doc-preview-comment.yml
```diff
@@ -0,0 +1,72 @@
...
+
+            if (!links.length) return;  // nothing to do
+
+            // 3) Build the comment body
+            const body = [
+              "🔍 **Preview links for changed docs:**",
+              "",
+              ...links,
+              "",
+              "🔔 *The preview site may take up to **5 minutes** to finish building. These links will become live once it completes.*"
```
Mpdreamz, line 48: 5 minutes? 😢 https://github.com/elastic/docs-content/actions/workflows/docs-build.yml 

Its a safe margin I get it 😄

Mpdreamz (approved):

.github/workflows/doc-preview-comment.yml
Mpdreamz: Just spotted the attempts to update on deployment created ❤️ that!

I think adding these permissions to the workflow would be sufficient. 

```
permissions:
  deployments: read
```

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#permissions

bmorelli25: oh snap! Thanks, I'll give this a try

PR #1368, opened 2025-05-05
Add documentation for failure stores.

jbaiera:
Adds a new section to the documentation to explain new failure store functionality.

Work in progress:
* Recipes section is currently TBD and open for suggestions. 
* Most links are not complete and need updating from "???".

### DIFF
--- a/manage-data/data-store/data-streams/failure-store.md
+++ b/manage-data/data-store/data-streams/failure-store.md
@@ -1,0 +1,607 @@
+---
+applies_to:
+  stack: ga 8.19.0
+  serverless: ga 9.1.0
+---
+
+# Failure store [failure-store]
+
+Failure stores are a secondary set of indices inside a data stream dedicated to storing failed documents. Failed documents are any documents that cause ingest pipeline exceptions or have a structure that conflicts with a data stream's mappings. These failures normally cause the indexing operation to fail, returning the error message in the response. When a data stream's failure store is enabled, these failures are instead captured and persisted to be analysed later, returning a successful response to the client in the meantime.
+
+## Set up a data stream failure store [set-up-failure-store]
+
+Each data stream has its own failure store that can be enabled to accept failures. By default, this failure store is disabled and any ingestion problems are raised in the response to write operations.
+
+### Set up for new data streams [set-up-failure-store-new]
+
+You can specify on a data stream's template if it should enable the failure store when it is first created. The `data_stream_options` field in a [template](../templates.md) contains the settings required to enable a data stream's failure store.
+
+:::{note}
+Unlike the `settings` and `mappings` fields on an [index template](../templates.md) which are repeatedly applied to new data stream write indices over time, the `data_stream_options` section of a template is applied to a data stream only once when the data stream is first created. To configure existing data streams, use the put data stream options API.
+:::
+
+To enable the failure store on a new data stream, enable it in the `data_stream_options` of the template:
+
+```console
+PUT _index_template/my-index-template
+{
+  "index_patterns": ["my-datastream-*"],
+  "data_stream": { },
+  "template": {
+    "data_stream_options": { <1>
+      "failure_store": {
+        "enabled": true <2>
+      }
+    }
+  }
+}
+```
+
+1. The options for a data stream to be applied at creation time.
+2. The failure store feature will be enabled for new data streams that match this template.
+
+
+After a matching data stream is created, its failure store will be enabled.
+
+### Set up for existing data streams [set-up-failure-store-existing]
+
+Enabling the failure store via [index templates](../templates.md) can only affect data streams that are newly created. Existing data streams that use a template will not apply any changes to the template's `data_stream_options` after they have been created.
+
+To modify an existing data stream's options, use the [put data stream options](???) API:
+
+```console
+PUT my-datastream-existing/_options
+{
+  "failure_store": {
+    "enabled": true <1>
+  }
+}
+```
+
+1. The failure store option will now be enabled.
+
+
+The failure store redirection can be suspended using this API as well. When the failure store is disabled, only failed document redirection is halted. Any existing failure data in the data stream will remain until removed by deletion or by retention.
+
+```console
+PUT my-datastream-existing/_options
+{
+  "failure_store": {
+    "enabled": false <1>
+  }
+}
+```
+
+1. Redirecting failed documents into the failure store will now be disabled.
+
+### Enable failure store via cluster setting [set-up-failure-store-cluster-setting]
+
+If you have a large number of existing data streams you may want an easier way to control if failures should be redirected. Instead of enabling the failure store using the [put data stream options](???) API, you can instead configure a set of patterns in the [cluster settings](???) which will enable the failure store feature by default.
+
+Configure a list of patterns using the `data_streams.failure_store.enabled` dynamic cluster setting. If a data stream matches a pattern in this setting and does not have the failure store explicitly disabled in its options, then the failure store will default to being enabled for that matching data stream.
+
+```console
+PUT _cluster/settings
+{
+  "persistent" : {
+    "data_streams.failure_store.enabled" : [ "my-datastream-*", "logs-*" ] <1>
+  }
+}
+```
+
+1. Indices that match `my-datastream-*` or `logs-*` will redirect failures to the failure store unless explicitly disabled.
+
+## Using a failure store [use-failure-store]
+
+The failure store is meant to ease the burden of detecting and handling failures when ingesting data to {{es}}. Clients are less likely to encounter unrecoverable failures when writing documents, and developers are more easily able to troubleshoot faulty pipelines and mappings.
+
+### Failure redirection [use-failure-store-redirect]
+
+Once a failure store is enabled for a data stream it will begin redirecting documents that fail due to common ingestion problems instead of returning errors in write operations. Clients are notified in a non-intrusive way when a document is redirected to the failure store.
+
+Each data stream's failure store is made up of a list of indices that are dedicated to storing failed documents. These indices function much like a data stream's normal backing indices: There is a write index that accepts failed documents, they can be rolled over, and are automatically cleaned up over time subject to a lifecycle policy.
+
+When a document bound for a data stream encounters a problem during its ingestion, the response is annotated with the `failure_store` field which describes how {{es}} responded to that problem. The `failure_store` field is present on both the [bulk](???) and [index](???) API responses when applicable. Clients can use this information to augment their behavior based on the response from {{es}}.
+
+Here we have a bulk operation that sends two documents. Both are writing to the `id` field which is mapped as a `long` field type. The first document will be accepted, but the second document would cause a failure because the value `invalid_text` cannot be parsed as a `long`. This second document will be redirected to the failure store: 
+
+```console
+POST my-datastream/_bulk
+{"create":{}}
+{"@timestamp": "2025-05-01T00:00:00Z", "id": 1234} <1>
+{"create":{}}
+{"@timestamp": "2025-05-01T00:00:00Z", "id": "invalid_text"} <2>
+```
+1. A correctly formatted document.
+2. Invalid document that cannot be parsed using the current mapping.
+
+```console-result
+{
+  "errors": false, <1>
+  "took": 400,
+  "items": [
+    {
+      "create": {
+        "_index": ".ds-my-datastream-2025.05.01-000001", <2>
+        "_id": "YUvQipYB_ZAKuDfZRosB",
+        "_version": 1,
+        "result": "created",
+        "_shards": {
+          "total": 1,
+          "successful": 1,
+          "failed": 0
+        },
+        "_seq_no": 3,
+        "_primary_term": 1,
+        "status": 201
+      }
+    },
+    {
+      "create": {
+        "_index": ".fs-my-datastream-2025.05.01-000002", <3>
+        "_id": "lEu8jZYB_ZAKuDfZNouU",
+        "_version": 1,
+        "result": "created",
+        "_shards": {
+          "total": 1,
+          "successful": 1,
+          "failed": 0
+        },
+        "_seq_no": 10,
+        "_primary_term": 1,
+        "failure_store": "used", <4>
+        "status": 201
+      }
+    }
+  ]
+}
+```
+
+1. The response code is 200 OK, and the response body does not report any errors encountered.
+2. The first document is accepted into the data stream's write index.
+3. The second document encountered a problem during ingest and was redirected to the data stream's failure store.
+4. The response is annotated with a field indicating that the failure store was used to persist the second document.
+
+
+If the document was redirected to a data stream's failure store due to a problem, then the `failure_store` field on the response will be `used`, and the response will not return any error information:
+
+```console-result
+{
+  "_index": ".fs-my-datastream-2025.05.01-000002", <1>
+  "_id": "lEu8jZYB_ZAKuDfZNouU",
+  "_version": 1,
+  "result": "created",
+  "_shards": {
+    "total": 1,
+    "successful": 1,
+    "failed": 0
+  },
+  "_seq_no": 11,
+  "_primary_term": 1,
+  "failure_store": "used" <2>
+}
+```
+
+1. The document for this index operation was sent to the failure store's write index.
+2. The response is annotated with a flag indicating the document was redirected.
+
+
+If the document could have been redirected to a data stream's failure store but the failure store was disabled, then the `failure_store` field on the response will be `not_enabled`, and the response will display the error encountered as normal.
+
+```console-result
+{
+  "error": {
+    "root_cause": [ <1>
+      {
+        "type": "document_parsing_exception",
+        "reason": "[1:53] failed to parse field [id] of type [long] in document with id 'Y0vQipYB_ZAKuDfZR4sR'. Preview of field's value: 'invalid_text'"
+      }
+    ],
+    "type": "document_parsing_exception",
+    "reason": "[1:53] failed to parse field [id] of type [long] in document with id 'Y0vQipYB_ZAKuDfZR4sR'. Preview of field's value: 'invalid_text'",
+    "caused_by": {
+      "type": "illegal_argument_exception",
+      "reason": "For input string: \"invalid_text\""
+    },
+    "failure_store": "not_enabled" <2>
+  },
+  "status": 400 <3>
+}
+```
+
+1. The failure is returned to the client as normal when the failure store is not enabled.
+2. The response is annotated with a flag indicating the failure store could have accepted the document, but it was not enabled.
+3. Status of 400 Bad Request due to the mapping problem.
+
+
+If the document was redirected to a data stream's failure store but that failed document could not be stored (e.g. due to shard unavailability or a similar problem), then the `failure_store` field on the response will be `failed`, and the response will display the error for the original failure, as well as a suppressed error detailing why the failure could not be stored:
+
+```console-result
+{
+  "error": {
+    "root_cause": [
+      {
+        "type": "document_parsing_exception", <1>
+        "reason": "[1:53] failed to parse field [id] of type [long] in document with id 'Y0vQipYB_ZAKuDfZR4sR'. Preview of field's value: 'invalid_text'",
+        "suppressed": [
+          {
+            "type": "cluster_block_exception", <2>
+            "reason": "index [.fs-my-datastream-2025.05.01-000002] blocked by: [FORBIDDEN/5/index read-only (api)];"
+          }
+        ]
+      }
+    ],
+    "type": "document_parsing_exception", <3>
+    "reason": "[1:53] failed to parse field [id] of type [long] in document with id 'Y0vQipYB_ZAKuDfZR4sR'. Preview of field's value: 'invalid_text'",
+    "caused_by": {
+      "type": "illegal_argument_exception",
+      "reason": "For input string: \"invalid_text\""
+    },
+    "suppressed": [
+      {
+        "type": "cluster_block_exception",
+        "reason": "index [.fs-my-datastream-2025.05.01-000002] blocked by: [FORBIDDEN/5/index read-only (api)];"
+      }
+    ],
+    "failure_store": "failed" <4>
+  },
+  "status": 400 <5>
+}
+```
+
+1. The root cause of the problem was a mapping mismatch.
+2. The document could not be redirected because the failure store was not able to accept writes at this time due to an unforeseeable issue.
+3. The complete exception tree is present on the response.
+4. The response is annotated with a flag indicating the failure store would have accepted the document, but it was not able to.
+5. Status of 400 Bad Request due to the original mapping problem.
+
+
+### Searching failures [use-failure-store-searching]
+
+Once you have accumulated some failures, they can be searched much like a regular index.
+
+:::{warning}
+Documents redirected to the failure store in the event of a failed ingest pipeline will be stored in their original, unprocessed form. If an ingest pipeline normally redacts sensitive information from a document, then failed documents in their original, unprocessed form may contain sensitive information.
+
+Furthermore, failed documents are likely to be structured differently than normal data in a data stream, and thus are not supported by [document level security](???) or [field level security](???).
+
+To limit visibility on potentially sensitive data, users require the [`read_failure_store`](???) index privilege for a data stream in order to search that data stream's failure store data.
+:::
+
+Searching a data stream's failure store can be done by making use of the existing search APIs available in {{es}}. 
+
+To indicate that the search should be performed on failure store data, use the [index component selector syntax](???) to indicate which part of the data stream to target in the search operation. Appending the `::failures` suffix to the name of the data stream indicates that the operation should be performed against that data stream's failure store instead of its regular backing indices.
+
+:::::{tab-set}
+
+::::{tab-item} {{esql}}
+```console
+POST _query?format=txt
+{
+    "query": """FROM my-datastream::failures | DROP error.stack_trace | LIMIT 1""" <1>
+}
+```
+1. We drop the `error.stack_trace` field here just to keep the example free of newlines.
+
+An example of a search result with the failed document present:
+
+```console-result
+       @timestamp       |    document.id     |document.index |document.routing|                                                            error.message                                                            |error.pipeline |error.pipeline_trace|error.processor_tag|error.processor_type|        error.type        
+------------------------+--------------------+---------------+----------------+-------------------------------------------------------------------------------------------------------------------------------------+---------------+--------------------+-------------------+--------------------+--------------------------
+2025-05-01T12:00:00.000Z|Y0vQipYB_ZAKuDfZR4sR|my-datastream  |null            |[1:45] failed to parse field [id] of type [long] in document with id 'Y0vQipYB_ZAKuDfZR4sR'. Preview of field's value: 'invalid_text'|null           |null                |null               |null                |document_parsing_exception
+```
+
+:::{note}
+Because the `document.source` field is unmapped, it is absent from the {{esql}} results. 
+:::
+
+::::
+
+::::{tab-item} _search API
+```console
+GET my-datastream::failures/_search
+```
+
+An example of a search result with the failed document present:
+
+```console-result
+{
+  "took": 0,
+  "timed_out": false,
+  "_shards": {
+    "total": 1,
+    "successful": 1,
+    "skipped": 0,
+    "failed": 0
+  },
+  "hits": {
+    "total": {
+      "value": 1,
+      "relation": "eq"
+    },
+    "max_score": 1,
+    "hits": [
+      {
+        "_index": ".fs-my-datastream-2025.05.01-000002", <1>
+        "_id": "lEu8jZYB_ZAKuDfZNouU",
+        "_score": 1,
+        "_source": {
+          "@timestamp": "2025-05-01T12:00:00.000Z", <2>
+          "document": { <3>
+            "id": "Y0vQipYB_ZAKuDfZR4sR",
+            "index": "my-datastream",
+            "source": {
+              "@timestamp": "2025-05-01T00:00:00Z",
+              "id": "invalid_text"
+            }
+          },
+          "error": { <4>
+            "type": "document_parsing_exception",
+            "message": "[1:53] failed to parse field [id] of type [long] in document with id 'Y0vQipYB_ZAKuDfZR4sR'. Preview of field's value: 'invalid_text'",
+            "stack_trace": """o.e.i.m.DocumentParsingException: [1:53] failed to parse field [id] of type [long] in document with id 'Y0vQipYB_ZAKuDfZR4sR'. Preview of field's value: 'invalid_text'
+	at o.e.i.m.FieldMapper.rethrowAsDocumentParsingException(FieldMapper.java:241)
+	at o.e.i.m.FieldMapper.parse(FieldMapper.java:194)
+	... 24 more
+Caused by: j.l.IllegalArgumentException: For input string: "invalid_text"
+	at o.e.x.s.AbstractXContentParser.toLong(AbstractXContentParser.java:189)
+	at o.e.x.s.AbstractXContentParser.longValue(AbstractXContentParser.java:210)
+	... 31 more
+"""
+          }
+        }
+      }
+    ]
+  }
+}
+```
+
+1. The document belongs to a failure store index on the data stream.
+2. The failure document timestamp is when the failure occurred in {{es}}.
+3. The document that was sent is captured inside the failure document. Failure documents capture the id of the document at time of failure, along with which data stream the document was being written to, and the contents of the document. The `document.source` fields are unmapped to ensure failures are always captured.
+4. The failure document captures information about the error encountered, like the type of error, the error message, and a compressed stack trace.
+::::
+
+::::{tab-item} SQL
+```console
+POST _sql?format=txt
+{
+    "query": """SELECT * FROM "my-datastream::failures" LIMIT 1"""
+}
+```
+
+An example of a search result with the failed document present:
+
+```console-result
+       @timestamp       |    document.id     |document.index |document.routing|                                                            error.message                                                            |error.pipeline |error.pipeline_trace|error.processor_tag|error.processor_type|                                                                                                                                                                                                                                                                            error.stack_trace                                                                                                                                                                                                                                                                            |        error.type        
+------------------------+--------------------+---------------+----------------+-------------------------------------------------------------------------------------------------------------------------------------+---------------+--------------------+-------------------+--------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------------
+2025-05-05T20:49:10.899Z|sXk1opYBL1dfU_1htCAE|my-datastream  |null            |[1:45] failed to parse field [id] of type [long] in document with id 'sXk1opYBL1dfU_1htCAE'. Preview of field's value: 'invalid_text'|null           |null                |null               |null                |o.e.i.m.DocumentParsingException: [1:45] failed to parse field [id] of type [long] in document with id 'sXk1opYBL1dfU_1htCAE'. Preview of field's value: 'invalid_text'
+	at o.e.i.m.FieldMapper.rethrowAsDocumentParsingException(FieldMapper.java:241)
+	at o.e.i.m.FieldMapper.parse(FieldMapper.java:194)
+	... 19 more
+Caused by: j.l.IllegalArgumentException: For input string: "invalid_text"
+	at o.e.x.s.AbstractXContentParser.toLong(AbstractXContentParser.java:189)
+	at o.e.x.s.AbstractXContentParser.longValue(AbstractXContentParser.java:210)
+	... 26 more
+|document_parsing_exception
+```
+
+:::{note}
+Because the `document.source` field is unmapped, it is absent from the SQL results.
+:::
+::::
+:::::
+
+Failure documents have a uniform structure that is handled internally by {{es}}.
+
+`@timestamp`
+:   (`date`) The timestamp at which the document encountered a failure in {{es}}.
+
+`document`
+:   (`object`) The document at time of failure. If the document failed in an ingest pipeline, then the document will be the unprocessed version of the document as it arrived in the original indexing request. If the document failed due to a mapping issue, then the document will be as it was after any ingest pipelines were applied to it.
+    
+    `document.id`
+    :   (`keyword`) The id of the original document at the time of failure.
+    
+    `document.routing`
+    :   (`keyword`, optional) The routing of the original document at the time of failure if it was specified.
+    
+    `document.index`
+    :   (`keyword`) The index that the document was being written to when it failed.
+
+    `document.source`
+    :   (unmapped object) The body of the document. This field is unmapped and unindexed to ensure failures are indexed reliably. If you need to include fields from the document source in your queries, use [runtime fields](???) on the search request.
+
+`error`
+:   (`object`) Information about the failure that prevented this document from being indexed.
+
+    `error.message`
+    :   (`match_only_text`) The error message that describes the failure.
+
+    `error.stack_trace`
+    :   (`text`) A compressed stack trace from {{es}} for the failure.
+
+    `error.type`
+    :   (`keyword`) The type classification of failure. Values are the same type returned within failed indexing API responses.
+
+    `error.pipeline`
+    :   (`keyword`, optional) If the failure occurred in an ingest pipeline, this will contain the name of the pipeline.
+
+    `error.pipeline_trace`
+    :   (`keyword`, optional array) If the failure occurred in an ingest pipeline, this will contain the list of pipelines that the document had visited up until the failure.
+
+    `error.processor_tag`
+    :   (`keyword`, optional) If the failure occurred in an ingest processor that is annotated with a [tag](???), the tag contents will be present here.
+
+    `error.processor_type`
+    :   (`keyword`, optional) If the failure occurred in an ingest processor, this will contain the processor type. (e.g. `script`, `append`, `enrich`, etc.)
+    
+
+## Manage a data stream's failure store [manage-failure-store]
+
+Failure data can accumulate in a data stream over time. To help manage this accumulation, most administrative operations that can be done on a data stream can be applied to the data stream's failure store.
+
+### Failure store rollover [manage-failure-store-rollover]
+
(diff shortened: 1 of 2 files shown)


manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,611 @@
...
+}
+```
+
+1. Redirecting failed documents into the failure store will now be disabled.
+
+### Enable failure store via cluster setting [set-up-failure-store-cluster-setting]
+
+If you have a large number of existing data streams you may want an easier way to control if failures should be redirected. Instead of enabling the failure store using the [put data stream options](./failure-store.md) API, you can instead configure a set of patterns in the [cluster settings](./failure-store.md) which will enable the failure store feature by default.
+
+Configure a list of patterns using the `data_streams.failure_store.enabled` dynamic cluster setting. If a data stream matches a pattern in this setting and does not have the failure store explicitly disabled in its options, then the failure store will default to being enabled for that matching data stream.
```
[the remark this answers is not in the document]
felixbarny: I was wondering that as well when reading this.

> then the failure store will default to being enabled for that matching data stream

Makes it sounds like it's not applying to existing data streams, just act as a default.

> If you have a large number of existing data streams [...] you can instead configure a set of patterns in the cluster settings

Makes it sound the setting is an alternative to enabling the failure store one by one via the DS options API if you have a large number of existing data streams.

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,611 @@
...
+    :   (`keyword`) The id of the original document at the time of failure.
+    
+    `document.routing`
+    :   (`keyword`, optional) The routing of the original document at the time of failure if it was specified.
+    
+    `document.index`
+    :   (`keyword`) The index that the document was being written to when it failed.
+
+    `document.source`
+    :   (unmapped object) The body of the document. This field is unmapped and unindexed to ensure failures are indexed reliably. If you need to include fields from the document source in your queries, use [runtime fields](./failure-store.md) on the search request.
```
felixbarny, line 412: > unindexed to ensure failures are indexed reliably.

I think I know what you mean but this sounds a bit confusing.

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,611 @@
...
+1. Information about the failure store is presented in the response under its own field.
+2. Indices are managed by data stream lifecycles by default.
+3. An effective retention period of thirty days (30d) is present by default.
+4. The retention is currently determined by the default.  
+
+:::{note}
+The default retention respects any maximum retention values. If [maximum retention](./failure-store.md) is configured lower than thirty days then the maximum retention will be used as the default value.
+:::
+
+You can update the default retention period for failure stores in your deployment by updating the `data_streams.lifecycle.retention.failures_default` cluster setting. Data streams that have no retention configured on their failure stores will use this value to determine their retention period.
```
felixbarny, line 532: Will this change the retention for existing data streams?

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,611 @@
...
+        }
+    }
+}
+```
+1. Ensure that the failure store remains enabled.
+2. Set only this data stream's failure store retention to ten days.
+
+### Add and remove from failure store [manage-failure-store-indices]
+
+Failure stores support adding and removing indices from them using the [modify data stream](./failure-store.md) API.
```
felixbarny, line 561: What happens if you add a failure store backing index that has incompatible mappings? Are we doing validations when adding a backing index or would it fail at runtime? Not suggesting one way is better than the other but maybe we should describe what happens here to set expectations.

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,611 @@
...
+2. The failure store feature will be enabled for new data streams that match this template.
+
+
+After a matching data stream is created, its failure store will be enabled.
+
+### Set up for existing data streams [set-up-failure-store-existing]
+
+Enabling the failure store via [index templates](../templates.md) can only affect data streams that are newly created. Existing data streams that use a template will not apply any changes to the template's `data_stream_options` after they have been created.
+
+To modify an existing data stream's options, use the [put data stream options](./failure-store.md) API:
```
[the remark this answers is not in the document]
jbaiera: Just a WIP tag - This is just a placeholder link for me to find and replace on later. I need to spend some time finding the right links for anything that has `./failure-store.md` as the link target.

manage-data/data-store/data-streams/failure-store.md
jbaiera: There's no special handling for mappings when adding an index to the failure store. You could add a completely unrelated index to the failure store and we allow it. Indices that are added to a data stream can never be treated as a write index, so we're less worried about their mappings than when doing a rollover operation. Even if the failure store is empty and we add a random index, the failure store is still marked for lazy rollover and will create a write index on redirection.

manage-data/data-store/data-streams/failure-store.md
jbaiera: Reworded a bit in 1591e7bb

manage-data/data-store/data-streams/failure-store.md
jbaiera: Yes, changed to "New and existing data streams that have no retention configured on their failure store will use this value..." in cec305f9

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,611 @@
...
+## Set up a data stream failure store [set-up-failure-store]
+
+Each data stream has its own failure store that can be enabled to accept failures. By default, this failure store is disabled and any ingestion problems are raised in the response to write operations.
+
+### Set up for new data streams [set-up-failure-store-new]
+
+You can specify on a data stream's template if it should enable the failure store when it is first created. The `data_stream_options` field in a [template](../templates.md) contains the settings required to enable a data stream's failure store.
+
+:::{note}
+Unlike the `settings` and `mappings` fields on an [index template](../templates.md) which are repeatedly applied to new data stream write indices over time, the `data_stream_options` section of a template is applied to a data stream only once when the data stream is first created. To configure existing data streams, use the put data stream options API.
```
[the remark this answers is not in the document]
jbaiera: Added placeholder link in f1cb727a

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,611 @@
+---
+applies_to:
+  stack: ga 8.19.0
+  serverless: ga 9.1.0
+---
+
+# Failure store [failure-store]
+
+Failure stores are a secondary set of indices inside a data stream dedicated to storing failed documents. Failed documents are any documents that cause ingest pipeline exceptions or have a structure that conflicts with a data stream's mappings. These failures normally cause the indexing operation to fail, returning the error message in the response. When a data stream's failure store is enabled, these failures are instead captured and persisted to be analysed later, returning a successful response to the client in the meantime.
```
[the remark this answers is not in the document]
jbaiera: Added in 28287fe4

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,611 @@
...
+
+## Using a failure store [use-failure-store]
+
+The failure store is meant to ease the burden of detecting and handling failures when ingesting data to {{es}}. Clients are less likely to encounter unrecoverable failures when writing documents, and developers are more easily able to troubleshoot faulty pipelines and mappings.
+
+### Failure redirection [use-failure-store-redirect]
+
+Once a failure store is enabled for a data stream it will begin redirecting documents that fail due to common ingestion problems instead of returning errors in write operations. Clients are notified in a non-intrusive way when a document is redirected to the failure store.
+
+Each data stream's failure store is made up of a list of indices that are dedicated to storing failed documents. These indices function much like a data stream's normal backing indices: There is a write index that accepts failed documents, they can be rolled over, and are automatically cleaned up over time subject to a lifecycle policy.
```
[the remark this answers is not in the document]
jbaiera: Updated in 08b2ac31

manage-data/data-store/data-streams/failure-store.md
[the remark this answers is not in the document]
jbaiera: Yeah, I tried to phrase this in a way that made it not seem like setting the property was the same as toggling the feature on permanently. Matching data streams only enable their failure stores so long as they are not explicitly disabled in the options, and only for as long as they match the setting. 

I've put up an edit that should simplify the explanation a bit. I also added an example of the explicit disabling of the failure store overriding the cluster setting.

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,631 @@
+---
+applies_to:
+  stack: ga 8.19.0
+  serverless: ga 9.1.0
+---
+
+# Failure store [failure-store]
+
+Failure stores are a secondary set of indices inside a data stream dedicated to storing failed documents. Failed documents are any documents that cause ingest pipeline exceptions or have a structure that conflicts with a data stream's mappings. These failures normally cause the indexing operation to fail, returning the error message in the response.
```
PeteGillinElastic, line 9: Nit: Maybe replace "These failures normally cause" with something like "Without the failure store, these failures would". I don't think it's obvious from your phrasing what "normally" means — like, maybe you're explaining the behaviour _with_ the failure store, since this is a page about failure store? — and it's better to be explicit. (I am aware of this because I've confused people before by saying things like this!)

PeteGillinElastic (commented):
This is looking good so far! I have one drive-by comment.

bvader:
Hi @jbaiera I noticed a small issue

https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1368/manage-data/data-store/data-streams/failure-store#use-failure-store

This example

POST my-datastream/_bulk
{"create":{}}
{"@timestamp": "2025-05-01T00:00:00Z", "id": 1234}
{"create":{}}
{"@timestamp": "2025-05-01T00:00:00Z", "id": "invalid_text"}
Should be add the -new otherwise it does not match the template at the top of the page and thus does not match and thus does not work :)

POST my-datastream-new/_bulk
{"create":{}}
{"@timestamp": "2025-05-01T00:00:00Z", "id": 1234}
{"create":{}}
{"@timestamp": "2025-05-01T00:00:00Z", "id": "invalid_text"}

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,631 @@
...
+
+When a document bound for a data stream encounters a problem during its ingestion, the response is annotated with the `failure_store` field which describes how {{es}} responded to that problem. The `failure_store` field is present on both the [bulk](./failure-store.md) and [index](./failure-store.md) API responses when applicable. Clients can use this information to augment their behavior based on the response from {{es}}.
+
+Here we have a bulk operation that sends two documents. Both are writing to the `id` field which is mapped as a `long` field type. The first document will be accepted, but the second document would cause a failure because the value `invalid_text` cannot be parsed as a `long`. This second document will be redirected to the failure store: 
+
+```console
+POST my-datastream/_bulk
```
bvader, line 129: Should be 

`POST my-datastream-new/_bulk` to match the data stream template at the top of the page.

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,2002 @@
+---
+applies_to:
+  stack: ga 8.19.0
+  serverless: ga 9.1.0
+---
+
+# Failure store [failure-store]
+
+Failure stores are a secondary set of indices inside a data stream dedicated to storing failed documents. Failed documents are any documents that cause ingest pipeline exceptions or have a structure that conflicts with a data stream's mappings. These failures normally cause the indexing operation to fail, returning the error message in the response.
```
kilfoyle, line 9:
```suggestion
A failure store is a secondary set of indices inside a data stream, dedicated to storing failed documents. A failed document is any document that, without the failure store enabled, would cause an ingest pipeline exception or that has a structure that conflicts with a data stream's mappings. In the absence of the failure store, a failed document would cause the indexing operation to fail, with an error message returned in the operation response.
```

I took a stab at rephrasing to incorporate Pete's comment above, and also to introduce the terms "failure store" and "failed document" as singular rather than plural, which is our usual convention, but it's just a suggestion. I think the original works nicely too.

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,2002 @@
...
+
+When a data stream's failure store is enabled, these failures are instead captured in a separate index and persisted to be analysed later. Clients receive a successful response with a flag indicating the failure was redirected. Failure stores do not capture failures caused by backpressure or document version conflicts. These failures are always returned as-is since they warrant specific action by the client.
+
+## Set up a data stream failure store [set-up-failure-store]
+
+Each data stream has its own failure store that can be enabled to accept failures. By default, this failure store is disabled and any ingestion problems are raised in the response to write operations.
+
+### Set up for new data streams [set-up-failure-store-new]
+
+You can specify on a data stream's template if it should enable the failure store when it is first created. The `data_stream_options` field in a [template](../templates.md) contains the settings required to enable a data stream's failure store.
```
kilfoyle, line 19:
```suggestion
You can specify in a data stream's [template](../templates.md) if it should enable the failure store when it is first created.
```
I'd put the link on the first instance of "template" and then we can remove the second sentence, since I think it's pretty similar to what you have above the example.

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,2002 @@
...
+
+1. The options for a data stream to be applied at creation time.
+2. The failure store feature will be enabled for new data streams that match this template.
+
+
+After a matching data stream is created, its failure store will be enabled.
+
+### Set up for existing data streams [set-up-failure-store-existing]
+
+Enabling the failure store via [index templates](../templates.md) can only affect data streams that are newly created. Existing data streams that use a template will not apply any changes to the template's `data_stream_options` after they have been created.
```
kilfoyle, line 50:
```suggestion
Enabling the failure store via [index templates](../templates.md) can only affect data streams that are newly created. Existing data streams that use a template are not affected by changes to the template's `data_stream_options` field.
```
Just a tweak. :-)

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,2002 @@
...
+
+# Failure store [failure-store]
+
+Failure stores are a secondary set of indices inside a data stream dedicated to storing failed documents. Failed documents are any documents that cause ingest pipeline exceptions or have a structure that conflicts with a data stream's mappings. These failures normally cause the indexing operation to fail, returning the error message in the response. 
+
+When a data stream's failure store is enabled, these failures are instead captured in a separate index and persisted to be analysed later. Clients receive a successful response with a flag indicating the failure was redirected. Failure stores do not capture failures caused by backpressure or document version conflicts. These failures are always returned as-is since they warrant specific action by the client.
+
+## Set up a data stream failure store [set-up-failure-store]
+
+Each data stream has its own failure store that can be enabled to accept failures. By default, this failure store is disabled and any ingestion problems are raised in the response to write operations.
```
kilfoyle, line 15:
```suggestion
Each data stream has its own failure store that can be enabled to accept failed documents. By default, this failure store is disabled and any ingestion problems are raised in the response to write operations.
```
Just for consistency with some of the other text.

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,2002 @@
...
+  "failure_store": {
+    "enabled": true <1>
+  }
+}
+```
+
+1. The failure store option will now be enabled.
+
+
+The failure store redirection can be disabled using this API as well. When the failure store is deactivated, only failed document redirection is halted. Any existing failure data in the data stream will remain until removed by manual deletion or by retention.
```
kilfoyle, line 66:
```suggestion
The failure store redirection can be disabled using this API as well. When the failure store is deactivated, only failed document redirection is halted. Any existing failure data in the data stream will remain until removed by manual deletion or by an expired data retention setting.
```
Just a tweak, but please ignore if I've got it wrong.

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,2002 @@
...
+
+Navigate to the data view page in Kibana and add a new data view. Set the index pattern to your failure store using the selector syntax.
+
+:::{image} /manage-data/images/elasticsearch-reference-management_failure_store_alerting_create_data_view.png
+:alt: create a data view using the failure store syntax in the index name
+:::
+::::
+
+::::{step} Create new rule
+Navigate to Management / Alerts and Insights / Rules. Create a new rule. Choose the Elasticsearch query option.
```
kilfoyle, line 1149:
```suggestion
Navigate to Management / Alerts and Insights / Rules. Create a new rule. Choose the {{es}} query option.
```

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,2002 @@
...
+    },
+    "failure_store": "not_enabled" <2>
+  },
+  "status": 400 <3>
+}
+```
+
+1. The failure is returned to the client as normal when the failure store is not enabled.
+2. The response is annotated with a flag indicating the failure store could have accepted the document, but it was not enabled.
+3. Status of 400 Bad Request due to the mapping problem.
```
kilfoyle, line 234:
```suggestion
3. The response status is `400 Bad Request` due to the mapping problem.
```
(just for consistency with the previous example)

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,2002 @@
...
+  },
+  "status": 400 <5>
+}
+```
+
+1. The root cause of the problem was a mapping mismatch.
+2. The document could not be redirected because the failure store was not able to accept writes at this time due to an unforeseeable issue.
+3. The complete exception tree is present on the response.
+4. The response is annotated with a flag indicating the failure store would have accepted the document, but it was not able to.
+5. Status of 400 Bad Request due to the original mapping problem.
```
kilfoyle, line 276:
```suggestion
5. The response status is `400 Bad Request` due to the original mapping problem.
```

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,2002 @@
...
+:   (`object`) Information about the failure that prevented this document from being indexed.
+
+    `error.message`
+    :   (`match_only_text`) The error message that describes the failure.
+
+    `error.stack_trace`
+    :   (`text`) A compressed stack trace from {{es}} for the failure.
+
+    `error.type`
+    :   (`keyword`) The type classification of failure. Values are the same type returned within failed indexing API responses.
```
kilfoyle, line 446:
```suggestion
    :   (`keyword`) The type classification of the failure. Values are the same type returned within failed indexing API responses.
```

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,2002 @@
...
+    ]
+  }
+}
+```
+1. The `document` field shows the state of the document is from before any pipeline executions.
+2. The pipeline failed after the timestamp would have been added.
+
+We can see that the document failed on the second processor in the pipeline. The first processor would have added a `@timestamp` field. Since the pipeline failed, we find that it has no `@timestamp` field added because it did not save any changes from before the pipeline failed.
+
+The second place failures can occur is during indexing. After the documents have been processed by any applicable pipelines, they are parsed using the index mappings before being indexed into the shard. If a document is sent to the failure store due to a failure in this process, then it will be stored as it was after any ingestion had occurred. This is because the original document is overwritten by the ingest pipeline changes by this point. This has the benefit of being able to see what the document looked like during the mapping and indexing phase of the write operation.
```
kilfoyle, line 604:
```suggestion
The second time when failures can occur is during indexing. After the documents have been processed by any applicable pipelines, they are parsed using the index mappings before being indexed into the shard. If a document is sent to the failure store due to a failure in this process, then it will be stored as it was after any ingestion had occurred. This is because, by this point, the original document has already been overwritten by the ingest pipeline changes. This has the benefit of allowing you to see what the document looked like during the mapping and indexing phase of the write operation.
```
Small tweak.

manage-data/data-store/data-streams/failure-store.md
kilfoyle: Looks good!

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+            "processor_type": "set" // It failed in the pipeline's set processor.
+          }
+        }
+      }
+    ]
+  }
+}
+```
+
+Despite not knowing the pipelines beforehand, we have some places to start looking. The `ingest-step-2` pipeline cannot find the `important.info` field despite it being present on the document that was sent to the cluster. If we pull that pipeline definition we find the following:
```
kilfoyle, line 96:
```suggestion
Despite not knowing the pipelines beforehand, we have some places to start looking. The `ingest-step-2` pipeline cannot find the `important.info` field despite it being present in the document that was sent to the cluster. If we pull that pipeline definition we find the following:
```
Global comment: I think 'in' makes more sense than 'on' when referring to the contents of a document.

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+    "successful": 1,
+    "failed": 0
+  },
+  "_seq_no": 1,
+  "_primary_term": 1,
+  "failure_store": "used"
+}
+```
+
+Upon checking the failure, we can quickly identify the tagged processor that caused the problem
```
kilfoyle, line 223:
```suggestion
On checking the failure, we can quickly identify the tagged processor that caused the problem.
```

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+        "tag": "copy to new counter again",
+        "field": "new_counter",
+        "copy_from": "counter"
+      }
+    }
+  ]
+}
+```
+
+We ingest some data and find that it was sent to the failure store
```
kilfoyle, line 197:
```suggestion
We ingest some data and find that it was sent to the failure store.
```

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+	at o.e.i.IngestDocument.getFieldValue(IngestDocument.java:202)
+	at o.e.i.c.SetProcessor.execute(SetProcessor.java:86)
+	... 14 more
+""",
+            "pipeline_trace": [
+              "complicated-processor"
+            ],
+            "pipeline": "complicated-processor",
+            "processor_type": "set", // Helpful, but which set processor on the pipeline could it be?
+            "processor_tag": "copy to new counter again" // The tag of the exact processor that it failed on.
```
kilfoyle, line 271:
```suggestion
            "processor_tag": "copy to new counter again" // The tag of the exact processor that the document failed on.
```

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+  "_version": 1,
+  "result": "created",
+  "_shards": {
+    "total": 1,
+    "successful": 1,
+    "failed": 0
+  },
+  "_seq_no": 2,
+  "_primary_term": 1,
+  "failure_store": "used" // The document was sent to the failure store
```
kilfoyle, line 33:
```suggestion
  "failure_store": "used" // The document was sent to the failure store.
```

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+    ]
+  }
+}
+```
+
+Without tags in place it would not be as clear where in the pipeline we encountered the problem. Tags provide a unique identifier for a processor that can be quickly referenced in case of an ingest failure.
+
+## Alerting on failed ingestion [failure-store-recipes-alerting]
+
+Since failure stores can be searched just like a normal data stream, we can use them as inputs to [alerting rules](./failure-store.md) in Kibana. Here is a simple alerting example to trigger on more than ten failures in the last five minutes for a data stream:
```
kilfoyle, line 284:
```suggestion
Since failure stores can be searched just like a normal data stream, we can use them as inputs to [alerting rules](./failure-store.md) in Kibana. Here is a simple alerting example that is triggered when more than ten indexing failures have occurred in the last five minutes for a data stream:
```
maybe this reads a wee bit better?

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+
+::::{warning}
+Care should be taken when replaying data into a data stream from a failure store. Any failures during the replay process may generate new failures in the failure store which can duplicate and obscure the original events.
+::::
+
+We recommend a few best practices for remediating failure data.
+
+**Separate your failures beforehand.** As described in the [failure document source](#use-failure-store-document-source) section above, failure documents are structured differently depending on when the document failed during ingestion. We recommend to separate documents by ingest pipeline failures and indexing failures at minimum. Ingest pipeline failures often need to have the original pipeline re-executed, while index failures should skip any pipelines. Further separating failures by index or specific failure type may also be beneficial.
```
kilfoyle, line 352:
```suggestion
**Separate your failures beforehand.** As described in the [failure document source](#use-failure-store-document-source) section above, failure documents are structured differently depending on when the document failed during ingestion. We recommend to separate documents by ingest pipeline failures and indexing failures at minimum. Ingest pipeline failures often need to have the original pipeline re-run, while index failures should skip any pipelines. Further separating failures by index or specific failure type may also be beneficial.
```
we generally avoid `executed` in favour of `run`.

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+
+**Perform a failure store rollover.** Consider rolling over the failure store before attempting to remediate failures. This will create a new failure index that will collect any new failures during the remediation process.
+
+**Use an ingest pipeline to convert failure documents back into their original document.** Failure documents store failure information along with the document that failed ingestion. The first step for remediating documents should be to use an ingest pipeline to extract the original source from the failure document and discard any other info on it.
```
kilfoyle, line 356:
```suggestion
**Use an ingest pipeline to convert failure documents back into their original document.** Failure documents store failure information along with the document that failed ingestion. The first step for remediating documents should be to use an ingest pipeline to extract the original source from the failure document and then discard any other information about the failure.
```
I'm not sure if I've got it right. I thought we could be explicit about what 'it' is in this case.

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+
+**Simulate first to avoid repeat failures.** If you must execute a pipeline as part of your remediation process, it is best to simulate the pipeline against the failure first. This will catch any unforeseen issues that may fail the document a second time. Remember, ingest pipeline failures will capture the document before an ingest pipeline was applied to it, which can further complicate remediation when a failure document becomes nested inside a new failure.
```
kilfoyle, line 358:
```suggestion
**Simulate first to avoid repeat failures.** If you must run a pipeline as part of your remediation process, it is best to simulate the pipeline against the failure first. This will catch any unforeseen issues that may fail the document a second time. Remember, ingest pipeline failures will capture the document before an ingest pipeline is applied to it, which can further complicate remediation when a failure document becomes nested inside a new failure.
```
rather than `will` capture and `was` applied, maybe we use future and present tense here.

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+}
+```
+1. This document is what we'll use for our simulations.
+2. It had a counter value.
+3. The document was missing a required field.
+4. The document failed in the `my-data-stream-default-pipeline`
+   ::::
+
+::::{step} Fix the original problem
+Because ingest pipeline failures need to be reprocessed by their original pipelines, any problems with those pipeline should be fixed before remediating failures. Investigating the pipeline mentioned in the example above shows that there is a processor that expects a field to be present that is not always present.
```
kilfoyle, line 469:
```suggestion
Because ingest pipeline failures need to be reprocessed by their original pipelines, any problems with those pipelines should be fixed before remediating failures. Investigating the pipeline mentioned in the example above shows that there is a processor that expects a field to be present that is not always present.
```

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+          "timestamp": "2025-05-01T20:58:03.566210529Z"
+        }
+      }
+    }
+  ]
+}
+```
+1. The index has been updated via the reroute processor.
+2. The id has stayed the same.
+3. The source should cleanly match what the original document should have been.
```
kilfoyle, line 636:
```suggestion
3. The source should cleanly match the contents of the original document.
```
Just a suggestion, as 'should have been' may be a bit open ended. :-)

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+```
+1. This document is what we'll use for our simulations.
+2. The document was missing a required `@timestamp` field.
+3. The document failed with a `document_parsing_exception` because of the missing timestamp.
+
+::::
+
+::::{step} Fix the original problem
+
+There are a broad set of possible indexing failures. Most of these problems stem from incorrect values for a particular mapping. Sometimes a large number of new fields are dynamically mapped and the maximum number of mapping fields is reached and no more can be added. In our example above, the document being indexed is missing a required timestamp.
```
kilfoyle, line 933:
```suggestion
There are a broad set of possible indexing failures. Most of these problems stem from incorrect values for a particular mapping. Sometimes a large number of new fields are dynamically mapped and the maximum number of mapping fields is reached, so no more can be added. In our example above, the document being indexed is missing a required timestamp.
```

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+}
+```
+1. Copy the original index name from the failure document over into the document's metadata. If you use custom document routing, copy that over too.
+2. Capture the source of the original document.
+3. Discard the `error` field since it wont be needed for the remediation.
+4. Also discard the `document` field.
+5. We extract all the fields from the original document's source back to the root of the document. The `@timestamp` field is not overwritten and thus will be present in the final document.
+
+:::{important}
+Remember that a document that has failed during indexing has already been processed by the ingest processor! It shouldn't need to be processed again unless you made changes to your pipeline to fix the original problem. Make sure that any fixes applied to the ingest pipeline is reflected in the pipeline logic here.
```
kilfoyle, line 973:
```suggestion
Remember that a document that has failed during indexing has already been processed by the ingest processor! It shouldn't need to be processed again unless you made changes to your pipeline to fix the original problem. Make sure that any fixes applied to the ingest pipeline are reflected in the pipeline logic here.
```

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1153 @@
...
+  "dest": {
+    "index": "my-datastream-indexing-example", <3>
+    "op_type": "create",
+    "pipeline": "my-datastream-remediation-pipeline" <4>
+  }
+}
+```
+1. Read from the failure store.
+2. Only reindex failure documents that match the ones we are replaying.
+3. Set the destination to the data stream the failures originally were sent to. The remediation pipeline above updates the index to be the correct one, but a destination is still required.
```
kilfoyle, line 1118:
```suggestion
3. Set the destination to the data stream the failures originally were sent to. The remediation pipeline in the example updates the index to be the correct one, but a destination is still required.
```

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,848 @@
...
+
+
+### Searching failures [use-failure-store-searching]
+
+Once you have accumulated some failures, they can be searched much like a regular index.
+
+:::{warning}
+Documents redirected to the failure store in the event of a failed ingest pipeline will be stored in their original, unprocessed form. If an ingest pipeline normally redacts sensitive information from a document, then failed documents in their original, unprocessed form may contain sensitive information.
+
+Furthermore, failed documents are likely to be structured differently than normal data in a data stream, and thus are not supported by [document level security](../../../deploy-manage/users-roles/cluster-or-deployment-auth/controlling-access-at-document-field-level.md#document-level-security) or [field level security](../../../deploy-manage/users-roles/cluster-or-deployment-auth/controlling-access-at-document-field-level.md#field-level-security).
```
slobodanadamovic, line 285: This is actually not true. We decided to support FLS and DLS against failure store. The main reason is because it would not be possible to prevent using FLS/DLS when users define implicit `read` access to backing `.fs*` indices. It felt wrong to prevent using FLS/DLS when users defined explicit `read_failure_store` access to data streams with FLS/DLS restrictions. Our biggest concern was that users would be expecting the DLS/FLS to stop certain docs/fields from being visible and it wouldn't. 

Right now, if users include FLS/DLS when granting access to failure store then we'll try to honour it. We should just make sure to highlight it (which you already did) that these documents are structured differently, and because they may contain sensitive information, the users should take extra care when defining access to them.

manage-data/data-store/data-streams/failure-store-recipes.md
```diff
@@ -0,0 +1,1154 @@
...
+  "throttled_millis": 0,
+  "requests_per_second": -1,
+  "throttled_until_millis": 0,
+  "failures": []
+}
+```
+1. The failures have been remediated.
+
+:::{tip}
+Since the failure store is enabled on this data stream, it would be wise to check it for any further failures from the reindexing process. Failures that happen at this point in the process may end up as nested failures in the failure store. Remediating nested failures can quickly become a hassle as the original document gets nested multiple levels deep in the failure document. For this reason, it is suggested to remediate data during a quiet period when no other failures are likely to arise. Furthermore, rolling over the failure store before executing the remediation would allow easier discarding of any new nested failures and only operate on the original failure documents.
```
dakrone, line 810: I think it would also be good to mention that reindexing the failed documents into the data stream does not remove the failures, so that needs to be done after the reindexing.

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,848 @@
+---
+applies_to:
+  stack: ga 8.19.0
```
dakrone, line 3: Can we put both 8.19 and 9.1 here? I want to make sure a user doesn't assume 9.0 > 8.19 and therefore should have this feature (even though it doesn't)

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,848 @@
...
+  serverless: ga 9.1.0
```
dakrone, line 4: Does Serverless have versions for the docs?

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,848 @@
...
+---
+
+# Failure store [failure-store]
+
+A failure store is a secondary set of indices inside a data stream, dedicated to storing failed documents. A failed document is any document that, without the failure store enabled, would cause an ingest pipeline exception or that has a structure that conflicts with a data stream's mappings. In the absence of the failure store, a failed document would cause the indexing operation to fail, with an error message returned in the operation response.
+
+When a data stream's failure store is enabled, these failures are instead captured in a separate index and persisted to be analysed later. Clients receive a successful response with a flag indicating the failure was redirected. Failure stores do not capture failures caused by backpressure or document version conflicts. These failures are always returned as-is since they warrant specific action by the client.
```
dakrone, line 11: I think it would be worth separating out the `Failure stores do not capture failures caused by backpressure or document version conflicts. These failures are always returned as-is since they warrant specific action by the client.` section into a "note" or "important" block, so it doesn't get missed.

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,848 @@
...
+## Set up a data stream failure store [set-up-failure-store]
+
+Each data stream has its own failure store that can be enabled to accept failed documents. By default, this failure store is disabled and any ingestion problems are raised in the response to write operations.
+
+### Set up for new data streams [set-up-failure-store-new]
+
+You can specify in a data stream's [index template](../templates.md) if it should enable the failure store when it is first created.
+
+:::{note}
+Unlike the `settings` and `mappings` fields on an [index template](../templates.md) which are repeatedly applied to new data stream write indices on rollover, the `data_stream_options` section of a template is applied to a data stream only once when the data stream is first created. To configure existing data streams, use the put [data stream options API](indices-put-data-stream-options).
```
dakrone, line 22: Will the `indices-put-data-stream-options` link work? Or do we need to fully-qualify it with the URL?

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,848 @@
...
+
+```console
+PUT _cluster/settings
+{
+  "persistent" : {
+    "data_streams.failure_store.enabled" : [ "my-datastream-*", "logs-*" ] <1>
+  }
+}
+```
+```console
```
dakrone, line 102: I think these two lines need to be removed to make this one continuous console block, since we use `<2>` below

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,848 @@
...
+1. The root cause of the problem was a mapping mismatch.
+2. The document could not be redirected because the failure store was not able to accept writes at this time due to an unforeseeable issue.
+3. The complete exception tree is present on the response.
+4. The response is annotated with a flag indicating the failure store would have accepted the document, but it was not able to.
+5. The response status is `400 Bad Request` due to the original mapping problem.
+
+
+### Searching failures [use-failure-store-searching]
+
+Once you have accumulated some failures, they can be searched much like a regular index.
```
dakrone, line 280:
```suggestion
Once you have accumulated some failures, the failure store can be searched much like a regular data stream.
```

sabarasaba:
@dakrone Any chance we might get this before FF for 8.19? We've made some work in the kibana UI and would be great if we can add a link to the docs

shainaraskas (approved):
hey @jbaiera! just adding a couple of comments to address the metadata of these pages / improve wayfinding between the pages. also approving to make sure that you're unblocked. let me know if you need a hand with anything.

manage-data/data-store/data-streams/failure-store.md
shainaraskas, line 22:
```suggestion
Unlike the `settings` and `mappings` fields on an [index template](../templates.md) which are repeatedly applied to new data stream write indices on rollover, the `data_stream_options` section of a template is applied to a data stream only once when the data stream is first created. To configure existing data streams, use the put [data stream options API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-put-data-stream-options).
```

shainaraskas: wonder if we need links to the serverless versions of these APIs too
https://www.elastic.co/docs/api/doc/elasticsearch-serverless/operation/operation-indices-put-data-stream-options

shainaraskas (commented):
to fix the failing test

slobodanadamovic (approved):

deploy-manage/users-roles/cluster-or-deployment-auth/elasticsearch-privileges.md
```diff
@@ -354,6 +354,9 @@ To learn how to assign privileges to a role, refer to [](/deploy-manage/users-ro
 `manage_data_stream_lifecycle`
 :   All [Data stream lifecycle](../../../manage-data/lifecycle/data-stream.md) operations relating to reading and managing the built-in lifecycle of a data stream. This includes operations such as adding and removing a lifecycle from a data stream.
 
+`manage_failure_store`
+:   All `monitor` privileges plus index and data stream administration limited to failure stores only. Only applies to data streams when accessing them via the [index component selector syntax](elasticsearch://reference/elasticsearch/rest-apis/api-conventions.md#api-component-selectors).
```
shainaraskas, line 358: temp fix to get CI working 

```suggestion
:   All `monitor` privileges plus index and data stream administration limited to failure stores only. Only applies to data streams when accessing them via the [index component selector syntax](elasticsearch://reference/elasticsearch/rest-apis/api-conventions.md).
```

deploy-manage/users-roles/cluster-or-deployment-auth/elasticsearch-privileges.md
```diff
@@ -383,6 +386,8 @@ To learn how to assign privileges to a role, refer to [](/deploy-manage/users-ro
 
     This privilege is not available in {{serverless-full}}.
 
+`read_failure_store`
+:   Read-only access to actions performed on a data stream's failure store. Required for access to failure store data (count, explain, get, mget, get indexed scripts, more like this, multi percolate/search/termvector, percolate, scroll, clear_scroll, search, suggest, tv). Only applies to data streams when accessing them via the [index component selector syntax](elasticsearch://reference/elasticsearch/rest-apis/api-conventions.md#api-component-selectors).
```
shainaraskas, line 390:
```suggestion
:   Read-only access to actions performed on a data stream's failure store. Required for access to failure store data (count, explain, get, mget, get indexed scripts, more like this, multi percolate/search/termvector, percolate, scroll, clear_scroll, search, suggest, tv). Only applies to data streams when accessing them via the [index component selector syntax](elasticsearch://reference/elasticsearch/rest-apis/api-conventions.md).
```

manage-data/data-store/data-streams/failure-store.md
```diff
@@ -0,0 +1,867 @@
...
+
+To limit visibility on potentially sensitive data, users require the [`read_failure_store`](../../../deploy-manage/users-roles/cluster-or-deployment-auth/elasticsearch-privileges.md#privileges-list-indices) index privilege for a data stream in order to search that data stream's failure store data.
+:::
+
+Searching a data stream's failure store can be done by making use of the existing search APIs available in {{es}}. 
+
+To indicate that the search should be performed on failure store data, use the [index component selector syntax](elasticsearch://reference/elasticsearch/rest-apis/api-conventions.md#api-component-selectors) to indicate which part of the data stream to target in the search operation. Appending the `::failures` suffix to the name of the data stream indicates that the operation should be performed against that data stream's failure store instead of its regular backing indices.
```
shainaraskas, line 311:
```suggestion
To indicate that the search should be performed on failure store data, use the [index component selector syntax](elasticsearch://reference/elasticsearch/rest-apis/api-conventions.md) to indicate which part of the data stream to target in the search operation. Appending the `::failures` suffix to the name of the data stream indicates that the operation should be performed against that data stream's failure store instead of its regular backing indices.
```

shainaraskas (commented):
removing links to anchors to get the PR green

PR #1377, opened 2025-05-07
Clarify Stack versioning on Understanding versioning page

natasha-moore-elastic:
Clarifies that the Stack `applies_to` badge applies to the stated version _and later_. Uses the <_version_>+ format for consistency with the rest of the page.


leemthompo (approved):
lgtm

PR #1384, opened 2025-05-08
Update ECK monitoring doc for connecting to an external monitoring elasticsearch cluster

laraMorenoIgle:
The example for "Connect to an external monitoring elasticsearch cluster" can lead to confusion or seem to be incorrect (because `secretName` is used only for metrics and not for logs), so I added a small explanation about what the config in the example does.

### DIFF
--- a/deploy-manage/monitor/stack-monitoring/eck-stack-monitoring.md
+++ b/deploy-manage/monitor/stack-monitoring/eck-stack-monitoring.md
@@ -116,7 +116,9 @@ If Logs stack monitoring is configured for a Beat, and custom container argument
 
 ## Connect to an external monitoring {{es}} cluster [k8s_connect_to_an_external_monitoring_elasticsearch_cluster]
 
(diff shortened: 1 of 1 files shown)


laraMorenoIgle (commented):
Small clarification on the example to send monitoring data to another cluster

deploy-manage/monitor/stack-monitoring/eck-stack-monitoring.md
```diff
@@ -143,7 +145,7 @@ spec:
 ```
 
 1. The `secretName` and `name` attributes are mutually exclusive, you have to choose one or the other.
-2. The `namespace` and `serviceName` attributes can only be used in conjunction with `name`, not with `secretName`.
+2. The `namespace` and `serviceName` attributes can only be used in conjunction with `name`, not with `secretName` and only to reference clusters managed by the same ECK instance.
```
eedugon, line 148:
```suggestion
2. The `namespace` and `serviceName` attributes can only be used in conjunction with `name`, not with `secretName`, and only to reference clusters managed by the same ECK instance.
```

eedugon (approved):
LGTM, thanks!!

PR #1388, opened 2025-05-08
Update autoscaling-in-ece-and-ech.md

essodjolo:
Punctuation, and space.


eedugon:
@shainaraskas : would you take a look at the entire paragraph?
The change suggested by @essodjolo makes all sense but we might take the chance to improve a bit more that big paragraph.

Otherwise this LGTM.

shainaraskas (approved):
I think this paragraph can stay as is for now, as it needs some restructuring of the page to accommodate a more elegant paragraph structure

PR #1394, opened 2025-05-09
Correct ECE doc KB reference

kunisen:
A small correction of a KB URL link.

cc @geekpete per our slack

FWIW, https://www.elastic.co/docs/release-notes/cloud-enterprise/known-issues has the correct link:
> In versions from 2.6 to 2.10 included, some or all platform certificates are generated with a 398-day expiration. Installations that ran on these versions, even temporarily, must have their certificates rotated manually before expiry. For details, check [our KB article](https://ela.st/ece-certificate-rotation).

Where

```
check [our KB article](https://ela.st/ece-certificate-rotation)
```


vishaangelova (approved):
I tested the new link, it works fine. LGTM.

PR #1406, opened 2025-05-11
Fix disabled dynamic mapping description in explicit mapping example

erfansaberi:
Hi. I updated the description for the dynamic: false setting in the "Define explicit mapping" section of the Elasticsearch documentation.

Previously, the documentation stated:

> Disables dynamic mapping for the index. Documents containing fields not defined in the mapping will be rejected.

This was misleading. In reality, when dynamic is set to false, fields not defined in the mapping are saved in the document but are not indexed or searchable. The document is still accepted.

The updated description now reads:

> Disables dynamic mapping for the index. Fields not defined in the mapping will be saved but not indexed and not searchable.

This change aligns the documentation with Elasticsearch's actual behavior and clarifies the difference between false and strict values. If the intent is to reject documents that contain unmapped fields, the dynamic: strict setting should be used instead.


solutions/search/elasticsearch-basics-quickstart.md
```diff
@@ -326,7 +326,7 @@ PUT /my-explicit-mappings-books
 }
 ```
 
-1. Disables dynamic mapping for the index. Documents containing fields not defined in the mapping will be rejected.
+1. Disables dynamic mapping for the index. Fields not defined in the mapping will still be saved in the document, but they won’t be indexed or searchable.
```
leemthompo, line 329:
```suggestion
1. Disables dynamic mapping for the index. Fields not defined in the mapping will still be stored in the document's `_source` field, but they won’t be indexed or searchable.
```

leemthompo (approved):
Nice catch, thanks for the fix @erfansaberi! I made one suggestion to help make your explanation more precise :)

erfansaberi:
Thanks for the great suggestion, @liamcr — just applied it!

PR #1412, opened 2025-05-14
Modify links and add repo to docset

theletterf:
This modified the links to use the new EDOT docs, incorporated inside reference.


vishaangelova (approved):
LGTM!

PR #1416, opened 2025-05-15
Fix example for creating realm.

rgshenoy:
Without the fully qualified path, user faces multiple errors when updating the user setting in ECH console. For example xpack.security is not directly editable in ECH, so when older example is directly used, it errors out. 

image


eedugon:
@rgshenoy , thanks for flagging this! I now see the bug. `authc` wasn't properly indented as part of `security`.

IMO the original structure looks a bit better than your proposal, but it's just a matter of taste.
I'm going to suggest a change in your proposal, which should also fix this. Let me know your thoughts.

shainaraskas:
> @rgshenoy , thanks for flagging this! I now see the bug. `authc` wasn't properly indented as part of `security`.
> 
> IMO the original structure looks a bit better than your proposal, but it's just a matter of taste.
> 
> Would you please update the PR to look like...
> 
> ```
> xpack:
>   security:
>     authc:
>       realms:
>         oidc:
>           oidc1:
>             order: 2
>             ...
> ```
> 
> If i'm not mistaken that should work in ECH without problems.
> 
> @shainaraskas , what do you think?


+1 I believe this is more consistent with our other docs as well

PR #1418, opened 2025-05-15
Update ece-manage-capacity for Podman limitation

frconil:
CPU quotas are not updated, and cannot be set to -1 (disable CPU hard limit) when the underlying container platform is podman.

cc @kunisen


eedugon:
thanks a lot for bringing this @frconil , and for sharing in private the reasoning behind not being able to customize the CPU quotas on Podman based installations.

As talked I'll tweak a bit the changes and share them with you before merging. I'd like to:

- Add a similar message in https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1418/deploy-manage/deploy/cloud-enterprise/resource-overrides

- Locate the important banner either at the beginning of the page or in the CPU section, but not in the memory section.

eedugon:
@kunisen , @frconil , I have a question about this change.

As far as I see in the private slack conversation shared by @frconil , what we want to explain in the docs is that disabling the CPU hard limit in Podman systems is not possible.

But we don't have any document that talks about CPU hard limit (which is something we can do only in the advanced editor).

Why are we proposing this change in the doc that talks about `Allocators capacity configuration`?
I was also thinking about the doc [Resource Overrides](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1418/deploy-manage/deploy/cloud-enterprise/resource-overrides), but that doc is focused on what the UI allows to override instances, and the UI doesn't show the CPU hard limit enable / disable option.

I'm totally ok with adding a message like:
`CPU quotas cannot be disabled (CPU hard limit) or updated when running ECE on Podman`.

But we need to ensure the message doesn't create confusion, as a reader running ECE on Docker might ask... and how can be disabled or updated in ECE running on Docker?

Any proposal to make this clearer?

Shouldn't this belong better (perhaps) to the Known Issues of ECE instead of the docs? Like explaining that CPU hard limit disabling doesn't really work on Podman systems?

eedugon:
Oh, I think the note is applicable to the `allocator capacity change` that @frconil was suggesting, because if the user changes the capacity of an allocator, the quotas won't be automatically updated for existing instances in Podman systems (is my understanding right?).

eedugon:
I'd like to merge first https://github.com/elastic/docs-content/pull/1511 and https://github.com/elastic/docs-content/pull/1507 before deciding final text on this one.

eedugon (approved):
LGTM

matt-elastic (approved):
LGTM

PR #1419, opened 2025-05-15
Update aliases.md

martinsbleu:


manage-data/data-store/aliases.md
```diff
@@ -209,8 +209,8 @@ PUT _index_template/my-index-template
 You can also specify index aliases in [create index API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-create) requests.
 
 ```console
-# PUT <my-index-{now/d}-000001>
-PUT %3Cmy-index-%7Bnow%2Fd%7D-000001%3E
+# Index with index aliases
+PUT <my-index-{now/d}-000001>
```
leemthompo, line 213: I don't think this a good suggestion because Kibana console requires URL-encoded characters for special syntax like angle brackets. The change would have replaced a working example (`PUT %3Cmy-index-%7Bnow%2Fd%7D-000001%3E`) with non-functional code (`PUT <my-index-{now/d}-000001>`). 

A better approach is to keep the URL-encoded (working) version while using clearer comments to explain what it represents. Perhaps something like:

```
# The following command uses URL encoding required by Kibana console.
# This is equivalent to the command: PUT <my-index-{now/d}-000001>
```

leemthompo: @kderusso can you keep me honest here? I'm not sure if we document this URL-encoding requirement anywhere and if that's a gap we should fill.

leemthompo (changes requested):

leemthompo:
Thanks for the contribution @martinsbleu! FYI it would be helpful to put a summary of the suggested changes in the PR description to help us understand the motivation :)

eedugon:
As far as I know, doing this in Kibana devTools console doesn't really work (at least with older versions):

```
PUT <my-index-{now/d}-000001>
```
That returns an error.

The way to create an index with the right date pattern in the name is:

```
PUT %3Cmy-index-%7Bnow%2Fd%7D-000001%3E
```

That returns:
```
{
  "acknowledged": true,
  "shards_acknowledged": true,
  "index": "my-index-2025.05.21-000002"
}
```

@martinsbleu : where are you able to do directly a `PUT <my-index-{now/d}-000001>` without encoding the characters?

manage-data/data-store/aliases.md
kderusso: Correct, the existing example works, but the updated example will return an error in the dev console: 

```
{
  "error": "no handler found for uri [/%3Cmy-index-%7Bnow/d%7D-000001%3E?pretty=true] and method [PUT]"
}
```

I like your suggestion to update the comment accordingly!

eedugon: ++ to that, your comment is a great idea. We need to show users that theoretically the call is `<my-index-{now/d}-000001>` but we have to encode the characters. That is also explained in some KBs already.

leemthompo: @martinsbleu if you'd like to keep the existing example but update the comment per my recommendation that would be a nice improvement

Issue #1391, opened 2025-05-08
[Issue]: Breaking Wrong Info on restoring system indices


stefnestor:
### Type of issue

Inaccurate

### What documentation page is affected

https://www.elastic.co/docs/manage-data/migrate/migrate-internal-indices

### What happened?

You're not allowed to restore system indices via `index` +8.0 you MUST use feature states. This (a) literally directs you wrong path & (b) uses renames pattern which will prevent those from working in destination cluster. 

### Additional info

- Example (`fleet`) how system indices are suppose to be restored: https://github.com/elastic/elasticsearch/issues/89261 
- Some feature states should be restored in conjunction with others. For example `kibana` and `security` can be restored by themselves but `fleet` has `kibana` dependencies and those should be restored together.

georgewallace:
@stefnestor Thanks again for this other piece of feedback, we will take a look at this

eedugon:
@stefnestor , thanks a lot for sharing this.

Quick question before preparing a fix:

> Some feature states should be restored in conjunction with others. For example kibana and security can be restored by themselves but fleet has kibana dependencies and those should be restored together.

Is there any doc where that is explained? I don't see any explanation in the doc where [feature states are introduced](https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore#feature-state), and neither in the [restore feature state](https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore/restore-snapshot#restore-feature-state) doc.

If you share more details about that we can also improve those docs highlighting that fact :)

stefnestor:
Cross-posting from internal : https://www.elastic.co/guide/en/elasticsearch/reference/8.17/snapshots-restore-snapshot.html#restore-feature-state

eedugon:
@shainaraskas : who could we contact from devs to get extra insights about restoring feature states so we can improve our document per this last comment? https://github.com/elastic/docs-content/issues/1391#issuecomment-3005449634

stefnestor:
@shahinakmal will you kindly CC me when you reach out to Dev?

FWIW I was wrong/incomplete you'd also need to consider `security`:
- From @lucabelluccini 's [article](https://support.elastic.dev/knowledge/view/954f0048) (sub-note https://github.com/elastic/ingest-docs/issues/388) 
    > A very high level overview:
    > - `.fleet*` for policies, packages, settings, etc...
    > - `.kibana*` for the dashboards and assets installed by the integrations
[...]

PR #1426, opened 2025-05-19
Manage data -> Migration guide to ECH or ECE updated

eedugon:
When working on the issue https://github.com/elastic/docs-content/issues/1391 I realized that the migration guide was a bit outdated and had multiple important issues to address.

I have refined the content and applied major updates.

Main docs updated:
- Migrate your Elasticsearch data
- Migrate system indices

Summary of fixes and changes:

Closes: https://github.com/elastic/docs-content/issues/1391


manage-data/migrate.md
```diff
@@ -107,86 +115,95 @@ Through the {{es}} reindex API, you can connect your new {{es}} Service deployme
...
+### Step 1: Set up the repository in the new cluster [migrate-repo-setup]
 
-2. Get the snapshot name:
+In this step, you’ll configure a snapshot repository in the new cluster that points to the storage location used by the old cluster. This allows the new cluster to access and restore snapshots created in the original environment.
 
-    ```sh
-    GET /_snapshot/NEW-REPOSITORY-NAME/_all
-    ```
+::::{tip}
+If your new {{ech}} or {{ece}} deployment cannot connect to the same repository used by your self-managed cluster, for example if it's a private NFS share, consider one of these alternatives:
```
shainaraskas, line 143:
```suggestion
If your new {{ech}} or {{ece}} deployment cannot connect to the same repository used by your self-managed cluster, for example if it's a private NFS share, consider one of the following alternatives:
```

manage-data/migrate.md
```diff
@@ -107,86 +115,95 @@ Through the {{es}} reindex API, you can connect your new {{es}} Service deployme
...
+    ```sh
+    GET /_snapshot/_all
     ```
 
+    Take note of the repository name and type (for example, `s3`, `gcs`, or `azure`), its base path, and any additional settings. Authentication credentials are often stored in the [secure settings](/deploy-manage/security/secure-settings.md) on each node. You’ll need to replicate all this configuration when registering the repository in the new ECH or ECE deployment.
+
+    If your old cluster has multiple repositories configured, identify the repository with the snapshots containing the data that you want to migrate.
 
-3. Add the snapshot repository:
+2. Add the snapshot repository on the new cluster:
```
shainaraskas, line 159: avoid two stacked introductory colons ("cluster:" "Considerations:")

```suggestion
2. Add the snapshot repository on the new cluster.
```

manage-data/migrate.md
```diff
@@ -107,86 +115,95 @@ Through the {{es}} reindex API, you can connect your new {{es}} Service deployme
...
 
-    ::::{tab-set}
+    The new cluster must register a snapshot repository that points to the same physical storage location used by the old cluster. This ensures the new cluster can access the existing snapshots.
```
shainaraskas, line 161: I think this is kind of a summary of the entire task (connecting the old snapshot repo to the new cluster) and therefore could be removed. let me know if this is serving a different purpose and I am misreading

manage-data/migrate.md
```diff
@@ -107,86 +115,95 @@ Through the {{es}} reindex API, you can connect your new {{es}} Service deployme
...
 
-    :::{tab-item} {{ech}}
+    Considerations:
 
-    From the [console](https://cloud.elastic.co?page=docs&placement=docs-body) of the **new** {{es}} cluster, add the snapshot repository.
+      * If you’re migrating [searchable snapshots](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md), the repository name must be identical in the source and destination clusters.
```
shainaraskas, line 165: is this true for a simple restore, or only if you want to restore search functionality?

manage-data/migrate.md
```diff
@@ -107,86 +115,95 @@ Through the {{es}} reindex API, you can connect your new {{es}} Service deployme
...
+      * If the old cluster still has write access to the repository, register the repository as read-only to avoid data corruption. This can be done using the `readonly: true` option.
 
-    For details, check our guidelines for:
-    * [Amazon Web Services (AWS) Storage](../deploy-manage/tools/snapshot-and-restore/ec-aws-custom-repository.md)
-    * [Google Cloud Storage (GCS)](../deploy-manage/tools/snapshot-and-restore/ec-gcs-snapshotting.md)
-    * [Azure Blob Storage](../deploy-manage/tools/snapshot-and-restore/ec-azure-snapshotting.md).
+    To configure a custom snapshot repository for your {{ech}} or {{ece}} deployment, follow the steps for the storage provider used by your existing repository:
```
shainaraskas, line 168: I think we can simplify this because most of this is clear from context, and it almost sounds like you're doing something brand new instead of just plugging the old thing and the new thing together. ignore this edit if you think it is not better I'm a little in the weeds

```suggestion
    To connect the existing snapshot repository to your new deployment, follow the steps for the storage provider where the repository is hosted:
```

manage-data/migrate.md
```diff
@@ -107,86 +115,95 @@ Through the {{es}} reindex API, you can connect your new {{es}} Service deployme
...
+        * [Create the repository](/deploy-manage/tools/snapshot-and-restore/ec-aws-custom-repository.md#ec-create-aws-repository)
+    * **Google Cloud Storage (GCS)**
+        * [Store credentials in the keystore](/deploy-manage/tools/snapshot-and-restore/ec-gcs-snapshotting.md#ec-configure-gcs-keystore)
+        * [Create the repository](/deploy-manage/tools/snapshot-and-restore/ec-gcs-snapshotting.md#ec-create-gcs-repository)
+    * **Azure Blob Storage**
+        * [Store credentials in the keystore](/deploy-manage/tools/snapshot-and-restore/ec-azure-snapshotting.md#ec-configure-azure-keystore).
+        * [Create the repository](/deploy-manage/tools/snapshot-and-restore/ec-azure-snapshotting.md#ec-create-azure-repository).
+    
+    ::::{important}
+    Although the previous instructions are focused on {{ech}}, you should follow the same steps for {{ece}} by configuring the repository directly **at the deployment level**.
```
shainaraskas, line 181: hmm, sounds like we should make it clear that this is possible in the snapshot docs. do you think it makes sense to open an issue to add new applies tags and notes with a use case?

shainaraskas, line 181: it's not really "previous" because you're still in the same section IMO.

```suggestion
    Although these instructions are focused on {{ech}}, you should follow the same steps for {{ece}} by configuring the repository directly **at the deployment level**.
```

manage-data/migrate.md
```diff
@@ -107,86 +115,95 @@ Through the {{es}} reindex API, you can connect your new {{es}} Service deployme
...
+    **Do not** configure the repository as an [ECE-managed repository](/deploy-manage/tools/snapshot-and-restore/cloud-enterprise.md), which is intended for automatic snapshots of deployments. In this case, you need to add a custom repository that already contains snapshots from another cluster.
+    ::::
 
-    :::{tab-item} {{ece}}
 
-    From the Cloud UI of the **new** {{es}} cluster add the snapshot repository.
+### Step 2: Run the snapshot restore [migrate-restore]
 
-    For details about configuring snapshot repositories on Amazon Web Services (AWS), Google Cloud Storage (GCS), or Azure Blob Storage, check [manage Snapshot Repositories](../deploy-manage/tools/snapshot-and-restore/cloud-enterprise.md).
+Once the repository has been registered and verified, you are ready to restore any data from any of its snapshots to your new cluster. You can do this using {{kib}} management UI, or directly with the {{es}} API.
```
shainaraskas, line 189: no manner is more direct than the other

you don't have API instructions here - consider linking to another resource that has them, or to the relevant endpoint in the API docs.

```suggestion
After the repository has been registered and verified, you are ready to restore any data from any of its snapshots to your new cluster. You can do this using {{kib}} management UI, or using the {{es}} API.
```

manage-data/migrate.md
```diff
@@ -107,86 +115,95 @@ Through the {{es}} reindex API, you can connect your new {{es}} Service deployme
...
 
-    If you’re migrating [searchable snapshots](../deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md), the repository name must be identical in the source and     destination clusters.
-    :::
+For extra details about the contents of a snapshot refer to [](/deploy-manage/tools/snapshot-and-restore.md#snapshot-contents).
```
shainaraskas, line 191:
```suggestion
For details about the contents of a snapshot, refer to [](/deploy-manage/tools/snapshot-and-restore.md#snapshot-contents).
```

manage-data/migrate.md
```diff
@@ -107,86 +115,95 @@ Through the {{es}} reindex API, you can connect your new {{es}} Service deployme
...
+4. Select the index or indices you wish to restore.
+5. Optionally, configure additional restore options, such as **Restore aliases**, **Restore global state**, or **Restore feature state**.
 
-    1. Open Kibana and go to **Management** > **Snapshot and Restore**.
-    2. Under the **Snapshots** tab, you can find the available snapshots from your newly added snapshot repository. Select any snapshot to view its details, and from there you can choose to restore it.
-    3. Select **Restore**.
-    4. Select the indices you wish to restore.
-    5. Configure any additional index settings.
-    6. Select **Restore snapshot** to begin the process.
+    Refer to [Restore a snapshot](/deploy-manage/tools/snapshot-and-restore/restore-snapshot.md) for more details about restore operations in {{es}}, including API based examples.
```
shainaraskas, line 201: consider moving this up so the link is closer to the 'you can do it this way or that way' statement

you can also say "this document explains how to blah blah in kibana. Refer to ... for ... API based examples."

manage-data/migrate.md
```diff
@@ -107,86 +115,95 @@ Through the {{es}} reindex API, you can connect your new {{es}} Service deployme
...
+    
+6. Select **Restore snapshot** to begin the process.
 
-5. Verify that the new index is restored in your deployment with this query:
+7. Verify that the new index is restored in your deployment with this query:
```
shainaraskas, line 205: could round this out to indicate that you could run it for each index you want to check

manage-data/migrate/migrate-internal-indices.md
```diff
@@ -7,109 +7,35 @@ applies_to:
...
   serverless: unavailable
 products:
   - id: cloud-hosted
 ---
 
-# Migrate internal indices
+# Migrate system indices
 
-When you migrate your {{es}} data into a new infrastructure you may also want to migrate your {{es}} internal indices, specifically the `.kibana` index and the `.security` index.
+When you migrate your {{es}} data into a new infrastructure you may also want to migrate your {{es}} system internal indices, specifically the `.kibana` index and the `.security` index.
```
shainaraskas, line 18:
```suggestion
When you migrate your {{es}} data into a new infrastructure, you might also want to migrate your {{es}} system internal indices, specifically the `.kibana` index and the `.security` index.
```

shainaraskas: Perhaps hint at what these indices contain or link to further info?

manage-data/migrate/migrate-internal-indices.md
```diff
@@ -7,109 +7,35 @@ applies_to:
...
 
-There are two ways to migrate the internal {{es}} indices:
+In {{es}} 8.0 and later versions, the snapshot and restore of [feature states](/deploy-manage/tools/snapshot-and-restore.md#feature-state) is the only way to back up and restore system indices and system data streams.
```
shainaraskas, line 20: consider making this more positive ("you can use")? maybe with a "this is the only available method" so people don't poke around for alternatives

manage-data/migrate/migrate-internal-indices.md
```diff
@@ -7,109 +7,35 @@ applies_to:
   deployment:
     eck: unavailable
     ess: ga
-    ece: unavailable
+    ece: ga
   serverless: unavailable
```
shainaraskas, line 11: same note about `unavailable`

imo, stack flavor is more about availability, deployment is more about applicability

manage-data/migrate/migrate-internal-indices.md
```diff
@@ -7,109 +7,35 @@ applies_to:
...
 
-There are two ways to migrate the internal {{es}} indices:
+In {{es}} 8.0 and later versions, the snapshot and restore of [feature states](/deploy-manage/tools/snapshot-and-restore.md#feature-state) is the only way to back up and restore system indices and system data streams.
 
-1. Reindex the indices from a remote cluster.
-2. Restore the indices from a snapshot.
+## Migrate system indices using snapshot and restore
 
-To reindex internal indices from a remote cluster, you can follow the same steps that you use to reindex regular indices when you [migrate your {{es}} data indices](../migrate.md#ech-reindex-remote).
+To restore system indices from a snapshot, follow the same procedure described in [](../migrate.md#ec-restore-snapshots) and select the appropriate **feature states** when preparing the restore operation, such as `kibana` or `security`.
```
shainaraskas, line 24: avoid bolding for emphasis

```suggestion
To restore system indices from a snapshot, follow the same procedure described in [](../migrate.md#ec-restore-snapshots) and select the appropriate feature states when preparing the restore operation, such as `kibana` or `security`.
```

manage-data/migrate/migrate-internal-indices.md
shainaraskas: consider also indicating that this method can be _paired_ with another restore method. indicating this makes this page make more sense as a standalone page

shainaraskas (approved):
beautiful and clear work as always! threw some nitpicks on there. I don't think there's anything major, but try to get that whitespace check notice to go away, and as long as @stefnestor gives this a green light, it's basically good to go

manage-data/migrate.md
eedugon: true for a simple restore. if the index being restored is actually a searchable snapshot, the only way to make it working in the destination is as a searchable snapshot where the snapshot is located in a repo with the same name, that's the reason for the message.

As far as I know (but I could be mistaken), there's no way to restore the data from a searchable snapshot index towards a standard index, which by the day would be a great feature and mechanism to convert a searchable snapshot index into a regular index, something I think we don't have today.

manage-data/migrate.md
eedugon: I like the concept of `custom repository` (that has always existed in ECH) because the platform comes with its own repositories (found-snapshot) for backups, and here the `user` is adding a snapshot repository at deployment level, now known by the platform.

If you find a better way of explaining this than using `custom repo` I'm totally opened for a change.

What we have never documented is doing the same in ECE. All our docs about snapshots in ECE rely on repositories defined at platform level (for automatic backups of deployments), which is the equivalent to the found-snapshots repos in ECH. Creating a custom snapshot repo at deployment level in ECE has never been documented, but it's perfectly possible (similar to ECH).

Let me know if what I've been trying to explain here is clear. About custom snapshot repo vs something else we can continue discussing, I agree `custom repository` is not intuitive at all, it's just what I've read for years in ECH docs.

manage-data/migrate.md
eedugon: do you mean configuring repositories at deployment level in ECE? Totally agreed, we should improve that visibility.
It's not a very common use case but a perfectly valid one, and I'm sure some customers are already doing that, as if for example they want a repo dedicated to searchable snapshots they have to create it at deployment level, and not managed by ECE.

We should create an issue to add this info and link from ECE docs to ECH docs + making sure the ECH docs in this are also applies_to ECE + explain it a bit.

manage-data/migrate.md
eedugon: yes, it feels duplicate, thanks

manage-data/migrate/migrate-internal-indices.md
```diff
@@ -5,111 +5,35 @@ mapped_pages:
...
 ---
 
-# Migrate internal indices
+# Migrate system indices
 
-When you migrate your {{es}} data into a new infrastructure you may also want to migrate your {{es}} internal indices, specifically the `.kibana` index and the `.security` index.
+When you migrate your {{es}} data into a new infrastructure, you might also want to migrate system-level indices and data streams, such as those used by {{kib}} or security features (for example, `.kibana` and `.security`).
 
-There are two ways to migrate the internal {{es}} indices:
+Starting in {{es}} 8.0, you can use [feature states](/deploy-manage/tools/snapshot-and-restore.md#feature-state) to back up and restore all system indices and system data streams. This is the only available method for migrating this data.
```
shainaraskas, line 18: I'd still possibly add a little info about how this can be paired with any method for restoring the rest of your data

shainaraskas (approved):
apart from some issues to be opened for future work and one tiny comment, ready to go

manage-data/migrate.md
eedugon: Created https://github.com/elastic/docs-content/issues/1494 as a follow up issue for this.

PR #1438, opened 2025-05-20
Make ESQL syntax reference link more visible

leemthompo:
Many folks used to the old docs where all ESQL docs were in one place, will often arrive here searching for the syntax reference.


florent-leborgne (approved):
Awesome

Issue #1393, opened 2025-05-08
[Issue]: Need to improve the inference documentation to articulate information around adaptive allocations.


georgewallace:
### Type of issue

None

### What documentation page is affected

Inference documentation

### What happened?

The cost implication of creating a custom inference endpoint without allowing adaptive allocations to scale down to 0 is not clearly explained in our documentation/UI. 

### Additional info

Additional email will be provided directly.

georgewallace:
@szabosteve please take a look at this. I will forward you more information, @ppf2 can also provide more information as this was raised in a support case.

kosabogi:
I started investigating this, and a few questions came up:

1. What exactly do we mean by custom inference endpoints? I initially assumed this referred to all endpoints created for services other than `ELSER` or `elasticsearch` - for example, endpoints that use third-party services. However, adaptive allocations don’t appear to be supported for those services. Could you please clarify that what we mean exactly by _custom_ here?

2. I'm considering updating [this](https://www.elastic.co/docs/explore-analyze/elastic-inference/inference-api#add-new-inference-endpoint) page. While testing, I found that if you create an inference endpoint via the UI, there’s no option to set `adaptive_allocations`. It also seems that you can’t enable adaptive allocations after creation either, since the `PUT /_inference/{inference_id}/_update` and `PUT /_inference/{task_type}/{inference_id}/_update` APIs don’t allow switching from static to adaptive allocation mode.

Can you confirm that the only way to configure adaptive allocations is to create the endpoint directly via the API? If so, I’d suggest adding a note to the docs to make this clearer on the “[Add new inference endpoint](https://www.elastic.co/docs/explore-analyze/elastic-inference/inference-api#add-new-inference-endpoint)” section.

cc @georgewallace @szabosteve @ppf2

szabosteve:
@kosabogi I think "custom" here means just that it's not a predefined endpoint (provided out-of-the-box), but something that the user themself creates.

kosabogi:
> [@kosabogi](https://github.com/kosabogi) I think "custom" here means just that it's not a predefined endpoint (provided out-of-the-box), but something that the user themself creates.

I see, thank you for the clarification!

PR #1454, opened 2025-05-22
Adds information about the importance of adaptive allocations

kosabogi:
### Description

This PR updates the Inference integration documentation to:

- Clearly state that not enabling adaptive allocations can result in unnecessary resource usage and higher costs.
- Expand the scope of the page to cover not only third-party service integrations, but also the Elasticsearch service.

### Related issue: https://github.com/elastic/docs-content/issues/1393

### DIFF
--- a/explore-analyze/elastic-inference/inference-api.md
+++ b/explore-analyze/elastic-inference/inference-api.md
@@ -9,15 +9,18 @@ products:
   - id: kibana
 ---
 
-# Integrate with third-party services
+# Inference integrations
 
-{{es}} provides a machine learning [inference API](https://www.elastic.co/docs/api/doc/elasticsearch/v8/operation/operation-inference-get-1) to create and manage inference endpoints to integrate with machine learning models provide by popular third-party services like Amazon Bedrock, Anthropic, Azure AI Studio, Cohere, Google AI, Mistral, OpenAI, Hugging Face, and more.
+{{es}} provides a machine learning [inference API](https://www.elastic.co/docs/api/doc/elasticsearch/v8/operation/operation-inference-get-1) to create and manage inference endpoints that integrate with services such as Elasticsearch (for built-in NLP models like [ELSER](/explore-analyze/machine-learning/nlp/ml-nlp-elser) and [E5](/explore-analyze/machine-learning/nlp/ml-nlp-e5)), as well as  popular third-party services like Amazon Bedrock, Anthropic, Azure AI Studio, Cohere, Google AI, Mistral, OpenAI, Hugging Face, and more.
 
-Learn how to integrate with specific services in the subpages of this section.
+You can create a new inference endpoint:
+
+- using the [Create an inference endpoint API](https://www.elastic.co/docs/api/doc/elasticsearch/v8/operation/operation-inference-put-1)
+- through the [Inference endpoints UI](#add-inference-endpoints).
 
 ## Inference endpoints UI [inference-endpoints]
 
-You can also manage inference endpoints using the UI.
+You can manage inference endpoints using the UI.
 
 The **Inference endpoints** page provides an interface for managing inference endpoints.
 
@@ -33,7 +36,7 @@ Available actions:
 * Copy the inference endpoint ID
 * Delete endpoints
 
-## Add new inference endpoint
+## Add new inference endpoint [add-inference-endpoints]
 
 To add a new interference endpoint using the UI:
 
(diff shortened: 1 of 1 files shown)


explore-analyze/elastic-inference/inference-api.md
```diff
@@ -42,18 +45,24 @@ To add a new interference endpoint using the UI:
 1. Provide the required configuration details.
 1. Select **Save** to create the endpoint.
 
+If your inference endpoint uses a model deployed in Elastic’s infrastructure, such as ELSER, E5, or a model uploaded through Eland, you can configure [adaptive allocations](#adaptive-allocations) to reduce resource usage and save costs.
```
szabosteve, line 48: It's not just about reducing resources and saving costs, it might be the opposite: spinning up more resources, which increases costs. It all depends on how resource-intensive the operation is. If it demands more computing power, we'll scale up – up until a certain limit. I'd rephrase this sentence, something like, "... you can configure adaptive allocations to dynamically adjust resources based on the demand of current operations", or something along those lines.

explore-analyze/elastic-inference/inference-api.md
```diff
@@ -42,18 +45,24 @@ To add a new interference endpoint using the UI:
...
+
+
 ## Adaptive allocations [adaptive-allocations]
 
 Adaptive allocations allow inference services to dynamically adjust the number of model allocations based on the current load.
+This feature is only supported for models deployed in Elastic’s infrastructure, such as ELSER, E5, or models uploaded through Eland. It is not available for third-party services like Alibaba Cloud, Cohere, or OpenAI.
```
szabosteve, line 54: It might be worth considering adding a short explanation why it's not available for third-party services (the models are not deployed in the cluster).

explore-analyze/elastic-inference/inference-api.md
```diff
@@ -42,18 +45,24 @@ To add a new interference endpoint using the UI:
...
 
 When adaptive allocations are enabled:
 
 * The number of allocations scales up automatically when the load increases.
 * Allocations scale down to a minimum of 0 when the load decreases, saving resources.
 
-For more information about adaptive allocations and resources, refer to the trained model autoscaling documentation.
+::::{warning}
+If you don't use adaptive allocations, the deployment will always use a fixed amount of resources, which can lead to unnecessary usage and higher costs.
```
szabosteve, line 62:
```suggestion
If you don't use adaptive allocations, the deployment will always consume a fixed amount of resources, regardless of actual usage. This can lead to inefficient resource utilization and higher costs.
```

szabosteve (commented):
It looks great! Left a couple of comments and suggestions.

kosabogi:
> It looks great! Left a couple of comments and suggestions.

Thank you! I applied your suggestions in my latest commit.

explore-analyze/elastic-inference/inference-api.md
```diff
@@ -9,15 +9,18 @@ products:
...
+
+- using the [Create an inference endpoint API](https://www.elastic.co/docs/api/doc/elasticsearch/v8/operation/operation-inference-put-1)
+- through the [Inference endpoints UI](#add-inference-endpoints).
 
 ## Inference endpoints UI [inference-endpoints]
 
-You can also manage inference endpoints using the UI.
+You can manage inference endpoints using the UI.
 
 The **Inference endpoints** page provides an interface for managing inference endpoints.
```
alaudazzi, line 25:
```suggestion
The **Inference endpoints** page provides an interface for managing inference endpoints.
```
The title and the second line of the paragraph are explicit enough. Perhaps we can remove the first line to avoid repeating "inference endpoints" three times over two lines of text.

alaudazzi (approved):
I left a minor suggestion, otherwise LGTM.

kosabogi:
Based on my conversation with Pius, I’ve updated my PR as follows:

- Removed the tables to avoid duplicating the content from the *Trained model autoscaling* page. Instead, I added links to the relevant sections so users can easily access the autoscaling settings guidance.
- Added a note to inform users about the implications of setting `min_number_of_allocations` greater than `0` when adaptive allocations are enabled.

@shubhaat  @arisonl I’d love to hear your thoughts - is this a good way to communicate how adaptive allocations work at this point in the guide?

cc @ppf2

explore-analyze/elastic-inference/inference-api.md
```diff
@@ -42,18 +43,33 @@ To add a new interference endpoint using the UI:
...
+
+The behavior of allocations depends on several factors:
+
+- Deployment type (Elastic Cloud Hosted, Elastic Cloud Enterprise, or Serverless)
+- Usage level (low, medium, or high)
+- Optimization type ([ingest](/deploy-manage/autoscaling/trained-model-autoscaling.md#ingest-optimized) or [search](/deploy-manage/autoscaling/trained-model-autoscaling.md#search-optimized))
+
+For more information about adaptive allocations and resources, refer to the [trained model autoscaling](/deploy-manage/autoscaling/trained-model-autoscaling.md) documentation.
+
+::::{note}
```
leemthompo, line 68: If we're delegating to the trained model page, should this note still live here or would it make sense to move that too?

kosabogi: Where do you think it would make more sense to place it?
IMO it should live here because it links to the Trained model autoscaling page, where all the autoscaling information is located.  But maybe I'm misunderstanding your comment? 🤔

shubhaat:
I think these changes are a good first step, eventually I think we want to hide some of these settings on serverless completely and simplify the UI and choices for customers. Very few users on serverless see the Trained models UI, so my guess is very few read the docs as well.

explore-analyze/elastic-inference/inference-api.md
leemthompo: - I agree it's good to have cost warning here. But I'd also expect this to be on the main page we're linking to, so I was just wondering if that duplication is OK.
- I'm not sure about the current placement of the note. 
  - Here's my thinking: The `::::{note}` goes into details about pricing implications but we've already sent readers to another page with "_For more information about adaptive allocations and resources, refer to the [trained model autoscaling](/deploy-manage/autoscaling/trained-model-autoscaling.md) documentation._" So the flow feels wrong. I'd argue for moving the cost note before the link. And also perhaps it should be `important` given it deals with real cost implications. 
- Nit: I think the wording could be more concise too.

kosabogi: Sorry, I thought you were referring to the sentence before the note - but it totally makes sense now.

My point is that we should include it here because the support ticket that raised this issue were about customers using inference services without realizing the cost implications.
Part of this is already mentioned on the trained model autoscaling page, specifically:

> Note: If you set the minimum number of allocations to 1, you will be charged even if the system is not using those resources.

In this case, I think the duplication is fine, as it's important to warn users about potential costs (according to the issue, this information was missing from this page.)

You're absolutely right that placing the note before that sentence interrupts the flow - I’ll move it and rework the wording a bit.

Thanks so much for your suggestions!

leemthompo (approved):
👍

Issue #1430, opened 2025-05-20
[Issue]: /etc/sysconfig/elasticsearch file referenced in Debian install instructions


jerrykan:
### Type of issue

Inaccurate

### What documentation page is affected

https://www.elastic.co/docs/deploy-manage/deploy/self-managed/install-elasticsearch-with-debian-package#deb-configuring

### What happened?

The following paragraph references the wrong file for Debian-base systems:

> The Debian package also has a system configuration file (/etc/sysconfig/elasticsearch), which allows you to set the following parameters:

The `/etc/sysconfig/elasticsearch` file is generally used on RedHat-based systems, while on Debian-based systems the equivalent file is `/etc/default/elasticsearch`.

eedugon:
good finding @jerrykan , thanks a lot.
Actually our old document has this right: https://www.elastic.co/guide/en/elasticsearch/reference/8.18/deb.html

We'll fix this soon.

PR #1457, opened 2025-05-22
Debian pkg config file updated in snippet

eedugon:
Minor change to fix the package configuration file in the Debian installation guide, which was incorrectly set to the RPM based file.

@shainaraskas : As the content came from a reused snippet we have lost the code formatting. Let me know your preference here. If you want to not use the `snippet` for this block of content or if we keep using it and we lose the code formatting.

If we use \`{{pkg-conf}}\` the substitution doesn't work.

I'm going to write a docs-builder issue to see if we can support inline code blocks with substitutions.

Closes https://github.com/elastic/docs-content/issues/1430

### DIFF
--- a/deploy-manage/deploy/self-managed/_snippets/etc-elasticsearch.md
+++ b/deploy-manage/deploy/self-managed/_snippets/etc-elasticsearch.md
@@ -4,7 +4,7 @@ The `setgid` flag applies group permissions on the `/etc/elasticsearch` director
 
 {{es}} loads its configuration from the `/etc/elasticsearch/elasticsearch.yml` file by default. The format of this config file is explained in [](/deploy-manage/deploy/self-managed/configure-elasticsearch.md).
 
-The {{distro}} package also has a system configuration file (`/etc/sysconfig/elasticsearch`), which allows you to set the following parameters:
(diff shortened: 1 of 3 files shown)


deploy-manage/deploy/self-managed/_snippets/etc-elasticsearch.md
```diff
@@ -4,7 +4,7 @@ The `setgid` flag applies group permissions on the `/etc/elasticsearch` director
 
 {{es}} loads its configuration from the `/etc/elasticsearch/elasticsearch.yml` file by default. The format of this config file is explained in [](/deploy-manage/deploy/self-managed/configure-elasticsearch.md).
 
-The {{distro}} package also has a system configuration file (`/etc/sysconfig/elasticsearch`), which allows you to set the following parameters:
+The {{distro}} package also has a system configuration file ({{pkg-conf}}), which allows you to set the following parameters:
```
shainaraskas, line 7: I don't mind the missing code formatting too much. you could do it like that, or alter it like this: 

````suggestion
The {{distro}} package also has a system configuration file at the following path:
 
```subs=true
{{pkg-conf}}
```

In this file, you can set the following parameters:
````

no strong preference.

shainaraskas (approved):
approved with a possible workaround to the code formatting thing you can take or leave

PR #1458, opened 2025-05-22
If we say the priority should be above 500 it should at least be 501

TimBosman:
If we say the priority should be above 500 it should at least be 501
[Current Doc page](https://www.elastic.co/docs/manage-data/data-store/templates)


kilfoyle (approved):
LGTM! Thanks @TimBosman!

PR #1459, opened 2025-05-22
Update minimum stack version to 8.0.0 for ECE 4

jakommo:
ECE 4 does only support Elastic Stack 8.0.0 and above, updating this to reflect this fact.

This should be back ported to the current live ECE 4.0 docs .


eedugon:
Thanks @jakommo , this looks perfect. Before merging I'd like to do a quick check over other ECE docs as we might want to change a similar statement in other places. But if you have already do so let me know and we can merge this straight away.

jakommo:
I did a quick search on this earlier today and could not find this mentioned anywhere else, but ++ on checking again because it feels like a rather important info for users upgrading.

shainaraskas (approved):

shainaraskas (approved):

eedugon (approved):

jakommo:
Thanks folks. @shainaraskas or @eedugon I merged it, is there anything else I need to do or will it automatically get deployed? 
Would be good if it could find it's way into the current ECE 4.0 docs

eedugon:
Change is live already: https://www.elastic.co/docs/deploy-manage/upgrade/orchestrator/upgrade-cloud-enterprise

There are periodic builds every 30 minutes from `main`.

PR #1492, opened 2025-05-26
Update breaking changes EA in Server mode

lucabelluccini:
Update the text to provide the solution.

We should add a link to https://www.elastic.co/docs/reference/fleet/install-elastic-agents#elastic-agent-servers-flavor-regular


theletterf (approved):
Thanks!

Issue #1460, opened 2025-05-22
Mismatch between query_delay documentation in old documentation and new documentation


herrBez:
The old documentation https://www.elastic.co/guide/en/machine-learning/8.18/ml-delayed-data-detection.html reports various sections including:
- Why worrying about delay data
- How we detect delayed data

While the new documentation https://www.elastic.co/docs/explore-analyze/machine-learning/anomaly-detection/ml-delayed-data-detection does not. I wonder if it's the result of the migration from asciidoc to markdown.

szabosteve:
@herrBez Thank you for opening the issue! Yes, the missing content is a result of an error during the migration. We'll add the content back soon.

PR #1497, opened 2025-05-27
Adds missing content to the 'Handling delayed data' page

kosabogi:
### Description

Due to an error during the migration, the _Handling delayed data_ page was missing some of its content. This PR restores the missing content.

### Related issue: https://github.com/elastic/docs-content/issues/1460


kilfoyle (approved):
LGTM! 🌵

PR #1506, opened 2025-05-28
Prepare stackpack download pages for 8.18.2 release.

gigerdo:
Adds 8.18.2 to the download pages. (Release date May 29)

Release issue: https://github.com/elastic/cloud/issues/141348


gigerdo:
> @gigerdo for now, we also have to maintain the stackpack list in [cloud-assets](https://github.com/elastic/cloud-assets/tree/master/docs) so the 3.8 docs can reflect current availability: https://www.elastic.co/guide/en/cloud-enterprise/3.8/ece-manage-elastic-stack.html

I didn't realize this download page is tied to the ECE version. Should we then actually remove all unsupported versions? The 4.0 download page currently lists all stackpacks <8, which aren't actually supported in 4.0.

PR #1507, opened 2025-05-28
ECE allocators capacity updated for svc restart

eedugon:
I have moved the page that describes how to change the allocators capacity, from "Prepare your environment" to "Configure ECE", as the page belongs more to that section than the preparations (it's impossible to follow the procedures describe in the page without an installed system).

Also the page has been updated to reflect that an `allocator service restart is needed` after changing the capacity with the API.

Preview:
- [Manage your allocators capacity](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1507/deploy-manage/deploy/cloud-enterprise/ece-manage-capacity)

@AlexP-Elastic , I'd like your review on this.

My main concern is if with `restart` you meant restarting the allocator service only or rebooting the host. I've assumed a restart of `frc-allocators-allocator` container only.

PS - I think we don't have any generic documentation about restarting services in ECE (which would be valid for podman & docker systems). Do you think we should have such a document? cc: @kunisen & @jakommo && @shainaraskas (we can discuss that in private to not deviate the attention of the PR).


eedugon:
Internal Jira reference: https://elasticco.atlassian.net/browse/CP-9161

deploy-manage/deploy/cloud-enterprise/ece-manage-capacity.md
```diff
@@ -31,17 +31,22 @@ ECE does not support hot-adding of resources to a running node.  When increasing
 ::::
 
 
-To adjust the allocator capacity, prior to ECE 3.5.0; it is necessary to reinstall ECE on the host with a new value assigned to the `--capacity` parameter. From ECE 3.5.0 onwards, just use the ECE API :
+To adjust the allocator capacity prior to ECE 3.5.0, you must reinstall ECE on the host with a new value assigned to the `--capacity` parameter. Starting with ECE 3.5.0, you can update the allocator capacity using the ECE API. After making this change, you must restart the allocator service for it to take effect.
```
shainaraskas, line 34: maybe this needs a sub-heading to separate the planning vs. adjusting task

suggestion: consider also pulling the pre-3.5.0 info into a note below the opening paragraph. reason being that a customer might decide to skip this paragraph if they're on 3.6+ . you could even merge the 3.5.0 note below up into a single note

also, you could share a link to the ECE API docs if you felt like it on "ECE API" :)

shainaraskas (approved):
some small feedback while we wait for dev validation :)

deploy-manage/deploy/cloud-enterprise/ece-manage-capacity.md
eedugon: great point, thanks

eedugon: https://www.elastic.co/docs/api/doc/cloud-enterprise/operation/operation-set-allocator-settings

PR #1513, opened 2025-05-28
Revert "Update `recore_vector` preview/GA status in knn.md"

leemthompo:
Reverts elastic/docs-content#1495

A [silly person](https://github.com/leemthompo) thought we already sanitized the unreleased version in the output, but we don't (yet) so this shouldn't have merged (yet).


leemthompo:
Lining this up for reversion when the time is right

PR #1549, opened 2025-05-30
[Observability] Add docs for enhanced logs component

mdbirnstiehl:
This PR closes [Issue 4888](https://github.com/elastic/observability-docs/issues/4888) and adds documentation for the enhanced application logs component, currently in technical preview.


alaudazzi (approved):
LGTM

mdbirnstiehl:
@weltenwort is this only available in stateful? I'm not seeing the setting in serverless, but I could be looking in the wrong place.

weltenwort:
hm, good question. I don't think we're intentionally hiding it from serverless. let me investigate...

weltenwort:
sounds reasonable 👍 thanks for catching this

PR #1576, opened 2025-06-02
clarify MFA for team use case

kunisen:
## Description

Clarify MFA for team use case by making it clear that the only secure and recommended way is to add members to the organization.

## Background & Motivation

Background is per https://github.com/elastic/sdh-control-plane/issues/9470#issuecomment-2820959159 and https://github.com/elastic/support-tech-lead/issues/1564, which I had with @brunofarache internally that we should clarify the docs to make it clear that the only recommended way is to add members to the org. 

## Before / After PR merged

:: Before 

https://www.elastic.co/docs/cloud-account/multifactor-authentication

image

:: After

Orange part will show up.

image


shainaraskas (approved):
approved with one suggestion. thanks for doing the work of checking with Alex and Bruno that this was safe to remove 👍

kunisen:
operational (ops) update - adding label. please ignore this message.

PR #1607, opened 2025-06-04
Limitation package policies per agent policy

lucabelluccini:
EA policy supports up to 10000 integration package policies as stated by https://github.com/elastic/kibana/issues/191308


lucabelluccini:
It applies to all stack versions of the past too.

theletterf:
@colleenmcginnis For things that affect all previous versions, I guess we don't need an apply_to tag, right?

PR #1620, opened 2025-06-05
[ECE] Clarify supported Ubuntu version

kunisen:
## Description

Previously we specify Ubuntu (Operating system / OS) version directly like 20.04 or 22.04. 
However, we later also 24.04, and maybe in the future we will have even more versions. 

All the compatible and supported version could be found in our support matrix - https://www.elastic.co/support/matrix#elastic-cloud-enterprise. 

But writing the version directly to public doc page other than support matrix seems to be easily forgotten to update, and thus we are left behind in public doc page that it does not mention 24.04 now. 
=> This causes confusion on users, because it's telling different in between public ECE docs and support matrix. 

This was brought up by @rheppe in an internal discussion [here](https://elastic.slack.com/archives/C045W6699NH/p1748919036870519?thread_ts=1748559251.195289&cid=C045W6699NH). 

The rationale is that support matrix is the source of truth, per what we have confirmed with @Kushmaro. So we would like to update the public docs based on support matrix. 

## Additional Notes

- We removed note `Docker 25 and higher are not compatible with ECE 3.7.` because the current branch of docs is for ECE 4.0, and we can see from support matrix, it's supported.
- We also explicitly mentioned that, the command mentioned in `sudo apt install -y docker-ce=5:24.0.* docker-ce-cli=5:24.0.* containerd.io` is just an example of install Docker 24.0 on Ubuntu. We added this because per support matrix, ECE 4.0 also supports other docker versions, and we can see from the screenshots below, it's hard to make an exclusive list in the public doc page, thus it makes more logical sense to say "it's a command example".

image

### Before

image

### After

No 20.04 or 22.04 version number present

image


kunisen:
It seems like this PR wrongly contains contents from other PRs / commits. I will close this one and file a new clean one. Sorry for the noise.

PR #1630, opened 2025-06-06
[ECE] Configure operating system instructions for Ubuntu with ECE 4.0

alstolten:
Hi, the following will update the docs about preparing an Ubuntu host for the installation of ECE 4.0. this concerns mostly to correct the docker version used, according to the Support Matrix here: https://www.elastic.co/support/matrix#ece

Edit: The current documentation uses Docker 24, which is not supported with ECE 4.0.

Edit2: I also changed the Ubuntu version to 24 LTS, which is the newest LTS version. Imho we should not provide instructions to install on 20 LTS anymore, as that has reached EOL, see https://ubuntu.com/about/release-cycle

### DIFF
--- a/deploy-manage/deploy/cloud-enterprise/configure-host-ubuntu.md
+++ b/deploy-manage/deploy/cloud-enterprise/configure-host-ubuntu.md
@@ -21,6 +21,7 @@ The following instructions show you how to prepare your hosts on 24.04 LTS (Nobl
 
 ::::{note}
 Please make sure to run the following steps with a user configured according to [Users and permissions](ece-users-permissions.md)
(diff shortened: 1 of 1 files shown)


kunisen:
@alstolten yes this was brought up by @rheppe last week so we discussed and decided to update our docs. After my PR is merged, I believe we should be fine. But please let me know if you have anything still in concern and we could discuss together.

benironside (approved):
I'm not familiar with the subject matter but the writing and formatting lgtm

shainaraskas:
I think that the changes [in @eedugon's PR](https://github.com/elastic/docs-content/pull/1621) address the issue that needed to be addressd, so I'm closing this PR. We can always reopen it if needed.

PR #1715, opened 2025-06-11
Doc: Remove note about agent certificate for control plane traffic

karenzone:
Note wrt agent cert for control plane traffic in https://www.elastic.co/docs/reference/fleet/fleet-agent-proxy-managed is no longer true.


karenzone:
Looks like this was fixed in https://github.com/elastic/elastic-agent/pull/4770 (May 16, 2024), so a similar doc fix should go back to 8.19. 8.18, and 8.17?

cmacknz (approved):

PR #1721, opened 2025-06-12
Bump ECE version to 4.0.1

kilfoyle:
Updates ECE docs version mentions to 4.0.1. Note that I've left "minimum required version" for all of the Docker lists at 4.0.0 as this shouldn't change as a result of the patch release.

Rel: https://github.com/elastic/release-eng/issues/1237
Rel: https://github.com/elastic/dev/issues/3176


deploy-manage/deploy/cloud-enterprise/default-system-deployment-versions.md
```diff
@@ -16,6 +16,7 @@ Note that since version 2.7.0, system deployments are automatically upgraded whe
 
 | {{ece}} version | Admin cluster | Logging & Metrics cluster | Security cluster |
 | --- | --- | --- | --- |
+| 4.0.1 | 8.18.0 | 8.18.0 | 8.18.0 |
```
eedugon, line 19: The system deployment versions associated to each release, are we sure they haven't changed? Do you know where is that information available? I'd like to learn it for future releases.

It would be nice to add the 3.8.1 in the table as we are also releasing 3.8.1 also, right? Anyway it's not very important.

eedugon:
@shainaraskas : I think we should work on a design with substitutions so we can bump the ECE version in an easier way in the future. What do you think?

deploy-manage/deploy/cloud-enterprise/default-system-deployment-versions.md
```diff
@@ -16,6 +16,7 @@ Note that since version 2.7.0, system deployments are automatically upgraded whe
...
 | 4.0.0 | 8.18.0 | 8.18.0 | 8.18.0 |
 | 3.8.0 | 8.17.4 | 8.17.4 | 8.17.4 |
```
kilfoyle, line 21:
```suggestion
| 3.8.1 | 8.17.4 | 8.17.4 | 8.17.4 |
| 3.8.0 | 8.17.4 | 8.17.4 | 8.17.4 |
```

deploy-manage/deploy/cloud-enterprise/default-system-deployment-versions.md
eedugon: @kilfoyle : it would be great if we get the information about where and how to check the system cluster versions that ECE expects (probably directly in the code or in an internal config file), so we could check this in the future directly ourselves.

The info of that table is really important, it was requested for years by support team and customers :)

PR #1729, opened 2025-06-13
Remove callout

theletterf:
Removes TA callout for Managed OTLP going GA for Serverless. On hold till 9.1 launches.


theletterf:
@raultorrecilla I believe the latest version already reflect those limitations. Could you check? https://github.com/elastic/opentelemetry/pull/403 (Limitations section)

Issue #1789, opened 2025-06-18
Update ECE docs release template and add a version variable


kilfoyle:
The last ECE patch release went less smoothly than we'd like. To fix this we should:
 - Have a variable for the ECE version.
 - Bring the ECE docs release issue template up to date.

PR #1790, opened 2025-06-18
Update ECE docs to use version variable

kilfoyle:
This adds an `ece_version` variable to make life a little easier for us on ECE release days. In addition to updating the variable we also need to append the [default system deployment versions](https://github.com/elastic/docs-content/blob/main/deploy-manage/deploy/cloud-enterprise/default-system-deployment-versions.md?expand=1) table.

I've updated the ECE docs release issue template accordingly, via https://github.com/elastic/dev/pull/3187.

Closes: https://github.com/elastic/docs-content/issues/1789

@eedugon  I know you mentioned that you were willing to do this, but since I had the release bump PRs in front of me I figured I'd just push the changes myself. Hope that's okay. :-)

### DIFF
--- a/deploy-manage/deploy/cloud-enterprise/ece-install-offline-images.md
+++ b/deploy-manage/deploy/cloud-enterprise/ece-install-offline-images.md
@@ -14,9 +14,9 @@ Versions of the {{stack}}, containing {{es}}, {{kib}}, and other products, are a
 
 The first table contains the stack versions that shipped with the 4.0 version of {{ece}}. You can also check the [most recent stack packs and Docker images](#ece-recent-download-list), which might have released after the 4.0 version of ECE, as well as the [full list of available stack packs and Docker images](#ece-full-download-list).
 
-| Docker images included with {{ece}} 4.0.1 |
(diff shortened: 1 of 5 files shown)


kilfoyle:
> For the system deployment versions reference file (not touched in this PR) it would be great to have it using this variable, at least for the latest release. Then whenever we prepare a new release we just need to add one line, while keeping the latest with the variable. What do you think? THe benefit will be that just searching for {{ece_version}} we will also find that file in the future.

It's a great idea, however I worry that it's a bit confusing that the line someone needs to add to the table would be for the previous release. At least, I would find that a bit unintuitive. I put a checklist item in the [docs ECE release template](https://github.com/elastic/dev/blob/75c3869c1316f14b905bb04c5146e9463aea8f32/.github/ISSUE_TEMPLATE/docs-ece-release.md?plain=1#L32) saying that we need to add a new line to the system deployment versions table, with a direct link. So that should help make sure people find it and they wouldn't need to search on the `ece_versions` variable. 

Also, I included a comment `This table is important so it's a good idea to confirm the system deployment versions with the ECE release coordinator` in the template, since you mentioned that the Support folks rely on it.

PR #1951, opened 2025-06-27
Removes applies_to tag for Serverless on Watcher page

yetanothertw:
It seems that the tag was added by mistake as according to the [comparison page](https://www.elastic.co/docs/deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings#elasticsearch), it's not avaialble in Serverless.


leemthompo:
This might be a good spot for the explicit `unavailable` tag? I can't remember the latest guidance on this though, so maybe I'm overthinking.

Issue #1569, opened 2025-05-30
Data lifecycle docs: Fix up Index management in Kibana


kilfoyle:
Feedback for [Index management in Kibana](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/index-management-in-kibana) (38:19):

 - [x] From Zoia: "This belongs right at the top of the ToC. Very important".
 - from Zoia: "Why is Enrich policies on that page? Maybe move this somewhere better."
    note: I think this should remain where it is. "Enrich policies" is documented there because it's one of the things manageable from the Index Management UI.

 - [x] There's a lot of overlap with [Index management](https://www.elastic.co/docs/manage-data/data-store/index-basics#index-management) so these pages should be combined, including checking which content applies to Serverless versus Stack-only.
 - [x] Once combined, we'll need a redirect for the one removed.
 - To Zoia's point, this page is super important. Maybe make "Index management" a top level page, below "Index basics". We can also have a link to it from somewhere in the lifecycle docs.

 - [x] Adding to the confusing-ness, there's a page "[Manage existing indices](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/manage-existing-indices)" which is not really about managing indices, but rather about migrating to ILM. Both this page and [Migrate index allocation filters to node roles](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/migrate-index-allocation-filters-to-node-roles) should be sub-pages of the Migrate to ILM page.

PR #1983, opened 2025-06-30
Nest two pages under 'Migrating to ILM'

kilfoyle:
Small fix to nest a couple of pages under the "Migrate to ILM" page. This is because without that context, "Manage existing indices" could be confused with the other two pages [Index management in Kibana](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/index-management-in-kibana) and [Index management](https://www.elastic.co/docs/manage-data/data-store/index-basics#index-management). I plan to combine those two latter pages as part of https://github.com/elastic/docs-content/issues/1569

---

screen26


yetanothertw (approved):
LGTM, thank you!

PR #1992, opened 2025-07-01
test

reakaleek:


leemthompo:
@reakaleek this isn't related to this test PR but one thing I've hit in `elasticsearch` repo is it's a bit greedy: if a README file ([example](https://github.com/elastic/elasticsearch/pull/130340)) or some other non-DocsV3 `.md` file ([example](https://github.com/elastic/elasticsearch/pull/130299#issuecomment-3017319593)) gets edited it will create a URL for that too (which obviously 404s)

So we might need some rules to avoid such cases. Lemme know if want me to open an issue or something :)

reakaleek:
> @reakaleek this isn't related to this test PR but one thing I've hit in `elasticsearch` repo is it's a bit greedy: if a README file ([example](https://github.com/elastic/elasticsearch/pull/130340)) or some other non-DocsV3 `.md` file ([example](https://github.com/elastic/elasticsearch/pull/130299#issuecomment-3017319593)) gets edited it will create a URL for that too (which obviously 404s)
> 
> So we might need some rules to avoid such cases. Lemme know if want me to open an issue or something :)

Did it already happen with the new implementation?

I'm reusing the result from tj-actions/changed-files

And this has the path-prefix ./docs/**

But I did not verify if it actually works.

leemthompo:
> Did it already happen with the new implementation?

no the cases I mentioned were using the old one, but just flagging for visibility :)

PR #1993, opened 2025-07-01
Fix permissions for comment

reakaleek:


reakaleek:
Force merging because not fixing this is blocking for other PRs.

PR #1999, opened 2025-07-02
Simplify workflow and remove secret dependency

reakaleek:


reakaleek:
This doesn't work as expected. In https://github.com/elastic/docs-content/actions/runs/16021553774/job/45199530388 the result is `false`.

I assume it's because visha's membership is private.

image

PR #2008, opened 2025-07-02
ECE redirect fix - APM Settings

eedugon:
Adds https://www.elastic.co/guide/en/cloud-enterprise/current/ece-apm-settings.html to the appropriate doc as a mapped_page.

This is part of the work for https://github.com/elastic/docs-projects/issues/494

In a parallel activity we will request the redirect of that page to https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/edit-stack-settings-apm


shainaraskas (commented):
can you explain a little more about this? the [mapped page](https://www.elastic.co/guide/en/cloud-enterprise/current/ece-apm-settings.html) has settings on it that are not present here. I know you mentioned there as some weirdness around settings that are no longer valid - is that the case here?

PR #2011, opened 2025-07-02
[KIBANA][9.x & Serv] Removes tech preview label from `alertuuid` action var 

nastasha-solomon:
Removed the tech preview label from the `alertuuid` action variable, as per request from @ymao1. 

Corresponding 8.x docs updated via: https://github.com/elastic/kibana/pull/226292


florent-leborgne (approved):
LGTM

PR #2012, opened 2025-07-02
ECE redirects - SAML sign outgoing messages

eedugon:
Adds https://www.elastic.co/guide/en/cloud-enterprise/current/ece-sign-outgoing-saml-message.html to the appropriate doc as a mapped_page.

This is part of the work for https://github.com/elastic/docs-projects/issues/494

In a parallel activity we will request the redirect of that page to https://www.elastic.co/docs/deploy-manage/users-roles/cluster-or-deployment-auth/saml#_configuring_es_for_signing

In this case I think the problem was caused by a typo (`_` instead of `-`). We already had this redirection and mapped_page configured, which is probably wrong:
https://www.elastic.co/guide/en/cloud-enterprise/current/ece_sign_outgoing_saml_message.html

I believe there's a typo and we could remove the redirection and the mapped_page of the underscored case. But I won't be doing it until I get confirmation from @colleenmcginnis or @shainaraskas .

@shainaraskas : I have also added a small update in the SAML doc as all the links about `custom bundles` need to be separated between ECH and ECE, because the procedures are completely different. let me know your thoughts and if you want to change the wording.

### DIFF
--- a/deploy-manage/deploy/cloud-enterprise/edit-stack-settings-apm.md
+++ b/deploy-manage/deploy/cloud-enterprise/edit-stack-settings-apm.md
@@ -2,6 +2,7 @@
 navigation_title: APM user settings
 mapped_pages:
   - https://www.elastic.co/guide/en/cloud-enterprise/current/ece-manage-apm-settings.html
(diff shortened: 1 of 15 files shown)


shainaraskas (approved):
changes lgtm

Issue #1714, opened 2025-06-11
[Internal]: ES|QL rule type ungrouped alerts


tiamliu:
### Description

What: we are introducing an ungrouped alerts option to configure alerts for ES|QL rule type
When: 9.1/8.19
Why: Update the ES query ES|QL rule type to allow users to generate an alert for each row in the query results, using a unique alert identifier

### Resources

GH: https://github.com/elastic/kibana/pull/212135

### Which documentation set does this change impact?

Elastic On-Prem and Cloud (all)

### Feature differences

The feature is identical in all deployment methods

### What release is this request related to?

9.1

### Serverless release

9.1

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @doakalexi @tiamliu 

**Stakeholders:**

PR #2015, opened 2025-07-02
[9.1 & Serverless][ResponseOps] ES|QL rule type ungrouped alerts

nastasha-solomon:
## Summary
Contributes to https://github.com/elastic/docs-content/issues/1714. 

In this PR, I made a couple of changes:
- Documented alert grouping options for the Elasticsearch query rules in Kibana (Stack Management) and Observability.
- Added `applies to` tags (Serverless and Stack) to the "Create an Elasticsearch query rule" page in the Observability docs.

**Corresponding 8.19 docs**
- Kibana: https://github.com/elastic/kibana/pull/226315

## Preview
[Here](https://github.com/elastic/docs-content/pull/2015#issuecomment-3029340713)

### DIFF
--- a/manage-data/data-store/index-basics.md
+++ b/manage-data/data-store/index-basics.md
@@ -158,7 +158,7 @@ In {{es-serverless}}, indices matching the `logs-*-*` pattern use the logsDB ind
 
 ### Manage index templates [index-management-manage-index-templates]
 
-An [index template](/manage-data/data-store/templates.md) is a way to tell {{es}} how to configure an index when it is created.
+An index template is a type of [template](/manage-data/data-store/templates.md) that tells {{es}} how to configure an index when it is created.
 
 Create, edit, clone, and delete your index templates in the **Index Templates** view. Changes made to an index template do not affect existing indices.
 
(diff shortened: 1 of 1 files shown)


tiamliu (approved):

explore-analyze/alerts-cases/alerts/rule-type-es-query.md
```diff
@@ -54,7 +54,9 @@ When you create an {{es}} query rule, your choice of query type affects the info
     | LIMIT 10
     ```
 
-2. If you use query DSL, KQL, or Lucene, set the group and theshold.
+2. Specify details for grouping alerts.
```
mdbirnstiehl, line 57: Maybe we could say something to make it more obvious that there are two options? It's a little hard to see that the time field is a different option especially with the `applies_to` tags. I'm not sure with the formatting if something like a sub unordered list would work? And the ordered list text could be something like:

```suggestion
2. Specify details for grouping alerts based on your query language.
```

mdbirnstiehl (approved):
LGTM!

PR #2026, opened 2025-07-03
Add Simplified Linear & RRF Retriever Examples

Mikep86:
Adds simplified `linear` and `rrf` retriever examples. Sibling PR to https://github.com/elastic/elasticsearch/pull/130559.


leemthompo:
@Mikep86 @kderusso

Question for the future: would it be better for the examples page to live in the elasticsearch repo eventually, as in a be a child to the reference page?

having the _overview_ page in the search solution/use case makes sense, but then having examples/reference separated feels wrong 🤷‍♂️ 

 this is an artifact of the docs migration...but we could re-home the page if it feels right

Mikep86:
@leemthompo I think it makes sense to make the examples page a child of the reference page

solutions/search/retrievers-examples.md
```diff
@@ -369,27 +381,156 @@ Which would return the following results:
...
+            }
+        ]
+    }
+}
+```
+
+::::
+
+
+## Example: RRF with the multi-field query format [retrievers-examples-rrf-multi-field-query-format]
```
kderusso, line 414: Reading through this again, I wonder if we should lead in with the simple way and then have the more complicated example below? Try to get more people to see the simplified retriever first? WDYT? 

Non blocking feedback

solutions/search/retrievers-examples.md
```diff
@@ -369,27 +381,156 @@ Which would return the following results:
...
+            }
+        ]
+    }
+}
+```
+
+::::
+
+We don't even need to specify the `fields` parameter when using the multi-field query format.
+If we omit it, the retriever will automatically query fields specified in the `index.query.default_field` index setting, which is set to `*` by default.
```
kderusso, line 481: 👍

solutions/search/retrievers-examples.md
leemthompo: ++

Mikep86: Done in https://github.com/elastic/docs-content/pull/2026/commits/56a19d4ef8235a02c9723a8b73a048c9891b1c86

Mikep86:
@elasticmachine update branch

leemthompo (approved):
Thanks @Mikep86!

PR #2035, opened 2025-07-04
Move EDOT Troubleshooting docs

theletterf:
This moves EDOT Troubleshooting docs to `docs-content`. After checking with @elastic/docs-engineering, it was decided that this was the only feasible way to have the docs within the Troubleshooting nav without significant effort.

### DIFF
--- a/troubleshoot/ingest/opentelemetry/edot-collector/index.md
+++ b/troubleshoot/ingest/opentelemetry/edot-collector/index.md
@@ -1,0 +1,30 @@
+---
+navigation_title: EDOT Collector
+description: Troubleshooting common issues with the EDOT Collector.
+applies_to:
+  stack:
+  serverless:
+    observability:
+products:
+  - id: cloud-serverless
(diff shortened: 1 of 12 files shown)


troubleshoot/ingest/opentelemetry/edot-collector/index.md
```diff
@@ -0,0 +1,30 @@
+---
+navigation_title: EDOT Collector
+description: Troubleshooting common issues with the EDOT Collector.
+applies_to:
+  stack:
```
reakaleek, line 5: Q: Why does these pages also apply to stack?

theletterf: EDOT can be used with Stack deployments.

reakaleek: Ahh, make sense.

alexandra5000 (approved):

PR #2038, opened 2025-07-04
apm: Update config documentation and examples to be consistent

carsonip:
- **Fix config prefix**
- **Add missing apm-server prefix to agent auth**
- **Update example config for TBS**
- **Update agent auth example**
- **Update TLS version**


florent-leborgne (approved):
LGTM!

PR #2052, opened 2025-07-07
Fix incorrect link to SAML Entra page

richard-dennehy:
The SAML Authentication page is intended to link to the SAML Entra guide, but this link currently goes to the same SAML Authentication page


kilfoyle (approved):
LGTM! 🚀 
Nice catch. Thanks @richard-dennehy!

PR #2057, opened 2025-07-07
Add troubleshooting page for propagating client metadata with EDOT

alexandra5000:
Added a new troubleshooting page that explains why `from_context: client_metadata` cannot extract application-level attributes like `project_id` or `tenant`. The page clarifies the distinction between transport-level and application-level metadata and provides a resolution using SDK instrumentation.


troubleshoot/ingest/opentelemetry/edot-collector/metadata.md
```diff
@@ -0,0 +1,81 @@
...
+    edot_collector: ga
+products:
+  - id: cloud-serverless
+  - id: observability
+  - id: edot-collector
+---
+
+# EDOT Collector doesn’t propagate client metadata
+
+If your EDOT Collector doesn’t propagate metadata like `project_id`, `tenant`, or `environment`, you may be misunderstanding what “client metadata” means in Elastic Distributions of OpenTelemetry (EDOT).
```
bmorelli25, line 18: The current wording implies user error (“you may be misunderstanding”). Could we reframe this to highlight the Collector’s behavior instead? That keeps the tone neutral and more helpful.

bmorelli25: I like the latter option. I would lead with the default though:

> By default, the Collector only propagates transport-level metadata. If you want the EDOT Collector to propagate metadata like `project_id`, `tenant`, or `environment`, you need a specific SDK instrumentation.

1pkg (approved):

PR #2069, opened 2025-07-08
fix example config of local_dynamic provider

andrzej-stencel:
The example configuration was invalid. I copied correct configuration from the [provider's docs](https://www.elastic.co/docs/reference/fleet/local-dynamic-provider).


colleenmcginnis (approved):
✨

Issue #2089, opened 2025-07-10
[Website]: Helm config documentation is incorrect


petdr:
### Type of issue

None

### What documentation page is affected

https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/configure-eck

### What happened?

On this page there is a piece of documentation

config:
  ca-cert-validity: 43800h

It should be instead

config:
  caValidity: 43800h

To find the valid values for config section mentioning the following command would be useful.

helm show values elastic/eck-operator

kilfoyle:
Thanks very much for reporting this @petdr! We'll try to get this fixed quickly.

kilfoyle:
This issue should be fixed by https://github.com/elastic/docs-content/pull/2098 The docs will be updated with the next build.

Thanks again for letting us know, @petdr!

PR #2098, opened 2025-07-10
Small fix in ECK Helm config docs

kilfoyle:
According to issue https://github.com/elastic/docs-content/issues/2089, the Helm `ca-cert-validity` setting documented on the [Apply ECK configuration settings](http://localhost:3000/deploy-manage/deploy/cloud-on-k8s/configure-eck) page should rather be `caValidity`. This also adds a tip for generating the list of valid settings.

I'm not familiar with these settings at all so I'd appreciate if anyone can confirm that this change is correct.

---

eck-cacert


eedugon:
This is interesting @kilfoyle !

This is actually a bug caused by me, let me explain.

I thought the helm chart passes directly everything inside `config` as configuration parameters to ECK, and in ECK, the theoretical configuration parameter to configure that is `ca-cert-validity`.

image

That's documented [here](https://www.elastic.co/docs/reference/cloud-on-k8s/eck-configuration-flags).

The problem is that the helm chart is using its own configuration parameters instead of directly relying on the actual ECK parameters ([this](https://github.com/elastic/cloud-on-k8s/blob/fee497796e8b2cdae994a99965fb2c9c2e881d1d/deploy/eck-operator/templates/configmap.yaml#L29) shows how `caValidity` is added to the actual `ca-cert-validity` setting).

To solve this we need to change a bit the narrative, because I wanted to tell users (on purpose) that they can configure ANY ECK setting documented in the ECK reference doc using the helm chart under the `config` section of the values...

And that's definitely incorrect now that I see the chart, sorry!

Let me share some small suggestions to address this change in a different way. For example, what you have added as a TIP should moved more to the beginning of the section, and probably we should add a note saying that the Helm chart does NOT use the standard configuration parameters documented in (LINK TO REFERENCE), and suggest to use the `helm show values` command to see the list of configurable settings (which should be aligned with the official config settings but with different syntax).

cc: @elastic/cloud-k8s , just in case you want to add anything here.

kilfoyle:
Thanks a lot @eedugon for explaining! I've updated the section so that we can explain at the top of the Helm section about the Helm settings being different from the ECK settings. I assume that Option 2, using the `--set` command, should similarly use the Helm version rather than the ECK version of the parameter. I've updated the PR changes as shown below.

---

eck-cacert2

deploy-manage/deploy/cloud-on-k8s/configure-eck.md
```diff
@@ -49,7 +55,7 @@ helm upgrade elastic-operator elastic/eck-operator -f my-values-file.yaml -n ela
 ### Option 2: Use `--set` in the helm upgrade command
 
 ```sh
-helm upgrade elastic-operator elastic/eck-operator --set config.ca-cert-validity=43800h -n elastic-system
+helm upgrade elastic-operator elastic/eck-operator --set config.caValidity=43800h -n elastic-system
```
eedugon, line 58: good catch :)

PR #2123, opened 2025-07-14
[Obs AI Assistant] Update summarize function and remove get_apm_timeseries

### LINKED ISSUE #2122: [Obs AI Assistant] Remove `get_apm_timeseries` and update `summarize` function description
- the get_apm_timeseries tool needs to be removed since it is deprecated (I had added it to the ticket later so maybe it was missed)

- the summarize tool is incorrectly described. This is the description we have in the codebase:
 ```
description: `Use this function to store facts in the knowledge database if the user requests it.
An embedding will be created that you can recall later with a semantic search.
When you create this summarisation, make sure you craft it in a way that can be recalled with a semantic
search later, and that it would have answered the user's original request., descriptionForUser:' 
This function allows the Elastic Assistant to summarize things from the conversation.',
```
also, for summarize we should mention that this tool is only available if the knowledge base is already installed

mdbirnstiehl:
This PR closes #2122 and updates the `summarize` function description and removes the `get_apm_timeseries` function.

### DIFF
--- a/solutions/observability/observability-ai-assistant.md
+++ b/solutions/observability/observability-ai-assistant.md
@@ -306,7 +306,7 @@ Main functions:
 :   Get relevant Elastic documentation. This function is only available if the product documentation is installed.
 
 `summarize`
-:   Store information and facts in the knowledge base for future use. This function is only available if the knowledge base has already been installed.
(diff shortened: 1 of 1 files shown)


eleonoramicozzi:
Thanks Mike, LGTM!

PR #2129, opened 2025-07-15
add ccs support for 8.19 and 9.1

ketkee-aryamane:
Linked to https://github.com/elastic/developer-docs-team/issues/313
Added versions 8.19 and 9.1 to the table listing support for cross cluster search. The following screenshot gives a preview of the change.

Added the 'applies_to' tag to signify GA for 9.1.

FireShot Capture 007 - Cross-cluster search - Elastic Docs -  localhost


leemthompo (commented):
Thanks for jumping on this @ketkee-aryamane! First off, the table is awful to work with 🤣 and there's additional complexity because `8.18` wasn't `8.last`.

Looking at the new table, it does look like we've introduced an error here: it now states that a local 9.0 is not compatible with a remote 8.18. This contradicts the explicit example text "_For example, a local 9.0 cluster can search a remote 8.18 or any remote 9.x cluster._ "

solutions/search/cross-cluster-search.md
```diff
@@ -1116,28 +1118,30 @@ $$$ccs-version-compatibility$$$
...
...com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![Yes](https://doc-icons.s3.us-east-2.amazonaws.com/icon-yes.png "") | ![Yes](https://doc-icons.s3.us-east-2.amazonaws.com/icon-yes.png "") | ![Yes](https://doc-icons.s3.us-east-2.amazonaws.com/icon-yes.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") | ![No](https://doc-icons.s3.us-east-2.amazonaws.com/icon-no.png "") |
```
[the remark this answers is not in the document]
ketkee-aryamane: @eedugon , but 9.0 becomes a major version right? How will 8.18 support it? I am unsure of 8.19 as well because of the same reason.

solutions/search/cross-cluster-search.md
```diff
@@ -1102,7 +1102,9 @@ Here’s how {{ccs}} works when you don’t minimize network roundtrips.
 
 
 ## Supported {{ccs}} configurations [ccs-supported-configurations]
-
+```{applies_to}
+stack: ga 9.1
+```
```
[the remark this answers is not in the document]
ketkee-aryamane: ok, got it.

solutions/search/cross-cluster-search.md
```diff
@@ -1100,52 +1100,55 @@ Here’s how {{ccs}} works when you don’t minimize network roundtrips.
...
+
+:::{note}
+As with version 8.19, cross-cluster searches are supported between a local 8.18 cluster and a remote 9.0 cluster, in both directions. This is possible because version 9.0 was released before 8.19. For more information on the order of releases, see [Elastic Search Releases.](https://github.com/elastic/elasticsearch/releases)
+:::
```
eedugon, line 1144: I like the note! It's a good way to explain the special case of the table that doesn't really follow the default rules described at the beginning of the section.

Have you checked how it looks adding also something like `(*)` in the items that are part of this exception?

What I don't like much on the proposed text is the link we are using, but not sure what to use.

I was thinking of [this link](https://www.elastic.co/guide/en/elasticsearch/reference/8.18/setup-upgrade.html#upgrade-newer-releases) because it talks about a similar topic, but that content needs to be added to the new docs (we lost it during the docs migration, and the link I'm showing you is for the legacy asciidoc docs).

@shainaraskas , could you share your thoughts here?

This note clarifies why certain items of the big table do not exactly follow the general rule.

eedugon (commented):
There's still a pending change that I've added in a comment.

About the note I think we should update the link and not linking to the github page of Elasticsearch from our official docs. If feels really weird. But I'm not sure where to link at the moment.

solutions/search/cross-cluster-search.md
ketkee-aryamane: I tried the asterisk mark, but it looks a little out of place Edu :). I'll have to search elsewhere in the repo of how I can include that next to the check marks and make it look seamless with the UI.
Also, wanted to understand why it would not be preferable to include our Git links.

eedugon: > On second thoughts @eedugon , do we really need the note? We're anyway offering more compatibility, not less. > So just check all correct boxes and get done? :))

In my opinion, and this is only my opinion, if a reader follows the statements about compatibility and then analyze a bit the table, they could believe the table is buggy and cause confusion like `is that really true?`, `which part of the doc is wrong?`, `are the statements about compatibility incorrect or is the table incorrect?`).

Because of that I think clarifying such an exception against the statements is good. Or simply make a note saying that the table might have extra compatible items due to special releases that were released "out of order" (with a link to the out of order info doc I have to work with in another PR :) ).

But.... do we really need the note? Maybe not :-D But I still belive we do, or at least something saying that the `rules` have certain exceptions due to `out order releases`.

Another option would be keeping the table totally standard, following the rules, and not showing this `extra compatibility`. We could check that with developers as soon as we are done with this PR preparations.

solutions/search/cross-cluster-search.md
```diff
@@ -1109,6 +1109,11 @@ In 8.0+, Elastic supports searches from a local cluster to a remote cluster runn
 
 Elastic also supports searches from a local cluster running the last minor version of a major version to a remote cluster running any minor version in the following major version. For example, a local 8.19 cluster can search any remote 9.x cluster.
 
+::::{note}
+[Elastic Search versions](https://www.elastic.co/downloads/past-releases#elasticsearch) were not released in a sequential order. Hence, additional compatibility conditions may apply.
+As shown in the following table, cross-cluster searches are supported between a local 8.18 cluster and a remote 9.0 cluster, in both directions. This is same as version 8.19 and is possible because version 9.0 was released before 8.19.
```
leemthompo, line 1114:
```suggestion
Version 8.19 is the last minor version of the 8.x series. Previously, the last minor version of a major series was released simultaneously with the first version of the next major series. However, 8.18 was released simultaneously with 9.0, creating bidirectional compatibility between these specific versions. As shown in the compatibility table, 8.18 can search 9.0 clusters, but unlike 8.19, it cannot search versions 9.1+.
```

I suggested this edit because the wording wasn't super precise for me.
I think the reference to "non-sequential order" is confusing if we don't explain the specifics here. 

Also, although, I ❤️ that linked page I'm not sure if it fits here. As mentioned earlier, the table itself should be accurate—so you don't need to reference that past releases page to determine compatibility.

eedugon: I'm ok with @leemthompo's suggestion, although the `previously part` might not add much benefit. I'd keep the rest.

My idea was just to say something `due to how certain versions are released out of order, the table shows extra compatibility cases not covered in the previous statements`, but if you prefer to show the specific example that's perfect too.

solutions/search/cross-cluster-search.md
```diff
@@ -1100,52 +1100,56 @@ Here’s how {{ccs}} works when you don’t minimize network roundtrips.
     ![ccs min roundtrip client response](/solutions/images/elasticsearch-reference-ccs-min-roundtrip-client-response.svg "")
 
 
-
 ## Supported {{ccs}} configurations [ccs-supported-configurations]
-
 In 8.0+, Elastic supports searches from a local cluster to a remote cluster running:
```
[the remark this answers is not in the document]
eedugon: ++ to that change. We don't need to specify 8.0 anymore in this context, as 7.x and earlier are irrelevant for the current doc set.

solutions/search/cross-cluster-search.md
ketkee-aryamane: I think it is too much info for a note & sounds confusing for me to read. Want to keep it concise. I can lose the vague statement and include the 8.18 thing as an example for the additional compatibility to make it clearer.
To avoid that explanation, I provided the link so that the reader could refer to release dates, if needed.

leemthompo: > I think it is too much info for a note & sounds confusing for me to read. Want to keep it concise.

We are not in a rush here, and if a note isn't the right format for delivering this information, this could be a heading about the specificities of 8.18 and 8.19 to help the user understand the context. For me explanation **>** link to a list of release dates (which don't explain CCS compatibility).

Take your time with this, we have until 9.1 releases to make this page as useful as possible :)

solutions/search/cross-cluster-search.md
eedugon: I'm going to work before 9.1 in a PR that explains the `out of order` releases (like [this](https://www.elastic.co/guide/en/elasticsearch/reference/8.18/setup-upgrade.html#upgrade-newer-releases)), and my plan is to include a statement that would indirectly explain this example (as I'll mention specifically 8.18 and 9.0).

After that is added, we could simplify this a bit and link to that doc for extra explanations.

What do you think @leemthompo and @ketkee-aryamane ? We can wait a bit to see how that looks.

The main objective (IMO) is to ensure the reader understands that certain `compatible cells` don't match the initial statements because exceptional out of order releases require that compatibility.

If a reader doesn't fully understand the `why` but at least the doc clearly says that the `unexpected` compatible items have a reason at least that shouldn't cause confusion.

leemthompo: Sure let's see, I assigned this ticket to work on a pretty small, but non-trivial piece of explanation, that is a pretty good microcosm of the complexities of working on Elastic docs. Ketkee needn't be in a rush to finish this, but should lean into the learnings and make sure the writing is clear as a first non-negotiable.

> If a reader doesn't fully understand the why but at least the doc clearly says that the unexpected compatible items have a reason at least that shouldn't cause confusion.

Yes the clear statement is what we should aim for, and if it's sufficiently clear it will clarify why 8.18 was exceptional and ensure the description and the table don't contradict each other.

solutions/search/cross-cluster-search.md
ketkee-aryamane: Note or not, I'm not sure if an explanation is needed and I'd prefer the link. I'll rephrase this a little bit and then prefer the devs to comment.

leemthompo: Fair enough, you know better than me ;)

leemthompo: sorry for the noise, I think there's a few too many cooks in the kitchen here, and there was a mix of copyediting + disagreements about how to present information. But as I said when I opened the issue we shouldn't have to guess too much here, we should ask the CCS experts how they want to communicate the specificities of 8.18/8.10. :)

solutions/search/cross-cluster-search.md
eedugon: @leemthompo , @ketkee-aryamane : 
FYI, and just in case it's useful to you, I'm adding the out-of-order release statement in this WIP PR: https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/2296/deploy-manage/upgrade/deployment-or-cluster#out-of-order-releases

My suggestion would be something simple, and if you want you can link to that page (still not merged).

As an example of something that is also a consequence of the same, which I'm writing in the same PR, I'm also adding something like this in another document (still working on the final wording)

> Because 8.18.0 and 9.0.0 were released simultaneously, upgrading from 8.18.x to 9.0.x is also supported. However, upgrading to 9.1.0 or later requires starting from 8.19.x.

I think that's aligned with the current text I see in the PR preview:

> Version 8.19 is the final minor release in the 8.x series. Unlike past releases, 8.18 was launched simultaneously with 9.0, allowing cross-version compatibility between them. Hence, as shown in the compatibility table, 8.18 can search 9.0 clusters, but only 8.19 supports searching 9.1 and later.

^^ I'm ok with the previous, but whatever you want! I also feel we shouldn't lose much time with this :)

solutions/search/cross-cluster-search.md
```diff
@@ -1100,52 +1100,55 @@ Here’s how {{ccs}} works when you don’t minimize network roundtrips.
...
-
+| Local cluster version | 7.17 | 8.0 | 8.1 | 8.2 | 8.3 | 8.4 | 8.5 | 8.6 | 8.7 | 8.8 | 8.9 | 8.10 | 8.11 | 8.12 | 8.13 | 8.14 | 8.15 | 8.16 | 8.17 | 8.18 | 8.19 | 9.0 | 9.1 |
```
naj-h, line 1121: @ketkee-aryamane I think the table is missing some references. E.g. a local 8.19 is listed as incompatible with a remote on 9.1, 8.x would be compatible with a remote on 8.19.

naj-h (commented):
I left a comment on the missing references

solutions/search/cross-cluster-search.md
naj-h: We should remove local 9.0 with remote 8.19
There are some paths that appear are unsupported in the table, for which we're going to double-check, but this might be something we change in the doc in the future, even after the release is out, so i'm good with merging this table

leemthompo (approved):

PR #2130, opened 2025-07-15
AIR Gapped - details about EPR health

lucabelluccini:


lucabelluccini:
This PR should be backported

colleenmcginnis:
@lucabelluccini are there any details you can link to to provide context about this change?

benironside (approved):

vishaangelova (approved):
@lucabelluccini, apologies! This PR had fallen through the cracks.

The changes look good to me. Based on [the implementation](https://github.com/elastic/package-registry/blob/main/health.go), it seems both the `/health` and `/health?ready=true` endpoints can be used to check if EPR is ready to serve requests, so I’ve added a suggestion to include that `/health` can be used for the same purpose (unless we want to distinguish between the two API endpoints).

PR #2131, opened 2025-07-15
Update s3-repository.md

eedugon:
I'd like to add an internal anchor to the admonition called `using MinIO with Elasticserach` in order to link to that specific statement from other places.

What do you think @shainaraskas ?


shainaraskas (approved):
lgtm. this type of thing is definitely up to your judgement 👍

PR #2146, opened 2025-07-16
ece medium and large installation instructions improved

eedugon:
This PR fixes and improves ECE medium and large installation instructions.

Fixes:
- The description of amount of allocators needed for a large installation wasn't coherent.
- We had a [bug](https://github.com/elastic/cloud/issues/133321) in the procedure where we were not adding the step of removing the allocator role from the initial host, which is important.
- Large instructions also needed to remove proxy role.
- Removing proxy role on first host has the implication of needing to update the endpoint URLs.

Enhancements:
- Added a small note to understand why the drawing shows the allocator role `greyed out` in one of the hosts. It improves the reading flow.
- Log into Cloud UI was in the wrong order :)
- Change deployment configuration link points now to the right place.

Closes https://github.com/elastic/cloud/issues/133321


deploy-manage/deploy/cloud-enterprise/deploy-large-installation.md
```diff
@@ -27,6 +27,10 @@ This type of installation is recommended for deployments with significant overal
 :alt: A large installation with nine to twelve hosts across three availability zones
 :::
 
+::::{note}
+The host with the allocator role greyed out represents the first host to be installed. This host temporarily holds all roles until the other nodes are added and configured. Eventually, the allocator role will be removed from this host.
```
kilfoyle, line 31:
```suggestion
In the diagram, the Director Coordinator host in Availability zone 1 has the allocator role greyed out. This host temporarily holds all roles until the other nodes are added and configured. Eventually, the allocator role will be removed from this host.
```

I'd rephrase this a bit for accessibility, since people reading using a screen reader can't access the diagram, and others can see the colour variances.

kilfoyle (approved):
LGTM! 🐅 
Looks great! Just one suggestion.

deploy-manage/deploy/cloud-enterprise/deploy-large-installation.md
eedugon: Ok, I'll rephrase it then, because I want to keep that the host represents the first host to be installed, as that information is not provided and it's relevant.

eedugon:
Thanks for your comments @kilfoyle .

I've realized that we had more bugs in the "large" doc, and I'll need to fix them before merging.

In the "large installation", we forgot to remove the proxy role (which is also greyed-out in first host).
Also by default ECE is configured with the IP address of the first installed node as the base to create the external `endpoint URLs`.
If that node removes the proxy role the URLs provided by cloud UI won't work, and the user needs to change the configuration of the Endpoint URL configuration in ECE -> Settings page. I'll also add a step and section to that.

kilfoyle (approved):
LGTM! ⛵ 
Looks superb @eedugon!

deploy-manage/deploy/cloud-enterprise/deploy-large-installation.md
```diff
@@ -100,8 +104,24 @@ Make sure you have completed all prerequisites and environment preparations desc
...
+
+8. [Vacate all instances from the initial host](/deploy-manage/maintenance/ece/move-nodes-instances-from-allocators.md#move-nodes-from-allocators). This host runs some {{es}} and {{kib}} instances from system deployments, which must be moved to other allocators before proceeding.
+
+    Wait until all instances have been moved off the initial host before continuing.
+
+9. [Remove the `allocator` and `proxy` roles](/deploy-manage/deploy/cloud-enterprise/assign-roles-to-hosts.md) from the initial host. You cannot remove the `allocator` role until all instances have been vacated.
+
+    ::::{note}
+    After removing the proxy role from the first host, the {{es}} and {{kib}} URLs shown in the Cloud UI will stop working. This happens because the **Deployment domain name** in **Platform** > **Settings** is set to the IP address of the first host, in a format like `FIRST_HOST_IP.ip.es.io`. For more details, refer to [Change endpoint URLs](./change-endpoint-urls.md).
```
eedugon, line 118:
```suggestion
    After removing the proxy role from the first host, the {{es}} and {{kib}} URLs shown in the Cloud UI will stop working. This happens because the **Deployment domain name** in **Platform** > **Settings** is set to the IP address of the first host, in the format `FIRST_HOST_IP.ip.es.io`. For more details, refer to [Change endpoint URLs](./change-endpoint-urls.md).
```

PR #2181, opened 2025-07-19
Add docs for async stop command

smalyshev:


explore-analyze/query-filter/languages/esql-rest.md
```diff
@@ -299,6 +299,24 @@ If the response’s `is_running` value is `false`, the query has finished and th
...
+
+```console-result
+{
+  "is_running": false,
+  "took": 48,
+  "is_partial": true,
+  "columns": ...
+}
+```
+This API will also return the result of the query if it has finished before the stop request was made, within the `keep_alive` period.
```
naj-h, line 317: Nit about the phrasing, I suggest "This API can be used to retrieve results even if the query has already completed, as long as it's within the `keep_alive` window."

naj-h (changes requested):
Small nit to clarify the phrasing. Otherwise LGTM

explore-analyze/query-filter/languages/esql-rest.md
```diff
@@ -299,6 +299,24 @@ If the response’s `is_running` value is `false`, the query has finished and th
 }
 ```
 
+You can interrupt a running async query and collect the results that have been computed so far by using the [async stop API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-esql-async-query-stop) with the query ID.
```
leemthompo, line 302: This is GA in 9.0 right, so we don't need additional applies_to tags?

leemthompo, line 302:
```suggestion
To stop a running async query and return the results computed so far, use the [async stop API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-esql-async-query-stop) with the query ID.
```

leemthompo: interrupt implies it can be resumed?

explore-analyze/query-filter/languages/esql-rest.md
```diff
@@ -299,6 +299,24 @@ If the response’s `is_running` value is `false`, the query has finished and th
...
+
+```console
+POST /_query/async/FmNJRUZ1YWZCU3dHY1BIOUhaenVSRkEaaXFlZ3h4c1RTWFNocDdnY2FSaERnUTozNDE=/stop
+```
+The query will be stopped and the response will contain the results computed so far, in the same format as the `get` API.
```
leemthompo, line 307:
```suggestion
The query will be stopped and the response will contain the results computed so far. The response format is the same as the `get` API.
```

leemthompo: sentence is a bit long

explore-analyze/query-filter/languages/esql-rest.md
```diff
@@ -299,6 +299,24 @@ If the response’s `is_running` value is `false`, the query has finished and th
...
+```console-result
+{
+  "is_running": false,
+  "took": 48,
+  "is_partial": true,
+  "columns": ...
+}
+```
+This API will also return the result of the query if it has finished before the stop request was made, within the `keep_alive` period.
+You can use the `is_partial` field to determine if the results are complete or not. If `is_partial` is `true`, the results may be incomplete.
```
leemthompo, line 318:
```suggestion
The `is_partial` field indicates result completeness. A value of `true` means the results are potentially incomplete.
```

Again just cutting sentence length. 

I'd also like to be more precise here if possible, like what's the difference between partial and incomplete? The more explicit the better where possible.

leemthompo (commented):
Few suggestions and I agree with @naj-h's rephrasing 👍

explore-analyze/query-filter/languages/esql-rest.md
smalyshev: Yes, that's correct.

leemthompo (approved):
LGTM

naj-h (approved):
LGTM

PR #2183, opened 2025-07-21
Update restore-snapshot.md

jaebongim:
HI, Team.
In order to restore the kibana session index, the Kibana instance must be terminated before deletion and restoration are possible. So I added the relevant part.
I would appreciate it if you could reflect it in the document after checking it.


eedugon:
@jaebongim : Do you have confirmation from devs or your own tests about stopping Kibana being needed in order to restore kibana data or the entire cluster?

I'd say in the past it was probably required, but not sure how it works today with the feature states.

eedugon (commented):
Updating the review to `comment` so we can review my previous question.

PR #2217, opened 2025-07-22
Small formatting fix for ILM docs

kilfoyle:
This fixes a small formatting glitch introduced while I was resolving feedback on the "Configure a policy" page.

Old (ugh - so bad!):

screen

---

New (much better!):

screen2


shainaraskas (approved):
the paragraph is finally safe in its home

PR #2224, opened 2025-07-23
Updates 'Trained model autoscaling' page for Serverless adaptive resources behavior

kosabogi:
On Serverless, adaptive resources are now always enabled.
This PR updates the Trained model autoscaling page to reflect this behavior.
Related issue: https://github.com/elastic/developer-docs-team/issues/309
This is a follow-up PR for https://github.com/elastic/docs-content/pull/2184.


deploy-manage/autoscaling/trained-model-autoscaling.md
```diff
@@ -73,16 +71,16 @@ You can choose from three levels of resource usage for your trained model deploy
 
 Refer to the tables in the [Model deployment resource matrix](#model-deployment-resource-matrix) section to find out the settings for the level you selected.
 
-:::{image} /deploy-manage/images/machine-learning-ml-nlp-deployment-id-elser-v2.png
+The image below shows the process of starting a trained model on an {{ech}} deployment. In {{serverless-short}} projects, the **Adaptive resources** toggle is not available when starting trained model deployments, as adaptive allocations are always enabled.
```
yetanothertw, line 74:
```suggestion
The image below shows the process of starting a trained model on an {{ech}} deployment. In {{serverless-short}} projects, the **Adaptive resources** toggle is not available when starting trained model deployments, as adaptive allocations are always enabled and cannot be disabled.
```

I know it repeats the sentence you added above in ln.25, but might be worth repeating in case it was missed? 
Feel free to ignore, as you know this area better than I do. :)

yetanothertw (approved):
Looks good, I've only added a minor comment, which is totally optional. 😸

deploy-manage/autoscaling/trained-model-autoscaling.md
kosabogi: I think it’s a great suggestion, as it makes the communication about this change clearer and more consistent. Thanks! :)

PR #2241, opened 2025-07-23
Mark semantic reranking as GA in 9.0

shainaraskas:
When something explicitly becomes GA in 9.0 we want to mark it as such ... anything below that we don't mark

part of the 9.1 audit


leemthompo (approved):
🙏

Issue #1235, opened 2025-04-22
[REQUEST]: Edit highlighted fields in alert flyout


christineweng:
### Description

In 8.19 and 9.1, we are introducing the ability to edit highlighted fields under `Investigations` -> `Highlighted Fields`. Users can add/remove highlighted fields for custom rule and prebuilt rules (with enterprise license). 

This feature is currently behind a feature flag, but we plan to enable it before 8.19/9.1 release.

![Image](https://github.com/user-attachments/assets/e111ebb4-6e54-4d36-895c-739259709f56)

![Image](https://github.com/user-attachments/assets/0a33985d-3b1b-48f8-9328-3ce3bb5e0c4a)

### Resources

PR: https://github.com/elastic/kibana/pull/216740

Issue: https://github.com/elastic/security-team/issues/9377

### Which documentation set does this change impact?

Elastic On-Prem and Cloud (all)

### Feature differences

No difference

### What release is this request related to?

N/A

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @christineweng 

**Stakeholders:** @paulewing

nastasha-solomon:
@natasha-moore-elastic I updated the Serverless docs via https://github.com/elastic/docs-content/pull/1540. For 9.1, we'll need to refresh the following images and move the tip into the description for the **Highlighted fields** section.

Media to update for 9.1:
- [ ] `security-open-alert-details-flyout.gif`
- [ ] `security-alert-details-flyout-preview-panel.gif`
- [ ] `security-alert-details-flyout-right-panel.png`
- [ ] `security-investigation-section-rp.png`


Media to update for 8.19: 
- [ ] `open-alert-details-flyout.gif`
- [ ] `alert-details-flyout-preview-panel.gif`
- [ ] `alert-details-flyout-right-panel.png`
- [ ] `investigation-section-rp.png`

nastasha-solomon:
Removing myself as an assignee since https://github.com/elastic/security-docs/pull/6924 is merged.

PR #2279, opened 2025-07-25
[Security]: Edit highlighted fields in alert flyout

natasha-moore-elastic:
Resolves https://github.com/elastic/docs-content/issues/1235.
Adds the appropriate 'applies' label to the tip about adding/removing custom highlighted fields.

This PR also removes all images on the page since they don't add significant value to the docs and cost a lot to maintain.


florent-leborgne (approved):
LGTM

PR #2307, opened 2025-07-29
ESQL: Update text formats documentation

ivancea:
Separated the ESQL endpoint formats docs in 3 blocks: Structured, Tabular and Binary (Open to suggestions for the names here).

Initially, it was only "Human readable" and "Binary". But CSV, TSV and TXT are different from JSON and YAML, as they don't show all the results, only the query results as a table (For example, no "took" or "is_partial" fields).

API specification change: https://github.com/elastic/elasticsearch-specification/pull/5054


explore-analyze/query-filter/languages/esql-rest.md
```diff
@@ -64,15 +64,21 @@ POST /_query?format=txt
 The URL parameter takes precedence over the HTTP headers. If neither is specified then the response is returned in the same format as the request.
 ::::
 
+::::{tip}
+In the "Tabular" formats, only the resulting data will be returned. Other response fields will be missing.
+Use them when you want a quick view of the results.
```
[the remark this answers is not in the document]
nik9000: It's fine with me. And lines up with my feelings about that format. Great for seeing the results in a dense way as a human being. CSV may even be good out of `curl` too. But "applications" should use json.

nik9000 (approved):

explore-analyze/query-filter/languages/esql-rest.md
```diff
@@ -64,15 +64,21 @@ POST /_query?format=txt
 The URL parameter takes precedence over the HTTP headers. If neither is specified then the response is returned in the same format as the request.
 ::::
 
+::::{tip}
+In the "Tabular" formats, only the resulting data will be returned. Other response fields will be missing.
```
bpintea, line 68:
```suggestion
In the tabular formats, only the result data is returned. The other response fields, such as those specific to async processing, are returned as HTTP headers.
```

bpintea (approved):
Added a suggested change, but LGTM, thanks for adding this.

explore-analyze/query-filter/languages/esql-rest.md
leemthompo: @ivancea alternatively you could have 3 little tables and introduce them with as much detail as you need in prose, just some ideas, don't have super strong opinions here :)

explore-analyze/query-filter/languages/esql-rest.md
```diff
@@ -64,22 +64,41 @@ POST /_query?format=txt
...
+
+Complete responses with metadata. Useful for automatic parsing.
 
 | `format` | HTTP header | Description |
 | --- | --- | --- |
-| Human readable |
-| `csv` | `text/csv` | [Comma-separated values](https://en.wikipedia.org/wiki/Comma-separated_values) |
+| Structured | | _Complete response with metadata_ |
 | `json` | `application/json` | [JSON](https://www.json.org/) (JavaScript Object Notation) human-readable format |
+| `yaml` | `application/yaml` | [YAML](https://en.wikipedia.org/wiki/YAML) (YAML Ain’t Markup Language) human-readable format |
```
leemthompo, line 74:
```suggestion
| `yaml` | `application/yaml` | [YAML](https://en.wikipedia.org/wiki/YAML) human-readable format |
```

nit: not sure if it's useful to provide the recursive acronym 😄

leemthompo (approved):
Looks great, and super helpful. 

_Optional addition suggestion:_ 
I was thinking at the start of this section, we could just have an example of how to set the format in Console after we say `You can set the format by specifying the format parameter in the URL or by setting the Accept or Content-Type HTTP header.` 

`For example:`


```sql
POST /_query?format=txt
```

explore-analyze/query-filter/languages/esql-rest.md
ivancea: I also don't think it adds much, but we have it in every other format, so I'd say to keep it, even if just for normalization and _flavor_ 👀

leemthompo: nothing wrong with a bit of flavor 🫡

bmorelli25 (approved):
✨

PR #2315, opened 2025-07-29
Support knn filter on nested metadata

mayya-sharipova:
This is to update our documentation that we now can support knn filter on nested metadata.

Related to https://github.com/elastic/elasticsearch/pull/113949


mayya-sharipova:
@leemthompo Thanks for the updated.

I have this section that is not true anymore from 9.2
```
What if you wanted to filter by some top-level document metadata? You can do this by adding `filter` to your `knn` clause.

::::{note}
`filter` will always be over the top-level document metadata. This means you cannot filter based on `nested` field metadata.
```
How it should be modified to this:
```
What if you wanted to filter by some top-level or nested metadata? You can do this by adding `filter` to your `knn` clause.

::::{note}
To ensure correct results: each individual filter must be either over
the top-level metadata or `nested` metadata. However, a single knn search
supports multiple filters, where some filters can be over the top-level
metadata and some over nested.
::::
```

leemthompo:
@mayya-sharipova Can achieve this by basically saying you have two ways to filter and noting that the nested metadata option is only available in 9.2+?

Something like this:

```markdown
Want to filter by metadata? You can do this by adding `filter` to your `knn` clause.

To ensure correct results: each individual filter must be either over:

-  Top-level metadata 
- `nested` metadata {applies_to}`stack: ga 9.2`

:::{note}
A single `knn` search supports multiple filters, where some filters can be 
over the top-level metadata and some over nested.
:::

**Comment by mayya-sharipova** [comment]
Thanks @leemthompo, I've added  `applies_to` 9.2 in the last [commit](63a2c7b2e64c39e7524ebe62050c32a0815a1524).

File: solutions/search/vector/knn.md [hunk 1]
```diff
@@ -740,6 +748,96 @@ Now we have filtered based on the top level `"creation_time"` and only one docum
...
+            }
+        }
+    }
+}
+```
+
+Below we have two filters: one over nested metadata
+and another over the top level metadata. For scoring parents' documents,
+this query only considers vectors that have "paragraph.language" set to "EN" 
+and whose parents have creation time within the request range.
```
leemthompo, line 783:
```suggestion
The following query has two filters: one over nested metadata and
another over the top-level metadata. When scoring parent documents, this
query only considers vectors that have "paragraph.language" set to "EN"
and whose parent documents were created within the specified range.
```

solutions/search/vector/knn.md
```diff
@@ -740,6 +748,96 @@ Now we have filtered based on the top level `"creation_time"` and only one docum
...
+                        ]
+                    }
+                }
+            }
+        }
+    }
+}
+```
+
+Retrieving "inner_hits" when filtering on sibling nested fiels is not supported.
```
leemthompo, line 840:
```suggestion
:::{note}
Retrieving "inner_hits" when filtering on sibling nested fields is not supported.
:::
```

This sentence feels lonely all on its own 😄

leemthompo (approved):
Thanks for working with the `applies_to` markers @mayya-sharipova! 🙏 

LGTM, with a few minor suggestions. Not sure if you want a peer dev review on the technical details before merging.

PR #2434, opened 2025-08-06
ECE 4.0.1 updates related to default stack packs

eedugon:
Updates related with the fact that ECE 4.0.1 comes with 8.18.2 and 9.0.1, and requires them when being installed offline.


eedugon:
Thanks @kilfoyle!

We still need an approval from @gigerdo , to be sure that the system deployment versions are indeed 8.18.2 in ECE 4.0.1.

gigerdo (approved):
LGTM :+1: Thank you for fixing this 🚀

PR #2453, opened 2025-08-07
Update troubleshooting-unbalanced-cluster.md

eedugon:
Suggestion by Mirko Bez, cross linked from an internal docs issue.

It clarifies that the last operation of the troubleshooting doc requires operator privileges, but maybe we want to explicitly mention that it's not allowed in ECE, ECH or ECK deployments (not totally sure about it).


kilfoyle (approved):
LGTM! 🦖

PR #2492, opened 2025-08-11
Link from file-realm to file-based roles doc

pebrc:
This lead to confusion with a user recently as it was not clear how to create the roles referenced in the basic-auth secret. 

~LMK if you prefer this to be one of the tip boxes.~ Actually changed my mind and the tip box seems to make more sense here.

Screenshot 2025-08-11 at 12 09 34


kilfoyle (approved):
LGTM! 🚀 
Nice! Thanks @pebrc!

PR #2501, opened 2025-08-11
Clarifies behavior of using Direct IO for vector data

szabosteve:
## Overview

This PR adds `applies_to` tags to the default behavior of Direct IO in 9.1.0 and 9.1.1.


kilfoyle (approved):
LGTM! 🦕

PR #2505, opened 2025-08-11
[Private preview] AutoOps for self-managed clusters: Hide K8s helm chart option

wajihaparvez:
Helm chart won't be available as a command format for now.

Rel: https://github.com/elastic/docs-content-internal/issues/169


arnonst (approved):
LGTM

kilfoyle (approved):
LGTM! 🌵

PR #2532, opened 2025-08-12
Clarify availability of ILM and data stream lifecycle

kilfoyle:
The scope of ILM and of data stream lifecycle needs to be called out more clearly, with respect to the supported deployment types and to the objects that they can manage (data streams versus standalone indices).

This adds an "Availability" section to the top of the main page for each.

See:
 - 

Rel: #190

### DIFF
--- a/manage-data/lifecycle/data-stream.md
+++ b/manage-data/lifecycle/data-stream.md
@@ -10,7 +10,7 @@ products:
 
 # Data stream lifecycle [data-stream-lifecycle]
 
-A data stream lifecycle is the built-in mechanism data streams use to manage their lifecycle. It enables you to easily automate the management of your data streams according to your retention requirements. For example, you could configure the lifecycle to:
+A data stream lifecycle is the built-in mechanism [data streams](/manage-data/data-store/data-streams.md) use to manage their lifecycle. It enables you to easily automate the management of your data streams according to your retention requirements. For example, you could configure the lifecycle to:
 
 * Ensure that data indexed in the data stream will be kept at least for the retention time you defined.
 * Ensure that data older than the retention period will be deleted automatically by {{es}} at a later time.
@@ -22,6 +22,13 @@ To achieve that, it supports:
 
 A data stream lifecycle also supports downsampling the data stream backing indices. See [the downsampling example](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-put-data-lifecycle) for more details.
 
+## Availability
+
+* Data stream lifecycle is supported only for data streams and cannot be used with standalone indices.
+
+* Like {{ilm-cap}} ({{ilm-init}}), data stream lifecycle is supported in both {{ecloud}} and self-managed environments. Data stream lifecycle is less feature rich than {{ilm-init}} and is focused on simplicity. For a detailed comparison refer to [Data lifecycle](/manage-data/lifecycle.md).
+
+* Owing to its simplicity compared with {{ilm-cap}} ({{ilm-init}}), data stream lifecycle is the data lifecycle tool used with {{es-serverless}}. For an {{ecloud}} or self-managed environment, {{ilm-init}} helps you to balance hardware costs with performance for your data, but this complexity isn't required in a {{serverless-short}} environment in which your cluster performance is managed automatically.
 
 ## How does it work? [data-streams-lifecycle-how-it-works]
 
--- a/manage-data/lifecycle/index-lifecycle-management.md
+++ b/manage-data/lifecycle/index-lifecycle-management.md
@@ -13,23 +13,25 @@ products:
 
 # Index lifecycle management
 
-{{ilm-cap}} ({{ilm-init}}) provides an integrated and streamlined way to manage time-based data such as logs and metrics, making it easier to follow best practices for managing your indices.
+% {{ilm-cap}} ({{ilm-init}}) provides an integrated and streamlined way to manage time-based data such as logs and metrics, making it easier to follow best practices for managing your indices.
 
-You can configure {{ilm-init}} policies to automatically manage indices according to your performance, resiliency, and retention requirements. For example, you could use {{ilm-init}} to:
+{{ilm-cap}} ({{ilm-init}}) provides an integrated and streamlined way to manage your time series data. You can configure {{ilm-init}} policies to automatically manage indices according to your performance, resiliency, and retention requirements. For example, you could use {{ilm-init}} to:
 
 * Spin up a new index when an index reaches a certain size or number of documents
 * Create a new index each day, week, or month and archive previous ones
 * Delete stale indices to enforce data retention standards
 
-::::{tip}
-{{ilm-init}} is not available on {{es-serverless}}.
+## Availability
 
-:::{dropdown} Why?
-In an {{ecloud}} or self-managed environment, ILM lets you automatically transition indices through data tiers according to your performance needs and retention requirements. This allows you to balance hardware costs with performance. {{es-serverless}} eliminates this complexity by optimizing your cluster performance for you.
+* {{ilm-init}} is supported for both data streams and standalone indices:
 
-Data stream lifecycle is an optimized lifecycle tool that lets you focus on the most common lifecycle management needs, without unnecessary hardware-centric concepts like data tiers.
-:::
-::::
+    * **Data stream:** A [data stream](/manage-data/data-store/data-streams.md) acts as a layer of abstraction over a set of indices that contain append-only, time series data. You can configure {{ilm-init}} using a data stream as a single named resource, so that rollover and any other configured actions are performed on the data stream's backing indices automatically.
+
+    * **Standalone index:** When you use {{ilm-init}} with standalone indices, you need to manage the lifecycle of each individual index directly, typically by configuring an index alias. This allows for more granular control over each index but requires considerably more effort compared to using a data stream, which is our recommended option.
+
+* {{ilm-init}} is available for {{ecloud}} or self-managed environments only, as a tool to help you balance hardware costs with performance for your data. {{es-serverless}} removes the need for this complexity by optimizing your cluster performance automatically. In a {{serverless-short}} environment, data stream lifecycle (see the following bullet) is available as a data lifecycle option.
+
+* If you're looking for a simpler option than using {{ilm-init}}, [data stream lifecycle](/manage-data/lifecycle/data-stream.md) is available as a less feature rich lifecycle management tool optimized for the most common lifecycle management needs. It enables you to configure the retention duration for your data and to optimize how the data is stored, without hardware-centric concepts like data tiers.
 
 ::::{important}
 To use {{ilm-init}}, all nodes in a cluster must run the same version. Although it might be possible to create and apply policies in a mixed-version cluster, there is no guarantee they will work as intended. Attempting to use a policy that contains actions that aren’t supported on all nodes in a cluster will cause errors.
--- a/manage-data/lifecycle/index-lifecycle-management.md
+++ b/manage-data/lifecycle/index-lifecycle-management.md
@@ -31,7 +31,7 @@ products:
 
 * {{ilm-init}} is available for {{ecloud}} or self-managed environments only, as a tool to help you balance hardware costs with performance for your data. {{es-serverless}} removes the need for this complexity by optimizing your cluster performance automatically. In a {{serverless-short}} environment, data stream lifecycle (see the following bullet) is available as a data lifecycle option.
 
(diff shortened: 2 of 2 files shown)


manage-data/lifecycle/index-lifecycle-management.md
```diff
@@ -13,23 +13,27 @@ products:
...
+* {{ilm-init}} is supported for both data streams and standalone indices:
+
+    * **Data stream:** A [data stream](/manage-data/data-store/data-streams.md) acts as a layer of abstraction over a set of indices that contain append-only, time series data. You can configure {{ilm-init}} using a data stream as a single named resource, so that rollover and any other configured actions are performed on the data stream's backing indices automatically.
+
+    * **Standalone index:** When you use {{ilm-init}} with standalone indices, you need to manage the lifecycle of each individual index directly, typically by configuring an index alias. This allows for more granular control over each index but requires considerably more effort compared to using a data stream, which is our recommended option.
+
+* {{ilm-init}} is available for {{ecloud}} or self-managed environments only, as a tool to help you balance hardware costs with performance for your data. {{es-serverless}} removes the need for this complexity by optimizing your cluster performance automatically. In a {{serverless-short}} environment, data stream lifecycle (see the following bullet) is available as a data lifecycle option.
```
yetanothertw, line 34: This would need some apply_to tags to draw attention to the deployment types where ILM is available. If I'm being completely honest, I think the previous tip type of admonition works better in this instance. Visually, it breaks up the text nicely, and also it sort of mentions Serverless and data stream lifecycles tangentially (because this page is not about that, but it still relates to the same concept: lifecycle management), so the connection is isolated and easy to find for someone who accidentally opened this page but is looking for directions on how to get to the data stream lifecycle one.

manage-data/lifecycle/index-lifecycle-management.md
```diff
@@ -13,23 +13,27 @@ products:
...
+
+* If you're looking for a simpler option than using {{ilm-init}}, [data stream lifecycle](/manage-data/lifecycle/data-stream.md) is available as a less feature rich lifecycle management tool optimized for the most common lifecycle management needs. It enables you to configure the retention duration for your data and to optimize how the data is stored, without hardware-centric concepts like data tiers. For a comparison of {{ilm-init}} and data stream lifecycle refer to [Data lifecycle](/manage-data/lifecycle.md).
```
yetanothertw, line 36: Instead of saying this option is `less feature rich` , I'd put a positive spin on it to say it's `simpler, and instead of multiple phase transitions, it focuses on retention`. Also, we should probably mention that this option only works for data streams (it's optimised for continuous data ingestion scenarios).

manage-data/lifecycle/data-stream.md
```diff
@@ -22,6 +22,15 @@ To achieve that, it supports:
 
 A data stream lifecycle also supports downsampling the data stream backing indices. See [the downsampling example](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-put-data-lifecycle) for more details.
 
+## Availability
```
yetanothertw, line 25: I think the heading should be more specific, so maybe like `## ILM availability`

Also, the `## Actions` heading should maybe be something like `## Index lifecycle actions`?

manage-data/lifecycle/index-lifecycle-management.md
yetanothertw: I think we're missing a comparative statement about ILM here.

You use ILM if you need tiered storage and precise control over your indices (for example with actions like rollover, shrink, or force merge). In contrast, you use data stream lifecycle if you're looking to automate retention for time-series data streams with minimal configuration.

manage-data/lifecycle/data-stream.md
```diff
@@ -22,6 +22,15 @@ To achieve that, it supports:
...
+
+Note the availability of data stream lifecycle to ensure that it's applicable for your use case:
+
+* Data stream lifecycle is supported only for data streams and cannot be used with standalone indices.
+
+* Like [{{ilm-cap}} ({{ilm-init}})](/manage-data/lifecycle/index-lifecycle-management.md), data stream lifecycle is supported in both {{ecloud}} and self-managed environments. Data stream lifecycle is less feature rich than {{ilm-init}} and is focused on simplicity. For a detailed comparison refer to [Data lifecycle](/manage-data/lifecycle.md).
```
shainaraskas, line 31: I'd avoid "{{ecloud}} and self-managed environments" because we've found that this is not comprehensive enough. to external users, "elastic cloud" could be "elastic cloud on kubernetes", which is technically a self-managed option. I'd indicate either that this is for deployments of the versioned stack, or list out all of the supported deployment types.

manage-data/lifecycle/index-lifecycle-management.md
```diff
@@ -13,23 +13,27 @@ products:
...
 * Create a new index each day, week, or month and archive previous ones
 * Delete stale indices to enforce data retention standards
 
-::::{tip}
-{{ilm-init}} is not available on {{es-serverless}}.
+## Availability
 
-:::{dropdown} Why?
-In an {{ecloud}} or self-managed environment, ILM lets you automatically transition indices through data tiers according to your performance needs and retention requirements. This allows you to balance hardware costs with performance. {{es-serverless}} eliminates this complexity by optimizing your cluster performance for you.
+Note the availability of {{ilm-init}} to ensure that it's applicable for your use case:
```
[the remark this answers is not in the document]
kilfoyle: Yup. I've changed it to a period as suggested.

manage-data/lifecycle/index-lifecycle-management.md
```diff
@@ -13,23 +13,27 @@ products:
...
 
-Data stream lifecycle is an optimized lifecycle tool that lets you focus on the most common lifecycle management needs, without unnecessary hardware-centric concepts like data tiers.
-:::
-::::
+* {{ilm-init}} is supported for both data streams and standalone indices:
+
+    * **Data stream:** A [data stream](/manage-data/data-store/data-streams.md) acts as a layer of abstraction over a set of indices that contain append-only, time series data. You can configure {{ilm-init}} using a data stream as a single named resource, so that rollover and any other configured actions are performed on the data stream's backing indices automatically.
+
+    * **Standalone index:** When you use {{ilm-init}} with standalone indices, you need to manage the lifecycle of each individual index directly, typically by configuring an index alias. This allows for more granular control over each index but requires considerably more effort compared to using a data stream, which is our recommended option.
```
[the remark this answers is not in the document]
kilfoyle: Thanks. I've rephrased it something like this, together with the point about data streams being the preferred option (which I think was Edu's main concern).

manage-data/lifecycle/data-stream.md
kilfoyle: I've changed the titles as suggested. Similarly `Data stream lifecycle availability` for the other page.

manage-data/lifecycle/index-lifecycle-management.md
```diff
@@ -7,35 +7,42 @@ mapped_pages:
...
-::::
+    * **Indices:** You use {{ilm-init}} to manage a specific index or set of indices by defining a lifecycle policy and applying it to the indices or an index alias. Each index is then evaluated against its policy and transitions through phases (`hot`, `warm`, `cold`, `frozen`, `delete`) based on pre-defined conditions. This approach allows for more granular control over each index but requires considerably more effort compared to using a data stream, which is our recommended option.
+
+    * **Data streams:** A [data stream](/manage-data/data-store/data-streams.md) acts as a layer of abstraction over a set of indices that contain append-only, time series data. You can configure {{ilm-init}} using a data stream as a single named resource, so that rollover and any other configured actions are performed on the data stream's backing indices automatically.
+
+* {{ilm-init}} is available for all deployment types on the versioned {{stack}} but is not available for {{es-serverless}}. In a {{serverless-short}} environment, data stream lifecycle (see the following tip) is available as a data lifecycle option.
+
+    :::{tip}
```
shainaraskas, line 35:
```suggestion
    :::{admonition} Simpler lifecycle management in Serverless environments
```

manage-data/lifecycle/index-lifecycle-management.md
```diff
@@ -7,35 +7,42 @@ mapped_pages:
...
+    {{ilm-init}} lets you automatically transition indices through data tiers according to your performance needs and retention requirements. This allows you to balance hardware costs with performance. {{es-serverless}} eliminates this complexity by optimizing your cluster performance for you. In a {{serverless-short}} environment, data stream lifecycle is available as a data management option.
+
+    {applies_to}`stack: ga` {applies_to}`serverless: ga` [Data stream lifecycle](/manage-data/lifecycle/data-stream.md) is a simpler lifecycle management tool optimized for the most common lifecycle management needs. It enables you to configure the retention duration for your data and to optimize how the data is stored, without hardware-centric concepts like data tiers. For a detailed comparison of {{ilm-init}} and data stream lifecycle refer to [Data lifecycle](/manage-data/lifecycle.md).
```
shainaraskas, line 38: these applies tags don't feel right here. I'd skip them.

the benefit is really much greater for serverless environments and dsl is not powerful enough for all stack cases, so communicating that it's available in stack isn't super critical

can drop the link if it's in the bullet item above

```suggestion
    Data stream lifecycle is a simpler lifecycle management tool optimized for the most common lifecycle management needs. It enables you to configure the retention duration for your data and to optimize how the data is stored, without hardware-centric concepts like data tiers. For a detailed comparison of {{ilm-init}} and data stream lifecycle refer to [Data lifecycle](/manage-data/lifecycle.md).
```

manage-data/lifecycle/index-lifecycle-management.md
```diff
@@ -7,35 +7,42 @@ mapped_pages:
...
+
+    * **Data streams:** A [data stream](/manage-data/data-store/data-streams.md) acts as a layer of abstraction over a set of indices that contain append-only, time series data. You can configure {{ilm-init}} using a data stream as a single named resource, so that rollover and any other configured actions are performed on the data stream's backing indices automatically.
+
+* {{ilm-init}} is available for all deployment types on the versioned {{stack}} but is not available for {{es-serverless}}. In a {{serverless-short}} environment, data stream lifecycle (see the following tip) is available as a data lifecycle option.
+
+    :::{tip}
+    {{ilm-init}} lets you automatically transition indices through data tiers according to your performance needs and retention requirements. This allows you to balance hardware costs with performance. {{es-serverless}} eliminates this complexity by optimizing your cluster performance for you. In a {{serverless-short}} environment, data stream lifecycle is available as a data management option.
```
shainaraskas, line 36: I think we need to double down on "ILM is not available in serverless BECAUSE" - splitting the info from the note has made that a little harder to parse

manage-data/lifecycle/index-lifecycle-management.md
```diff
@@ -7,35 +7,42 @@ mapped_pages:
...
+
+    {applies_to}`stack: ga` {applies_to}`serverless: ga` [Data stream lifecycle](/manage-data/lifecycle/data-stream.md) is a simpler lifecycle management tool optimized for the most common lifecycle management needs. It enables you to configure the retention duration for your data and to optimize how the data is stored, without hardware-centric concepts like data tiers. For a detailed comparison of {{ilm-init}} and data stream lifecycle refer to [Data lifecycle](/manage-data/lifecycle.md).
+    :::
+
+    ::::{important}
+    To use {{ilm-init}}, all nodes in a cluster must run the same version. Although it might be possible to create and apply policies in a mixed-version cluster, there is no guarantee they will work as intended. Attempting to use a policy that contains actions that aren’t supported on all nodes in a cluster will cause errors.
```
shainaraskas, line 42: is this an availability concern or a prereq?

manage-data/lifecycle/index-lifecycle-management.md
```diff
@@ -7,35 +7,42 @@ mapped_pages:
...
 
 * Spin up a new index when an index reaches a certain size or number of documents
 * Create a new index each day, week, or month and archive previous ones
 * Delete stale indices to enforce data retention standards
 
-::::{tip}
-{{ilm-init}} is not available on {{es-serverless}}.
+    ::::{important}
+    To use {{ilm-init}}, all nodes in a cluster must run the same version. Although it might be possible to create and apply policies in a mixed-version cluster, there is no guarantee they will work as intended. Attempting to use a policy that contains actions that aren’t supported on all nodes in a cluster will cause errors.
+    ::::
```
shainaraskas, line 25: I don't think we need this to be indented because it's not a child of deleting stale indices

```suggestion
::::{important}
To use {{ilm-init}}, all nodes in a cluster must run the same version. Although it might be possible to create and apply policies in a mixed-version cluster, there is no guarantee they will work as intended. Attempting to use a policy that contains actions that aren’t supported on all nodes in a cluster will cause errors.
::::
```

shainaraskas (approved):
coolcoolcool

one tiny indent thing

PR #2568, opened 2025-08-13
Fix applies_to tag

mdbirnstiehl:
Noticed this while doing a little research. Also, my VS code deleted some extra spaces.

### DIFF
--- a/explore-analyze/visualize/esorql.md
+++ b/explore-analyze/visualize/esorql.md
@@ -10,7 +10,7 @@ products:
 
 # ES|QL [esql-visualizations]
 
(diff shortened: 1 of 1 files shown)


bmorelli25:
I'm torn. I do think this is right long term solution, but I don't think we've agreed on that as a team yet. I added this as a discussion point in https://github.com/elastic/docs-content/pull/2568.

florent-leborgne:
Yep and actually what we're currently [in the process of agreeing on](https://github.com/elastic/docs-builder/pull/1614), beyond this, is that we shouldn't repeat the serverless badge and consider it implied. I expect that this will be a separated task that we handle at once to clean up the docs.

PR #2581, opened 2025-08-14
Add missing applies_to tag on the Observability Cloud page

alaudazzi:
This PR adds the missing `serverless` tag to the https://www.elastic.co/docs/solutions/observability/cloud page.


mdbirnstiehl (approved):
I think we may want to add "ga" to these. I know that is missing from a lot of pages, and I need to update in other places as well.

PR #2583, opened 2025-08-14
[Fleet+Elastic Agent] Clarify that upgrade rollbacks are not ready for use yet in relevant changelog entries

cmacknz:
The wording of the rollback changelogs is great but they imply to some users that the feature is ready for use yet when it isn't. Add language indicating these changes are building towards the feature in a future release.


benironside (approved):
LGTM

PR #2602, opened 2025-08-15
Explain VCU consumption better

shubhaat:
Explain Ingest, Search, ML VCU consumption better. I see this confusion on - 

Are search VCUs directly related to number of searches?  Are ingest VCUs directly related to number of ingest operations? 

I have added info to break down what search/ingest/ML VCU consumption is driven by.


kilfoyle (approved):
LGTM! 🚀 
Thanks @shubhaat

PR #2613, opened 2025-08-18
Add groups workaround for BYOK GCP config

shainaraskas:
- Fixes https://github.com/elastic/docs-content-internal/issues/150
- Adds note with workaround to address the domain restricted sharing case for GCP

image


alxchalkias (approved):
LGTM!

kilfoyle (approved):
LGTM! 🦕 

It's outside the scope of this issue, but it would be nice for this page to use tab sets so that the aws/azure/gcp selection persists across all tab sections on the page.

PR #2614, opened 2025-08-18
terminology fix - invoices instead of receipts

shainaraskas:
"invoice" is present in the UI. confirmed with product ([slack link](https://elastic.slack.com/archives/CJZ9DPRHC/p1745955123079989))

do we need to inject "statements" anywhere?

fixes https://github.com/elastic/docs-feedback/issues/49


briannebartlett:
We currently have Statements and Invoices available in the Billing History as well as legacy Receipts. We migrated to Statements and Invoices for monthly customers ~1 year ago, so any monthly customer with an account older than 1 year could see Statements, Invoices and old legacy Receipts. 

Screenshot 2025-08-19 at 11 07 50 AM

shainaraskas:
@briannebartlett how would you like us to position this in the docs? "statements, invoices, and receipts" everywhere we said "receipts"?

PR #2618, opened 2025-08-19
[API docs contrib guidance] Add clarification about `@availability since` in ES spec

leemthompo:
Replicates https://github.com/elastic/elasticsearch-specification/pull/5149

h/t @pquentin


szabosteve (approved):
Thank you for this addition!

Issue #1570, opened 2025-05-30
Data lifecycle docs: Clarify ILM versus "data stream lifecycle" usage


kilfoyle:
We should check the tutorials to make sure the the above terms are used clearly and consistently, with some explanation of the differences. We could just make sure that the main Data lifecycle page is crystal clear, and then provide links in the tutorials where those terms are used back to that main page.

Discussion:
 - [Data lifecycle](https://www.elastic.co/docs/manage-data/lifecycle) (42:00): George: We talk about ILM and data stream lifecycle, but we 
don't really compare them. Zoya: I'd probably differentiate between Index and Data stream.
 - Tutorial [Update existing data stream](https://www.elastic.co/docs/manage-data/lifecycle/data-stream/tutorial-update-existing-data-stream) (and others in that set) (42:30): “Data stream lifecycle policy” and “index lifecycle policy” are exactly the same thing.

(See 42:00 in recording)

yetanothertw:
We can add applies_to tags for each heading on the [Data lifecycle](https://www.elastic.co/docs/manage-data/lifecycle) page.

yetanothertw:
Cool, we can chat about this during our weekly sync and make a plan. 👍

yetanothertw:
As agreed in the sync, marking this ticket as done.

PR #2631, opened 2025-08-20
Adding tags to clarify data management options

yetanothertw:
Relates to #1570

Adding `applies_to` tags to visually indicate that ILM and Curator are not available in Serverless, whereas Data streams is.


kilfoyle (approved):
LGTM! 🌵 
Nice catch on `visualation`!

PR #2632, opened 2025-08-20
Fix synthetics docs for read-only user !!

shahzad31:
This seems a bug which we should fixed, for synthetics read only user we only need read permissions !!

Flagged by a user https://github.com/elastic/enhancements/issues/25472


mdbirnstiehl (approved):
LGTM, thanks for the fix!

shahzad31:
@mdbirnstiehl how can this be backported?

shahzad31:
> > @mdbirnstiehl how can this be backported?
> 
> For 8.x docs, a separate PR needs to be opened in the observability-docs repo. I can handle that for you if you like.

Sure please, thank you

PR #2634, opened 2025-08-20
Note that upgrades to ECK 3.1 cause a restart

pebrc:
I would like to add a note that the restart only happens if the user is make use of the stack monitoring feature in ECK but I don't know the best way to express that in the docs syntax. 

Normally I would have wanted to add a footnote but I don't think we have those in the docs/markdown. Open for suggestions. If there is no good way that an unqualified mention is still better than the status quo.

### DIFF
--- a/deploy-manage/security/fips-ingest.md
+++ b/deploy-manage/security/fips-ingest.md
@@ -85,14 +85,23 @@ When you use {{agent}} and {{fleet-server}}, these limitations apply:
 * Running {{agent}} in [OpenTelemetry mode](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/README.md) is not yet supported. This includes all receivers, such as Filebeat Receiver, Metricbeat Receiver, [Prometheus Receiver](https://www.elastic.co/docs/reference/integrations/prometheus).
 * Some Elastic Integrations are not FIPS compatible, as they depend on functionality that is not yet supported for FIPS configuration. In general, when using {{agent}} and {{fleet-server}}, the same restrictions listed previously for {{metricbeat}} and {{filebeat}} modules, inputs, and processors apply.
 
-  These Elastic Integrations have components that are **not** FIPS compatible, and **cannot** be used in FIPS environments, even if combined with other ingest tools that offer FIPS mode. 
+### Elastic Integrations that are not FIPS compatible [ingest-limitations-integrations]
 
-  - [Azure Logs Integration (v2 preview)](integration-docs://reference/azure/events.md)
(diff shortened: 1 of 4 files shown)


kilfoyle:
@pebrc I think there are two options:

Using a [code callout](https://elastic.github.io/docs-builder/syntax/code/#code-callouts):

```
(three backticks)
1.6, 1.9, 2.0, 2.1, 2.2, 2.4, 2.5, 2.6, 2.7, 2.8, 2.14, 3.1 <1>
(three backticks)

1. The restart happens only if you're using the stack monitoring feature in ECK.
````


Which renders like this:

code

One problem is that the admonition must appear at the end of the line.

---

Alternatively, we could update the [admonition](https://elastic.github.io/docs-builder/syntax/admonitions/) that appears below the versions list:


```
(three backticks)
1.6, 1.9, 2.0, 2.1, 2.2, 2.4, 2.5, 2.6, 2.7, 2.8, 2.14, 3.1
(three backticks)
::::{admonition} Notes
Note the following about rolling restarts:
- The restart only happens if the user is making use of the stack monitoring feature in ECK. 
- Stepping over one of these versions, for example, upgrading ECK from 2.6 to 2.9, still triggers a rolling restart.
::::
````
which renders like this:

notes

File: deploy-manage/upgrade/orchestrator/upgrade-cloud-on-k8s.md [hunk 1]
```diff
@@ -101,8 +101,11 @@ This will update the ECK installation to the latest binary and update the CRDs a
 Upgrading the operator results in a one-time update to existing managed resources in the cluster. This potentially triggers a rolling restart of pods by Kubernetes to apply those changes. The following list contains the ECK operator versions that would cause a rolling restart after they have been installed.
 
 ```
-1.6, 1.9, 2.0, 2.1, 2.2, 2.4, 2.5, 2.6, 2.7, 2.8, 2.14
+1.6, 1.9, 2.0, 2.1, 2.2, 2.4, 2.5, 2.6, 2.7, 2.8, 2.14, 3.1 <1>
 ```
+
+<1> The restart when upgrading to version 3.1 happens only for applications using [stack monitoring](/deploy-manage/monitor/stack-monitoring/eck-stack-monitoring.md).
```
kilfoyle, line 107:
```suggestion
1. The restart when upgrading to version 3.1 happens only for applications using [stack monitoring](/deploy-manage/monitor/stack-monitoring/eck-stack-monitoring.md).
```
I think this will fix the rendering.

kilfoyle (approved):
LGTM! 🚀

PR #2690, opened 2025-08-25
[ResponseOps][9.x & Serverless] Clarify untracked alerts 

nastasha-solomon:
Contributes to https://github.com/elastic/platform-docs-team/issues/641 by adding that untracked alerts can no longer have their status updated.

**Corresponding 8.x updates** 
- Kibana: https://github.com/elastic/kibana/pull/232790
- Observability: https://github.com/elastic/observability-docs/pull/4952

[Previews](https://github.com/elastic/docs-content/pull/2690#issuecomment-3220291309)

### DIFF
--- a/manage-data/ingest.md
+++ b/manage-data/ingest.md
@@ -16,7 +16,7 @@ products:
   - id: elasticsearch
 ---
 
-# Bring your data to Elastic
+# Ingest: Bring your data to Elastic
 
 Whether you call it *adding*, *indexing*, or *ingesting* data, you have to get the data into {{es}} before you can search it, visualize it, and use it for insights.
 
--- a/reference/fleet/fleet-agent-release-process.md
+++ b/reference/fleet/fleet-agent-release-process.md
@@ -1,0 +1,20 @@
+---
+navigation_title: "{{agent}} release process"
+applies_to:
+  stack: ga
+products:
+  - id: fleet
+  - id: elastic-agent
+---
+
+# {{agent}} release process [fleet-agent-release-process]
+
+## Scheduled releases
+
+{{agent}} follows a release process aligned with the broader {{stack}} release schedule. The latest features, enhancements, and fixes are documented in the [release notes](../../release-notes/fleet-elastic-agent/index.md).
+
+## Independent {{agent}} releases [independent-agent-releases]
+
+Independent {{agent}} releases deliver critical fixes and updates for {{agent}} and {{elastic-defend}}, independently of the full stack release. This is a more conservative process than a typical patch release, and only modifies the specific {{agent}} components needed for a targeted fix. For example, an independent {{agent}} hotfix release for {{elastic-defend}} would only change the endpoint-security executable, with the remaining executables being exactly those released in the previous patch.
+
+In independent {{agent}} releases, a build identifier is appended to the semantic version of the base release in the format `+build{yyyymmddhhmm}`, where `{yyyymmddhhmm}` is the release timestamp of the build.
--- a/reference/fleet/toc.yml
+++ b/reference/fleet/toc.yml
@@ -19,6 +19,7 @@ toc:
       - file: fleet-server-monitoring.md
   - file: install-elastic-agents.md
     children:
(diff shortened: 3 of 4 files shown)


mdbirnstiehl (approved):
🚀

explore-analyze/alerts-cases/alerts/view-alerts.md
```diff
@@ -77,11 +77,11 @@ Alert flapping is turned on by default. You can modify the criteria for changing
...
 
 ## Mute alerts [mute-alerts]
 
-If an alert is active or flapping, you can mute it to temporarily suppress future actions. In both **{{stack-manage-app}} > Alerts** and **{{rules-ui}}**, you can open the action menu (…) for the appropriate alert and select **Mute**. To permanently suppress actions for an alert, open the actions menu and select **Mark as untracked**.
+If an alert is active or flapping, you can mute it to temporarily suppress future actions. In both **{{stack-manage-app}} > Alerts** and **{{rules-ui}}**, you can open the action menu (…) for the appropriate alert and select **Mute**. To permanently suppress actions for an alert, open the actions menu and select **Mark as untracked**. When an alert is marked as untracked, its status can no longer be changed.
```
lcawl, line 84: In the support issue that initiated this work, @JiaweiWu said "Muted alerts suppresses it's actions but it's lifecycle is still updated. Also it can be unmuted, untracked alerts are forever untracked and will no longer be updated: see this for more details https://github.com/elastic/kibana/issues/164059"

Maybe we can make this distinction more obvious? I've drafted a suggestion here but it ought to be confirmed with Jiawei or another member of the dev team:

```suggestion
If an alert is active or flapping, you can mute it to temporarily suppress future actions. In both **{{stack-manage-app}} > Alerts** and **{{rules-ui}}**, you can open the action menu (…) for the appropriate alert and select **Mute**. The status of the alert continues to be updated while it is muted but no actions run.

To permanently suppress actions for an alert, open the actions menu and select **Mark as untracked**.
In this case the alert status is no longer updated and actions no longer run. You cannot revert this change.

  
```

pmuellr: I believe that is technically correct, but there is a wrinkle.

If you just untrack an active alert, on the next run, if that alert would have been active again, a brand new alert will be created for it.  So from a high level view, that does not "mute" the alert (another one is created) and since it's a new alert, the actions will be run.

And then, oddly, if I mute the alert (which can only be done from the Alerts page off the Stack Management menu and not within the Alerts tab of the rule?), and then untrack it, it does actually mute that alert, but then you to go to the top-level Alerts page to unmute it again if you want to.  

@doakalexi can you double-check this?  Feels like we need mute/unmute controls in the Alerts tab of the Rule UX?

So ... it's confusing.

Mute/unmute alert is based on the alert id; over time, you can have multiple alerts that use the same alert id, but there can only be one active alert for an alert id at any point in time - the other ones would have been resolved or untracked (or closed or whatever).  So "muting an alert" from a rule's POV will mute any alert generated with that same alert id.  Untracking an alert will only "close" that particular version of the alert with that alert id. 

And I think muting only prevents actions from running, so I think the alerts will still be generated, but they won't run any actions.

But basically, "untracking" is for a single alert, "muting" is for all instances of alerts that use the same alert id, and muting really only deals with not running actions.

nastasha-solomon: Gotcha, thanks.

```suggestion
If an alert is active or flapping, you can mute it to temporarily suppress future actions. In **{{stack-manage-app}} > Alerts**, open the action menu (…) for the appropriate alert, then select **Mute**. While muted, the alert's status will continue to update but rule actions won't run. All future alerts with the same alert ID will also be muted.

To permanently suppress an alert's actions, open the actions menu for the appropriate alert, then select **Mark as untracked**. In this case, the alert's status is no longer updated and actions are no longer run. These changes are only applied to the alert that you untracked and cannot be reverted. Future alerts with the same alert ID are unaffected.
```

pmuellr (approved):
LGTM

PR #2733, opened 2025-08-27
Update work-with-params-secrets.md

mmahacek:
Update note about using environment variables in lightweight monitors to wrap with quotes to ensure proper replacement.


mdbirnstiehl (approved):
LGTM, thanks!

PR #2734, opened 2025-08-27
Adds recommendation about choosing project profile

szabosteve:
## Overview

This PR adds guidance for selecting the appropriate project profile based on the use case. It outlines when to use the General Purpose (GP) profile versus the Vector Optimized (VO) profile.


lcawl:
> Suggesting additional changes to the Billing dimensions page: https://github.com/elastic/docs-content/blob/ffcbcf3bb6b57e11caf9aa5ad0442c6418561dc9/deploy-manage/cloud-organization/billing/elasticsearch-billing-dimensions.md

I had related updates already underway in https://github.com/elastic/docs-content/pull/2744 so I've added this suggestion there.

shubhaat:
I think this is ok for now, but I think there will be big changes coming to the VO optimized profile based on feedback from Shay. We want to de-emphasize Vector optimized a LOT, and I wonder if we can make it clearer in the documentation that use VO only for very specific needs - when you need high memory for ingest/search. + @giladgal and @petegaleotti  - I might want to rethink the docs updates once we know what we are doing in onboarding.

solutions/search/serverless-elasticsearch-get-started.md
```diff
@@ -55,14 +55,24 @@ Use your {{ecloud}} account to create a fully-managed {{es}} project:
 6. Once the project is ready, select **Continue**.
 
 ::::{tip}
+
+**The General Purpose (GP) profile is recommended for most search use cases such as:**
+
+- full text search
```
lcawl, line 61: Hyphen is used in https://www.elastic.co/docs/solutions/search/full-text so ideally both should be aligned one way or the other

```suggestion
- full-text search
```

solutions/search/serverless-elasticsearch-get-started.md
```diff
@@ -55,14 +55,24 @@ Use your {{ecloud}} account to create a fully-managed {{es}} project:
...
+- sparse vector-based semantic search (such as when using ELSER)
+- dense vectors compressed using BBQ (enabled by default)
```
lcawl, line 63: It seems from https://www.elastic.co/docs/reference/elasticsearch/index-settings/bbq#bbq-hnsw like it's only the default in 9.1 and later. If that's correct, we either need to qualify this statement or remove the assertion about default behaviour:
```suggestion
- dense vectors compressed using BBQ
```

lcawl:
> @szabosteve , @lcawl I suggest we stash this documentation update. @petegaleotti is working on a major change to the VO profile and competely removing this from the UI and deemphasizing it in the docs etc. Sorry for the thrash, and fast turnaround time.

Based on a chat with @petegaleotti today it seems like it's worth implementing the updates in `serverless-elasticsearch-get-started.md` for now and then revamping them again if the UI changes. IMO the changes in the `vector.md` file can be deferred for now.

solutions/search/vector/dense-vector.md
```diff
@@ -30,6 +30,17 @@ To implement dense vector search in {{es}}, you need both an index configuration
 2. **Query the index with k-NN search**
    - Use the [`knn` query](knn.md) to retrieve results based on vector similarity
 
+### General and vector optimized projects [vector-profiles]
```
szabosteve, line 33:
```suggestion
### General purpose and vector optimized projects [vector-profiles]
```

PR #2740, opened 2025-08-27
Warn that cloned templates are not automatically updated

karenzone:
Closes: #1851

### DIFF
--- a/solutions/observability/apm/index-lifecycle-management.md
+++ b/solutions/observability/apm/index-lifecycle-management.md
@@ -140,3 +140,9 @@ POST /traces-apm-default/_rollover/
 ## Namespace-level index lifecycle policies [apm-data-streams-custom-policy-namespace]
 
 It is also possible to create more granular index lifecycle policies that apply to individual namespaces. This process is similar to the above tutorial, but includes cloning and modify the existing index template to use a new `*@custom` component template.
+
+:::{important}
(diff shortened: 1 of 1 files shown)


mdbirnstiehl (approved):
LGTM!

solutions/observability/apm/index-lifecycle-management.md
```diff
@@ -140,3 +140,9 @@ POST /traces-apm-default/_rollover/
 ## Namespace-level index lifecycle policies [apm-data-streams-custom-policy-namespace]
 
 It is also possible to create more granular index lifecycle policies that apply to individual namespaces. This process is similar to the above tutorial, but includes cloning and modify the existing index template to use a new `*@custom` component template.
```
lucabelluccini, line 142:
```suggestion
It is also possible to create more granular index lifecycle policies that apply to individual namespaces. This process is similar to the above tutorial, but includes cloning the original index template(s) and use the index lifecycle you've created in it.
```

lucabelluccini (commented):
I've left some suggestions.
I would like to "reinforce" the message that cloning index template is really necessary only when a user wants a customization per namespace.

The other info we might share here is that this `@custom` component template technique can be used also for other customizations (not just ILM).

I would also warn users explicitly to never edit the built-in/provided index templates nor to add extra component templates on top of the existing ones, or change the order of the component templates as it might break the correct behavior of the Observability applications relying on the APM data.

PR #2767, opened 2025-09-01
[SEARCH] Improves vector search docs based on SEO guidelines

szabosteve:
## Description

This PR optimizes the content of the `Vector search` and `Dense vector` pages based on the SEO best practices outlined [here](https://stunning-adventure-qrvr1k2.pages.github.io/style-guide/seo/).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,11 @@
 # docs-builder default output
 .artifacts
 .DS_store
+
+# Add LLM/AI related files
+AGENTS.md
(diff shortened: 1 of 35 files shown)


leemthompo (approved):
LGTM, might be good to get an SEO-expert review too though to make sure we're not missing anything going forward 🤷 

One question I have is if the `TIP`s we use to disambiguate or reorient users at the tops of these pages aren't messing up SEO somehow 😨

PR #2773, opened 2025-09-01
[SEARCH] Improves kkn docs based on SEO guidelines

szabosteve:
## Description

This PR optimizes the content of the `kNN search` page based on the SEO best practices outlined [here](https://stunning-adventure-qrvr1k2.pages.github.io/style-guide/seo/).


leemthompo (approved):
Looks good! The rewritten content is much nicer now! Just a few ideas for some headings in line with the guidance.

I would say that the biggest lever we have to improving this content longer-term is to break this page up into nicer, more standalone sub-pages.

PR #2797, opened 2025-09-03
Update ec-remote-cluster-self-managed.md

jamie-wilson88:
changed "into" to "to"


kilfoyle (approved):
LGTM! 🚀 
Thanks @jamie-wilson88!

Issue #2218, opened 2025-07-22
Reconcile duplicate elasticsearch privileges pages


shainaraskas:
same page is in two places:

https://www.elastic.co/docs/deploy-manage/users-roles/cluster-or-deployment-auth/elasticsearch-privileges

https://www.elastic.co/docs/reference/elasticsearch/security-privileges

and the info is drifting. likely will need to keep the reference page and make the deploy-manage one an xref only.

Related to #2215

PR #2811, opened 2025-09-04
Adds an xlink to the toc that points to reconciled page

yetanothertw:
As part of replacing the Elasticsearch [Built-in roles page in the Deploy and manage section](https://www.elastic.co/docs/deploy-manage/users-roles/cluster-or-deployment-auth/built-in-roles) with its [counterpart in the Reference section](https://www.elastic.co/docs/reference/elasticsearch/roles) this PR will retire the existing page in the Deploy and manage section and instead link to the correct, deduplicated page in the 
Reference section
Relates #2218

### DIFF
--- a/deploy-manage/toc.yml
+++ b/deploy-manage/toc.yml
@@ -601,7 +601,8 @@ toc:
               - file: users-roles/cluster-or-deployment-auth/manage-authentication-for-multiple-clusters.md
           - file: users-roles/cluster-or-deployment-auth/user-roles.md
             children:
-              - file: users-roles/cluster-or-deployment-auth/built-in-roles.md
+              - title: "Roles"
+                crosslink: elasticsearch://reference/elasticsearch/roles.md
               - file: users-roles/cluster-or-deployment-auth/defining-roles.md
                 children:
                   - file: users-roles/cluster-or-deployment-auth/role-structure.md


deploy-manage/toc.yml
```diff
@@ -601,7 +601,8 @@ toc:
               - file: users-roles/cluster-or-deployment-auth/manage-authentication-for-multiple-clusters.md
           - file: users-roles/cluster-or-deployment-auth/user-roles.md
             children:
-              - file: users-roles/cluster-or-deployment-auth/built-in-roles.md
+              - title: "Roles"
```
shainaraskas, line 604: to differentiate from creating your own roles

```suggestion
              - title: "Built-in roles"
```

shainaraskas (approved):
beautiful! one toc labeling suggestion (hypothetically could extend to the target page for clarity)

deploy-manage/toc.yml
yetanothertw: Thanks, @shainaraskas , that's a good point! 

Fabrizio was wondering if we should somehow indicate to users that we're going to send them to the Reference section. Something like this maybe? Wdyt?
```suggestion
  - title: "Built-in roles (Reference)"
```

PR #2822, opened 2025-09-05
Clarify trusted application behavior and alert 

jalogisch:
The previous description for Trusted Applications was ambiguous. It contained these two statements:

* `Doesn’t monitor the application for threats, nor does it generate alerts, even if it behaves like malware, ransomware, etc.`
* `Might still generate malicious behavior alerts, if the application’s process events indicate malicious behavior.`

For users unfamiliar with the underlying mechanics of Elastic Defend, this created a logical conflict: How can an application that "isn't monitored" for threats still generate a "malicious behavior alert"? This lack of clarity made it difficult for users to understand the feature's true behavior and configure it with confidence.

### Solution

This change rewrites the description to be technically precise and to remove the ambiguity. The new text now clearly differentiates between two distinct layers of protection:

1.  **File-based Threat Analysis:** It clarifies that "trusting" an application disables the direct scanning of the application's binary file (its code and signature). This is the "blind spot" created for performance and compatibility reasons.
2.  **Behavioral Analysis:** It explains that a separate detection engine continuously monitors system-wide *patterns of activity*. The actions of a trusted application are still part of this monitoring, and an alert will be generated if its behavior matches a malicious pattern (e.g., ransomware-like file encryption).

By explicitly defining these two concepts, the documentation now accurately explains how a trusted application can be exempt from direct scanning while still being subject to behavioral monitoring. This resolves the contradiction and provides a much clearer picture for our users.


natasha-moore-elastic (approved):
Thanks for your contribution, @jalogisch – LGTM!

PR #2835, opened 2025-09-05
Update data lifecycle tooling applicability

shainaraskas:
the serverless tag on data stream lifecycle reads like data stream lifecycle is serverless only. we should remove the badge [so the section "inherits" the page-level availability](https://elastic.github.io/docs-builder/contribute/cumulative-docs/example-scenarios/#page-section-varies-product)


yetanothertw (approved):
Oh ooops, I think I got this wrong. For some reason I thought it only applied to serverless. 

Approved and thank you

Issue #1707, opened 2025-06-11
[Website]: broken formatting on Configure index management for deployment templates


shainaraskas:
### Type of issue

Other

### What documentation page is affected

https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-configure-templates-index-management

### What happened?

<img width="632" alt="Image" src="https://github.com/user-attachments/assets/3c0aa1b2-782a-4744-ba8d-5d105a4d138c" />

PR #2836, opened 2025-09-05
Fix index management for ece formatting

shainaraskas:
This PR was meant to be a simple formatting bug fix (https://github.com/elastic/docs-content/issues/1707)

Ended up restructuring the index management selection step for ece templates to break up the index management options from the procedures of setting them up, and adding some applicability information. also did a little formatting/nesting cleanup on the parent templates page

validated against a 3.8 support lab: 

image

QUESTION: couldn't find the "extensions" step for deployment templates (step 10 in the [create templates](http://localhost:3000/deploy-manage/deploy/cloud-enterprise/ece-configuring-ece-create-templates#ece-configuring-ece-create-templates-ui) topic). are we worried about this?

### DIFF
--- a/deploy-manage/deploy/cloud-enterprise/ece-configure-templates-index-management.md
+++ b/deploy-manage/deploy/cloud-enterprise/ece-configure-templates-index-management.md
@@ -37,16 +37,11 @@ Index curation (Curator) {applies_to}`stack: deprecated 6.7`
 
 To configure {{ilm}} as part of your deployment template:
 
-1. On the **Index Management** page, select the **{{ilm}}** index curation method.
+On the **Index Management** page, under **{{ilm}} ({{ilm-init}})**, specify the node attributes for your data configurations. 
 
-2. To configure {{ilm-init}}, specify the node attributes for your data configurations. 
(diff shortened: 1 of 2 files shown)


eedugon:
I'd like ECE devs or a PM (maybe @Kushmaro ) to review this part:

image

We are suggesting to configure node attributes for ILM, which is ok. But the examples we give (note_type: hot, etc) feel super outdated and would interfere with role based ILM, which is the recommended way to handle ILM for hot / warm / cold / frozen architectures.

So I'm not sure if this UI is a bit outdated, the doc is outdated, or what's the exact purpose of the doc and the UI here.

But I guess that with ECE templates a user could be able to do hot / warm / cold based ILM without touching this part or adding the attributes the doc mentions, as it works out of the box with node roles.

Also, still showing `curator` as an option feels similar.

What do you think @shainaraskas ? @Kushmaro ?

eedugon (approved):
The changes are very good and a clear improvement.

But still i'm concerned about the accuracy and usefulness of the doc about curation and ILM.

Shared my thoughts in comments. Maybe we could treat them as separate issues for the future. The current change is a clear improvement IMO.

PR #2841, opened 2025-09-05
Add optimization profiles to Elasticsearch Serverless billing

lcawl:
Relates to https://github.com/elastic/cloud/issues/146427

This PR updates https://www.elastic.co/docs/deploy-manage/cloud-organization/billing/elasticsearch-billing-dimensions with information about the impact of the `optimized_for` value.

It must not be merged until the related changes to the default behaviour occur.


petegaleotti:
> * **Project subtype or profile**: When you use the [API]({{cloud-serverless-apis}}operation/operation-createelasticsearchproject) to create projects, be aware that the `optimized_for` option affects the VCU usage and costs.

I recommend rewording this to "...option affects VCU allocation and costs."

>   The `vector` option is recommended only for uncompressed dense vectors ([dense_vector](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md) fields with `int4` or `int8` quantization strategies) and high dimensionality. Though the per VCU cost is the same for general purpose and vector profiles, the latter provides a larger amount of RAM for searchable data. This leads to higher VCU consumption in order to improve the performance for uncompressed vector data.

I recommend rewording this with, "...the latter allocates more VCUs for searchable data. This leads to...".

PR #2842, opened 2025-09-05
[Ingest] Update title for data ingestion documentation

karenzone:
Update title to enhance reader scanability and improve SEO.

### Scanability
Leadership recently said, "To truly keep a modern enterprise secure, you need to **_ingest_**, **_retain_** and **_analyze_** all data." [emphasis mine]. "Ingest" is a word that users are likely to scan the TOC for as part of a normal workflow, so let's make it easy for them to find. 

#### SEO rankings
Search terms in headings get higher relevance rankings. By removing "ingest" or "ingestion" from the heading, we're missing an opportunity for better rankings.

### DIFF
--- a/manage-data/ingest.md
+++ b/manage-data/ingest.md
@@ -16,7 +16,7 @@ products:
   - id: elasticsearch
 ---
 
-# Bring your data to Elastic
+# Ingest: Bring your data to Elastic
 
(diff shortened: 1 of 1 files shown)


manage-data/ingest.md
```diff
@@ -16,7 +16,7 @@ products:
   - id: elasticsearch
 ---
 
-# Bring your data to Elastic
+# Ingest: Bring your data to Elastic
```
karenzone, line 19: I'd be OK with "ingestion," too.

PR #2854, opened 2025-09-08
Update create-detection-rule.md

bczifra:
Explicitly noted that `hits` aren't available in the threshold rule results context, indicated that `context.alerts.kibana.alert.threshold_result.terms` contains fields and values from any group by specified by the rule.


nastasha-solomon:
Thanks for opening this @bczifra! @yctercero can someone from the DE team review these changes?

solutions/security/detect-and-alert/create-detection-rule.md
```diff
@@ -152,7 +152,14 @@ To filter noisy {{ml}} rules, use [rule exceptions](/solutions/security/detect-a
...
         ::::{important}
-        Alerts created by threshold rules are synthetic alerts that do not resemble the source documents. The alert itself only contains data about the fields that were aggregated over (the **Group by** fields). Other fields are omitted, because they can vary across all source documents that were counted toward the threshold. Additionally, you can reference the actual count of documents that exceeded the threshold from the `kibana.alert.threshold_result.count` field.
+        Alerts created by threshold rules are synthetic alerts that do not resemble the source documents. The alert itself only contains data about the fields that were aggregated over (the **Group by** fields). Other fields are omitted, because they can vary across all source documents that were counted toward the threshold. Because of this, it isn’t possible to iterate on the search hits, so those aren’t available in the context. You can reference the actual count of documents that exceeded the threshold from the `kibana.alert.threshold_result.count` field. `context.alerts.kibana.alert.threshold_result.terms` contains fields and values from any group by specified by the rule. For example:
```
vitaliidm, line 155: we do not use terminology `hits` anywhere on this page for any of alert types.
Should we just focus on available fields from original documents in resulting alert?

florent-leborgne: What would be a better way to say this?

vitaliidm: Hits are results returned from ES search request. Hit == document in ES index.
For other rule types, we copy all existing fields in that document into alert.
For threshold rule, only **Group by** fields are getting copied.

In what context `hits` are referred here? Particularly `Because of this, it isn’t possible to iterate on the search hits, so those aren’t available in the context.`

florent-leborgne:
```suggestion
        Alerts created by threshold rules are synthetic alerts that do not resemble the source documents:
        
          - The alert itself only contains data about the fields that were aggregated over (the **Group by** fields specified in the rule).
          - All other fields are omitted and aren't available in the alert. This is because these fields can vary across all source documents that were counted toward the threshold. 
          - You can reference the actual count of documents that exceeded the threshold from the `kibana.alert.threshold_result.count` field. 
          - `context.alerts.kibana.alert.threshold_result.terms` contains fields and values from any **Group by** fields specified in the rule. For example:
```
How about we structure the note a little bit with something like this (because it's getting too big and hard to read). That way, the important information is easier to read and understand.
Lmk how this looks to you

vitaliidm: @florent-leborgne 

I like this update

florent-leborgne: I'm committing this change, please take a look at the updated preview and let me know how this now looks to you both @vitaliidm @bczifra

vitaliidm (approved):

Issue #2796, opened 2025-09-03
[ResponseOps][Docs]: Remove technical preview warning from maintenance window docs


js-jankisalvi:
### Description

We are making maintenance windows GA in 9.2, for which we need to update current MW docs to remove technical preview warning.

<img width="904" height="413" alt="Image" src="https://github.com/user-attachments/assets/b54b5ac2-d91a-419a-b8ee-4afb653e078c" />

### Resources

[Issue link](https://github.com/elastic/kibana/issues/233643) 

[PR link](https://github.com/elastic/kibana/pull/233870)

[doc link](https://www.elastic.co/docs/explore-analyze/alerts-cases/alerts/maintenance-windows)

### Which documentation set does this change impact?

Elastic On-Prem and Cloud (all)

### Feature differences

It is identical in all deployments.

### What release is this request related to?

9.2

### Serverless release

Sep 15, 2025

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @js-jankisalvi 

**Stakeholders:** @elastic/response-ops

PR #2855, opened 2025-09-08
[ResponseOps][Docs]: Maintenance window GAing and category options removed

nastasha-solomon:
Closes https://github.com/elastic/docs-content/issues/2796. Updates the maintenance windows docs as follows:
- Updates the page-level `applies to` tags to show that the feature is in tech preview for 9.0-91 and GA in 9.2 and Serverless.
- Removed the note at the top of the page that showed the feature was in tech preview. 
- Adds a note that explains that category-specific maintenance window options are unavailable (removed) in 9.2 and Serverless. Also updated the first item in the note further down that explains how you can select only a single category when you turn on filters.
- Refreshes both images on the page since they show elements that are no longer available in 9.2 and Serverless (the tech preview label and the option to choose categories for the maintenance windows

Preview: https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/2855/explore-analyze/alerts-cases/alerts/maintenance-windows


explore-analyze/alerts-cases/alerts/maintenance-windows.md
```diff
@@ -52,6 +48,15 @@ When you create a maintenance window, you must provide a name and a schedule. Yo
 
 By default, maintenance windows affect all categories of rules. The category-specific maintenance window options alter this behavior. For the definitive list of rule types in each category, refer to the [get rule types API](https://www.elastic.co/docs/api/doc/kibana/group/endpoint-alerting).
 
+::::{note}
+
+{applies_to}`stack: removed 9.2` {applies_to}`serverless: removed` The option to specify rule categories for a maintenance window is no longer available. Maintenance windows apply to all rule types.
+
+Note that existing maintenance windows are still applied to rule categories that were specified. However, if you edit a maintenance window after upgrading or using the latest version of {{serverless-short}}, specified rule categories are removed and the maintenance window will be applied to all rules types.
```
js-jankisalvi, line 55: Sorry my bad for not pointing out earlier, that removing category will cause MW to apply to all solutions, not all rule types.
nit: `specified rule categories are removed and the maintenance window will be applied to all rules types.` -> `specified rule categories are removed and the maintenance window will be applied to rules in all solutions.`

PR #2894, opened 2025-09-10
Update ece-ech-stack-monitoring.md bubble up limitation on service provider, version, region.

predogma:
The warning that the monitoring deployment `The monitoring deployment and production deployment must be on the same major version, cloud provider, and region.` needs to be up front at beginning and not hidden were it is after step 5 in enable section.  It also really needs to be in the limitations section as well


benironside (approved):
LGTM once you incorporate the other comment

PR #2900, opened 2025-09-10
update image to fix snapshot wording

georgewallace:


kilfoyle (approved):
LGTM! 🏎️ 
Though I do kind of like "snapshop". :-)

Issue #1642, opened 2025-06-06
[Internal]: Discrepancy between docs and UI for remote cluster API key instructions


shainaraskas:
### Description

from slack

> a customer had issues trying to follow the docs for setting up remote clusters using an API key. 
> In step 4 of this [section](https://www.elastic.co/docs/deploy-manage/remote-clusters/ec-remote-cluster-same-ess#ec_add_the_cross_cluster_api_key_to_the_local_deployment), it references Setting name and Secret  but the actual fields (in the screenshot below) are different. Also in the screenshot, it said the name must match the remote cluster name in this deployment's Stack Management -> Remote Clusters settings . But there was no step up to this point to add a remote cluster in the local cluster.

![Image](https://github.com/user-attachments/assets/4ac6e8ee-9b33-4c9a-9cfb-4d60ac9566d7)

### Resources

-

### Which documentation set does this change impact?

Elastic Cloud Hosted only

### Feature differences

-

### What release is this request related to?

N/A

### Serverless release

-

### Collaboration model

The documentation team

### Point of contact.

-

PR #2966, opened 2025-09-16
remote clusters UI api key and alias updates

eedugon:
Fixes https://github.com/elastic/docs-content/issues/1642


shainaraskas (approved):
nice to see these fixed! don't need to do it now, but because this content is essentially repeated, it might be nice to do with snippets as I mentioned [here](https://github.com/elastic/docs-projects/issues/437#issuecomment-3299829925)

PR #2994, opened 2025-09-17
Add entry for ingest pipeline field access pattern feature

jbaiera:
Adds a docs section explaining the new field access pattern feature on ingest pipelines.


karenzone (approved):
LGTM

PR #2795, opened 2025-09-03
Add new Elastic intro section

theletterf:
This rebuilds the **Get Started** section into the Elastic fundamentals section to improve Onboarding experience.

* Existing files and URLs have been preserved to minimize disruption.
* Added a new Next steps and resources topic with instructions on how to use the docs.

Pending:

* Moving the Glossary (more than 250 links to edit) or just adding a crosslink entry in the Elastic fundamentals' nav.
* Rewriting the Solutions landing page together with @shainaraskas.

PR #3004, opened 2025-09-17
Update versioning and availability

lcawl:
Relates to https://github.com/elastic/docs-content/pull/2795

This PR updates the introductory [Versioning and availability](https://www.elastic.co/docs/get-started/versioning-availability) page.
It is a work in progress until we address the following feedback:

- https://github.com/elastic/docs-content/pull/2795#discussion_r2321576573: "I don't know if this [Examples of where availability can vary] section is very clear. I think it could use some more details or introductory content. Somewhere on this page, we should try to explain a few things, how these multiple, differently versioned, products work together and how this can impact availability:
    - Elastic Cloud, ECE, and ECK are orchestrators used to create and manage serverless projects (Elastic Cloud only) and deployments of the Elastic Stack
    - The versions of the Elastic Stack that you can deploy depend on the version of these orchestrators. For example, Elastic Stack 9.0 is not available to deploy on ECE 3.1.
    - Because some products are released on different schedules and have different architecture or network constraints, it is possible that a feature isn't available for some deployment types, or becomes available at a different time for each deployment type."
- https://github.com/elastic/docs-content/pull/2795#discussion_r2322287185 "we also do some of this explanation on the following surfaces and you can mine them for language / make sure they align:
    - https://www.elastic.co/docs/deploy-manage/deploy#about-orchestration
    - https://www.elastic.co/docs/deploy-manage/deploy#versioning-and-compatibility
    - https://www.elastic.co/docs/deploy-manage/upgrade/plan-upgrade"
- https://github.com/elastic/docs-content/pull/2795#discussion_r2322287185 "> Because some products are released on different schedules... would be careful with this because a lot of functionality is controlled by the stack version. might indicate that these are usually at the admin layer rather than the storage/querying layer of elastic. examples would be helpful (e.g. not all auth types are available for all deployment types [e.g. [ldap](https://www.elastic.co/docs/deploy-manage/users-roles/cluster-or-deployment-auth/ldap)] - that's kind of a layer "above" / "around" the es engine)"


get-started/versioning-availability.md
```diff
@@ -87,130 +91,52 @@ To learn what's changed, check the [release notes](/release-notes/index.md) for
...
-
-
-
+| ----- | ----- |
+| {{stack}} versions | [{{stack}}](/get-started/the-stack.md) version 9.0.0 and later, including 9.1.0 |
+| Deployment types | [{{serverless-full}}](/deploy-manage/deploy/elastic-cloud/serverless.md), [{{ech}}](/deploy-manage/deploy/elastic-cloud/cloud-hosted.md), [{{ece}}](/deploy-manage/deploy/cloud-enterprise.md), [{{eck}}](/deploy-manage/deploy/cloud-on-k8s.md), and [Self-managed deployments](/deploy-manage/deploy/self-managed.md) |
+| Orchestrator versions | [{{ece}}](/deploy-manage/deploy/cloud-enterprise.md) 4.0.0 and later, [{{eck}}](/deploy-manage/deploy/cloud-on-k8s.md) 3.0.0 and later |
+| Serverless project types | {{es}}, Elastic {{observability}}, and {{elastic-sec}} |
+
+% will replace this with a ref link.
```
georgewallace, line 140: Is this section of the docs not ready or moving? Until then where should customers find this info?

georgewallace (approved):
I'm good with the changes, I just have one question. If the content at the bottom is removed we need to find a proper place for it to not have loss of information.

PR #3033, opened 2025-09-18
Add redirects for deduplicated roles and privileges pages

shainaraskas:
Followup to the following PRs:

https://github.com/elastic/docs-content/pull/2989
https://github.com/elastic/docs-content/pull/2811

We forgot a redirect and so now when someone looks for these pages in google they run into a sad 404


yetanothertw:
Thank you for fixing this!

PR #3049, opened 2025-09-19
[Private preview] Unhide Linux as an agent installation option

wajihaparvez:
Linux is now available as an installation option for Elastic agent in AutoOps for self-managed clusters. 

Closes https://github.com/elastic/docs-content/issues/3047


arnonst (approved):
Approved

kilfoyle (approved):
LGTM! 🌵

PR #3055, opened 2025-09-19
Fixes formatting for code block

yetanothertw:
The Rollover page includes a code block that doesn't render as expected, so I'm removing the `console` option.


kilfoyle (approved):
LGTM! 🦘 
I'm glad you were able to fix that!

PR #3059, opened 2025-09-19
reword to clarify when people should consider index size

shainaraskas:
- fixes https://github.com/elastic/docs-content-internal/issues/32

clarify that people should consider index size / growth when they're designing their indices, not when they're about to reach their max size.


kilfoyle (approved):
LGTM! 🥕

Issue #1573, opened 2025-05-31
[Website]: Document join field type as a difference on Serverless


ppf2:
### Type of issue

Missing information

### What documentation page is affected

https://www.elastic.co/docs/deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings

### What happened?

Join field type is not available on Serverless, while https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/parent-join does have the unavailable tag, let's also document this on the differences page.  Thx!

PR #3060, opened 2025-09-19
Document in serverless diffs that `join` fields are unavail

shainaraskas:
- fixes https://github.com/elastic/docs-content/issues/1573
- [join field type](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/parent-join) is unavailable in serverless in the reference, but not surfaced in the serverless comparison doc

@elastic/developer-docs is my info correct that ESQL LOOKUP JOIN is the serverless alternative?


kilfoyle (approved):
LGTM! 🍅

leemthompo (approved):
Got 👍 from @alex-spies about using LOOKUP JOIN instead

PR #3122, opened 2025-09-23
ECE load balancers: remove proxy protocol references from on‑prem guidance

igor-kupczynski:
Removed mentions of proxy protocol from ece-load-balancers.md

Proxy protocol isn’t used in on‑prem ECE load balancer setups; references were misleading.


AlexP-Elastic (commented):
Suggested we also fix the 9443 line (which is TCP mode as well)

kilfoyle (approved):
LGTM! 🏎️

PR #3151, opened 2025-09-24
[Remote Clusters] Connection modes and Security models presented in their own pages for all deployment types

eedugon:
WIP.

This PR presents RCS connection modes (proxy & sniff) and security models (TLS certificates & API key) properly within Remote Clusters sections, as currently the detailed information was only available in the self-managed configuration guide for remote clusters.

Pending items after validating the pages:

- Update links in other docs-content content (and other repos) to the new docs instead of the self-managed configuration guides.

- Remote cluster settings are still part of the self-managed, but eventually we will move them to Elasticsearch reference docs, so everything will look better then. For the moment we will keep the links to the self-managed item.

- Analyze Elasticsearch docs that share more or less the same information and determine the best way to address it:
  -  If we try to use snippets and try to make the ES docs to look exactly the same as they are not
  -  If we update ES docs to refer to the new docs for details and focus more on the configuration / setup steps (this would be ideal but it will change the reader experience if they are used to the old docs).


deploy-manage/remote-clusters/security-models.md
```diff
@@ -0,0 +1,86 @@
+---
+navigation_title: Security models
+applies_to:
+  deployment:
+    ece: ga
+    ess: ga
+    eck: ga
+    self: ga
+  serverless: unavailable
```
shainaraskas, line 9:
```suggestion
  stack: ga
  serverless: unavailable
```

deploy-manage/remote-clusters/security-models.md
```diff
@@ -0,0 +1,86 @@
...
+products:
+  - id: elasticsearch
+---
+# Remote clusters security models
```
shainaraskas, line 13:
```suggestion
# Remote cluster security models
```

deploy-manage/remote-clusters/security-models.md
```diff
@@ -0,0 +1,86 @@
...
+
+Remote clusters security models determine how authentication and authorization works between clusters. {{es}} has evolved from a TLS certificate–based model, relying on mutual TLS authentication over the transport interface and duplicated roles across clusters, to a more flexible API key–based model that uses a dedicated service endpoint and supports fine-grained authorization on both local and remote clusters.
```
shainaraskas, line 15:
```suggestion
Remote cluster security models determine how authentication and authorization works between clusters. {{es}} has evolved from a TLS certificate–based model, relying on mutual TLS authentication over the transport interface and duplicated roles across clusters, to a more flexible API key–based model that uses a dedicated service endpoint and supports fine-grained authorization on both local and remote clusters.
```

shainaraskas, line 15: this is a big mouthful. is the evolution of this feature important? 

```
Remote cluster security models determine how authentication and authorization works between clusters. 

There are two models available: 
* [API ..]
* [TLS] (deprecated)

API key [benefits], so use it / migrate to it
```

deploy-manage/remote-clusters/security-models.md
```diff
@@ -0,0 +1,86 @@
...
+
+TLS certificate–based authentication is now deprecated, and users are encouraged to migrate to the API key–based model.
+
+The following sections describe both models in detail and highlight their key differences.
+
+::::{note}
```
shainaraskas, line 21:
```suggestion
::::{tip}
```

deploy-manage/remote-clusters/security-models.md
```diff
@@ -0,0 +1,86 @@
...
+Security models work independently of [connection modes](./connection-modes.md). Both security models are compatible with either connection mode.
+::::
+
+## API key authentication [api-key]
+
+API key authentication enables a local cluster to authenticate itself with a remote cluster via a [cross-cluster API key](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-security-create-cross-cluster-api-key). The API key needs to be created by an administrator of the remote cluster. The local cluster is configured to provide this API key on each request to the remote cluster. The remote cluster verifies the API key and grants access, based on the API key’s privileges.
```
shainaraskas, line 27: always avoid `via`
```suggestion
API key authentication enables a local cluster to authenticate itself with a remote cluster using a [cross-cluster API key](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-security-create-cross-cluster-api-key). The API key needs to be created by an administrator of the remote cluster. The local cluster is configured to provide this API key on each request to the remote cluster. The remote cluster verifies the API key, and grants access based on the API key’s privileges.
```

deploy-manage/remote-clusters/security-models.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Authorization
+
+With this security model, authorization is enforced jointly by the local and remote cluster, as follows:
+
+* All cross-cluster requests from the local cluster are bound by the API key’s privileges, regardless of local users associated with the requests. For example, if the API key only allows read access to `my-index` on the remote cluster, even a superuser from the local cluster is limited by this constraint. This mechanism enables the remote cluster’s administrator to have full control over who can access what data with cross-cluster search and/or cross-cluster replication. The remote cluster’s administrator can be confident that no access is possible beyond what is explicitly assigned to the API key.
```
shainaraskas, line 33: avoid `and/or` - we can use an "inclusive or"

```suggestion
* All cross-cluster requests from the local cluster are bound by the API key’s privileges, regardless of local users associated with the requests. For example, if the API key only allows read access to `my-index` on the remote cluster, even a superuser from the local cluster is limited by this constraint. This mechanism enables the remote cluster’s administrator to have full control over who can access what data with cross-cluster search or cross-cluster replication. The remote cluster’s administrator can be confident that no access is possible beyond what is explicitly assigned to the API key.
```

deploy-manage/remote-clusters/security-models.md
```diff
@@ -0,0 +1,86 @@
...
+
+* On the local cluster side, not every local user needs to access every piece of data allowed by the API key. An administrator of the local cluster can further configure additional permission constraints on local users so each user only gets access to the necessary remote data. Note it is only possible to further reduce the permissions allowed by the API key for individual local users. It is impossible to increase the permissions to go beyond what is allowed by the API key.
```
shainaraskas, line 35:
```suggestion
* On the local cluster side, not every local user needs to access every piece of data allowed by the API key. An administrator of the local cluster can further configure additional permission constraints on local users so each user only gets access to the necessary remote data. It is only possible to further reduce the permissions allowed by the API key for individual local users. It is impossible to increase the permissions to go beyond what is allowed by the API key.
```

deploy-manage/remote-clusters/security-models.md
```diff
@@ -0,0 +1,86 @@
...
+
+  ::::{tip}
+  To configure fine-grained authorization for remote resources, use the `remote_indices` and `remote_clusters` fields in [role definitions](/deploy-manage/users-roles/cluster-or-deployment-auth/role-structure.md).
+  ::::
+
+### Connection details
+
+In this model, cross-cluster operations use [a dedicated server port](elasticsearch://reference/elasticsearch/configuration-reference/networking-settings.md#remote_cluster.port) (remote cluster interface), for communication between clusters. The default port is `9443`. The remote cluster must enable this port for local clusters to connect.
```
shainaraskas, line 43: while these brackets make sense, we technically should not use them

```suggestion
In this model, cross-cluster operations use [a dedicated server port](elasticsearch://reference/elasticsearch/configuration-reference/networking-settings.md#remote_cluster.port), referred to as the remote cluster interface, for communication between clusters. The default port is `9443`. The remote cluster must enable this port for local clusters to connect.
```

deploy-manage/remote-clusters/security-models.md
```diff
@@ -0,0 +1,86 @@
...
+
+From a TLS perspective, the local cluster must trust the remote cluster on the remote cluster interface. This means the local cluster must trust the certificate authority (CA) that signs the server certificate used by the remote cluster interface. When establishing a connection, all nodes from the local cluster that participate in cross-cluster communication verify certificates from nodes on the other side, based on the TLS trust configuration.
+
+Mutual TLS is not required in this model.
+
+### Setup
+
+Refer to [Remote clusters setup](../remote-clusters.md#setup) for configuration guidance across all deployment types.
```
shainaraskas, line 51: each setup can be thought of on a one-to-one basis, so we should always make this singular except for when we refer to the concept as a whole

```suggestion
Refer to [Remote cluster setup](../remote-clusters.md#setup) for configuration guidance across all deployment types.
```

deploy-manage/remote-clusters/security-models.md
```diff
@@ -0,0 +1,86 @@
...
+In this model, a superuser on the local cluster gains total read access to the remote cluster, so it is only suitable for clusters that are in the same security domain.
+::::
+
+### Connection details
+
+The local cluster uses the [transport interface](elasticsearch://reference/elasticsearch/configuration-reference/networking-settings.md) to establish communication with remote clusters. The coordinating nodes in the local cluster establish [long-lived](elasticsearch://reference/elasticsearch/configuration-reference/networking-settings.md#long-lived-connections) TCP connections with specific nodes in the remote cluster. These connections must remain open, even if the connections are idle for an extended period.
+
+### Setup
+
+Refer to [Remote clusters setup](../remote-clusters.md#setup) for configuration guidance across all deployment types.
```
shainaraskas, line 74:
```suggestion
Refer to [Remote cluster setup](../remote-clusters.md#setup) for configuration guidance across all deployment types.
```

shainaraskas (approved):
generally looks really good. only have style comments really

one question: where did you source this information? is it duplicated somewhere where we should remove it and point to this as the official source of truth?

deploy-manage/remote-clusters/security-models.md
eedugon: @shainaraskas , I think understanding its evolution helps to:
- understand why all this is so complicated and why the modes are so different
- understand why we have 2 modes (it doesn't make much sense from design point of view aiming for the 2 modes since the beginning).
- understand why the legacy / deprecated mode shouldn't be used

But anyway I'm totally ok removing that part or simplifying it if you think it doesn't add any value.

deploy-manage/remote-clusters/connection-modes.md
```diff
@@ -0,0 +1,68 @@
...
+
+::::{note}
+Connection modes work independently of [security models](./security-models.md). Both connection modes are compatible with either security model.
+::::
+
+The choice between sniff and proxy mode depends on your network architecture and deployment type.  
+
+- **Self-managed environments:** If direct connections on the publish addresses between {{es}} nodes in both clusters are possible, you can use sniff mode. If direct connectivity is difficult to implement—for example, when clusters are separated by NAT, firewalls, or containerized environments—you can place a reverse proxy or load balancer in front of the remote cluster and use proxy mode instead.  
+
+- **Managed environments ({{ece}}, {{ech}}, {{eck}}):** Direct node-to-node connectivity is generally not feasible, so these deployments always rely on the proxy connection mode.
```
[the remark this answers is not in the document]
eedugon: It's not impossible, the remote in sniff mode can be configured without any restriction (with ES API), but it won't work in the majority of cases.

deploy-manage/remote-clusters/connection-modes.md
```diff
@@ -0,0 +1,68 @@
...
+::::
+
+
+## Proxy mode
+
+In proxy mode, a cluster alias is registered with a name of your choosing and the address of a TCP (layer 4) reverse proxy specified with the `cluster.remote.<cluster_alias>.proxy_address` setting. You must configure this proxy to route connections to one or more nodes of the remote cluster. The service port to forward traffic to depends on the [security model](./security-models.md) in use, as each model uses a different service port.
+
+When you register a remote cluster using proxy mode, {{es}} opens several TCP connections to the proxy address and uses these connections to communicate with the remote cluster. In proxy mode {{es}} disregards the publish addresses of the remote cluster nodes which means that the publish addresses of the remote cluster nodes do not need to be accessible to the local cluster.
+
+Proxy mode is not the default connection mode, so you must set `cluster.remote.<cluster_alias>.mode: proxy` to use it. See [Proxy mode remote cluster settings](remote-clusters-settings.md#remote-cluster-proxy-settings) for more information about configuring proxy mode.
```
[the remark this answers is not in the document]
eedugon: I've tried to make all clearer, including the statement that Kibana only works with proxy mode.

deploy-manage/remote-clusters/connection-modes.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Connection modes: comparison
+
+The following table summarizes the key differences between sniff and proxy mode to help you choose the most suitable option for your deployment.
+
+| Aspect                  | Sniff mode                                                                                       | Proxy mode                                                                                   |
+|-------------------------|--------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------|
+| **Defaul when adding a remote**             | Yes                                                                                              | No (must be explicitly configured)                                                           |
+| **Recommended deployment types**    | Self-managed                                                                                     | Self-managed, {{ech}}, {{ece}}, and {{eck}}                                                  |
```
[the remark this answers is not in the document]
eedugon: removed all comparison tables

shainaraskas (approved):
much clearer - thank you!

PR #3155, opened 2025-09-24
Rate limits are 500 RPM

seanhandley:
This was changed today to match initial plan of 100 docs/sec


yetanothertw (approved):
👍 Thank you for updating the docs!

PR #3157, opened 2025-09-24
Change 'index' to 'data stream' in retention task

kilfoyle:
Small fix. On the [Manage a data stream](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/3143/manage-data/data-store/data-streams/manage-data-stream) page and in this [Manage data streams](https://www.elastic.co/docs/manage-data/data-store/index-basics#manage-data-streams) section we mention "selecting an index" to update retention settings. This should be "data stream" instead (it's not an index setting).


yetanothertw (approved):
Looks good, thanks, David! 🚀

PR #3164, opened 2025-09-25
Fix anchor link By shutting down the host

jguay:
Just fixing anchor currently [By shutting down the host (less destructive)](https://www.elastic.co/docs/deploy-manage/maintenance/ece/perform-ece-hosts-maintenance#ece-perform-host-maintenance-delete-runner) to point to correct anchor that should be https://www.elastic.co/docs/deploy-manage/maintenance/ece/perform-ece-hosts-maintenance#ece-perform-host-maintenance-shutdown-host


yetanothertw (approved):
LGTM, thank you for fixing that! 🚀

Issue #2577, opened 2025-04-09
Document Re-enrollment when agents are moved to another cluster


michalpristas:
### DoD
- [ ] Docs covering dependent work is created
- [ ] Warning users to create sufficient number of fleet server prior to executing re-enrollment process is present

#### Dependencies
- [x] Depends on https://github.com/elastic/elastic-agent/issues/7772
- [x] Depends on https://github.com/elastic/kibana/issues/217616
- [x] Depends on https://github.com/elastic/kibana/issues/217619

#### Resources
Important background content in [parent topic](https://github.com/elastic/ingest-dev/issues/4805)

nimarezainia:
@karenzone as per our discussion.

juliaElastic:
Can we add the documentation soon? The feature is out in 9.2 version. cc @vishaangelova

PR #3216, opened 2025-09-29
[Fleet] Add docs for re-enrolling multiple agents after migration

karenzone:
Closes: #2577

* On 9.1.0 and earlier, you need to run commands on each individual host to re-enroll agents. 
* On 9.2.0 and later, you can migrate remote agents directly from the Fleet UI, eliminating the need to run commands on each individual host.


reference/fleet/migrate-elastic-agent.md
```diff
@@ -224,13 +226,26 @@ This is best performed one policy at a time. For a given policy, you need to cap
...
+
+**For 9.2.0 and later, you can migrate remote agents directly from the {{fleet}} UI, eliminating the need to run commands on each individual host.**
+
+5. In the source cluster, select the agents you want to migrate. Click the three dots next to the agents, and select **Migrate agents**.
+6. In the migration dialog, provide the URI and enrollment token you obtained from the target cluster.
+7. Use replace_token (Optional): When you are migrating a single agent, you can use the `replace_token` field to preserve the agent's original ID from the source cluster. This step helps with event matching, but will cause the migration to fail if the target cluster already has an agent with the same ID.
+:::
+
+:::{tab-item} 9.1.0
+**On 9.1.0 and earlier, you need to run commands on each individual host.**
```
colleenmcginnis, line 240: Using `9.2.0` and `9.1.0` as the tab titles might be confusing because:

* What about versions before 9.1.0 or 9.1.x versions after 9.1.0? It's not clear to the reader without interacting with the tab that it's possible to do this at all in 9.0.x or 9.1.1 - 9.1.5.
* For the most part, we have been using version tabs to specify the version a feature was added / made available rather than the last version in which is was available.
* Is it still possible to run commands on each individual host in 9.2.0 or does the user _have_ to use the Fleet UI starting in 9.2.0? (Note: My recommended approach below assumes it's possible to use the command OR UI in 9.2.0+.)
* Hard-coding the `9.2.0` in the tab title and in the text means that you need to hold on merging this PR until the 9.2.0 release date or we will be making promises about future releases.

I wonder if it would be clearer to do something like this:

Screenshot 2025-10-07 at 11 49 27 AM

And the `Command line` tab:

Screenshot 2025-10-07 at 11 49 40 AM

<details>
<summary>Markdown syntax</summary>

```markdown
5. Choose an approach:

    ::::{tab-set}
    :::{tab-item} Fleet UI

    {applies_to}`stack: ga 9.2` Migrate remote agents directly from the {{fleet}} UI:

    1. In the source cluster, select the agents you want to migrate. Click the three dots next to the agents, and select **Migrate agents**.
    2. In the migration dialog, provide the URI and enrollment token you obtained from the target cluster.
    3. Use replace_token (Optional): When you are migrating a single agent, you can use the `replace_token` field to preserve the agent's original ID from the source cluster. This step helps with event matching, but will cause the migration to fail if the target cluster already has an agent with the same ID.
    :::

    :::{tab-item} Command line
    Run commands on each individual host:

    1. On the host machines where the current agents are installed, enroll the agents again using this copied URL and the enrollment token:

        ```shell
        sudo elastic-agent enroll --url=<fleet server url> --enrollment-token=<token for the new policy>
        ```

        The command output should resemble this:

        :::{image} images/migrate-agent-install-command-output.png
        :alt: Install command output
        :screenshot:
        :::

    2. The agent on each host will now check into the new {{fleet-server}} and appear in the new target cluster. In the source cluster, the agents will go offline as they won’t be sending any check-ins.

        :::{image} images/migrate-agent-newly-enrolled-agents.png
        :alt: Newly enrolled agents in the target cluster
        :screenshot:
        :::

    3. Repeat this procedure for each {{agent}} policy.
    :::
    ::::
```

</details>

reference/fleet/migrate-elastic-agent.md
```diff
@@ -210,7 +208,11 @@ The agents can now be re-enrolled into these policies and migrated over to the n
 
 ## Migrate {{agent}}s to the new target cluster [migrate-elastic-agent-migrated-agents]
 
-In order to ensure that all required API keys are correctly created, the agents in your current cluster need to be re-enrolled into the new, target cluster.
+::::{note}
+Agents to be migrated cannot be tamper-protected or running as a {{fleet}} server.
```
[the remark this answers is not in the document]
nimarezainia: @michalpristas @karenzone shoul we have a list of caveats? even if they are in the works and we plan to address them in the future, just highlighting to the user that these caveats exist.

I think tamper protection we would never want to support. by definition, tamper protection means ".. don't .. touch me ...". Fleet server is another great example. Anyway if there are more let's list them out.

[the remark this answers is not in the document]
karenzone: @nimarezainia, my suggestion is to merge what we have if it is correct, and then do a fast follow with the additional caveats and examples.  Those details can be slow to trickle in, and it takes some time to build consensus.  Let's not hold up the core value while we wait for the _value added_.  Please LMKWYT.

Who are the SMEs for this additional information?

reference/fleet/migrate-elastic-agent.md
```diff
@@ -208,10 +208,6 @@ The agents can now be re-enrolled into these policies and migrated over to the n
 
 ## Migrate {{agent}}s to the new target cluster [migrate-elastic-agent-migrated-agents]
 
-::::{note}
-Agents to be migrated cannot be tamper-protected or running as a {{fleet}} server.
-::::
-
```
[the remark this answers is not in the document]
michalpristas: yes still applies.

michalpristas:
i dont think so, "migration blocked during upgrade" as i originally referred to as "this will be done later" was agreed on in a team won't be supported and we will allow to trigger it.

reference/fleet/migrate-elastic-agent.md
colleenmcginnis: The official format is Fleet Server. You can use that directly or use the `fleet-server` sub.

```suggestion
Agents to be migrated cannot be tamper-protected or running as a {{fleet-server}}.
```

reference/fleet/migrate-elastic-agent.md
```diff
@@ -224,27 +226,44 @@ This is best performed one policy at a time. For a given policy, you need to cap
...
 
-    The command output should be like the following:
+    {applies_to}`stack: ga 9.2` Migrate remote agents directly from the {{fleet}} UI:
 
-    :::{image} images/migrate-agent-install-command-output.png
-    :alt: Install command output
-    :screenshot:
+    1. In the source cluster, select the agents you want to migrate. Click the three dots next to the agents, and select **Migrate agents**.
+    2. In the migration dialog, provide the URI and enrollment token you obtained from the target cluster.
+    3. Use replace_token (Optional): When you are migrating a single agent, you can use the `replace_token` field to preserve the agent's original ID from the source cluster. This step helps with event matching, but will cause the migration to fail if the target cluster already has an agent with the same ID.
```
colleenmcginnis: This would just make it clearer that you're referencing a specific field:

```suggestion
    3. Use `replace_token` (Optional): When you are migrating a single agent, you can use the `replace_token` field to preserve the agent's original ID from the source cluster. This step helps with event matching, but will cause the migration to fail if the target cluster already has an agent with the same ID.
```

reference/fleet/migrate-elastic-agent.md
```diff
@@ -224,27 +226,44 @@ This is best performed one policy at a time. For a given policy, you need to cap
...
+        :::{image} images/migrate-agent-newly-enrolled-agents.png
+        :alt: Newly enrolled agents in the target cluster
+        :screenshot:
+        :::
+
+     3. Repeat this procedure for each {{agent}} policy.
+     :::
+     ::::
+
+If all has gone well, you’ve successfully migrated your {{fleet}}-managed {{agent}}s to a new cluster.
```
colleenmcginnis: This is a personal preference, but to me the phrase "If all has gone well" suggests that it might not go well. I would rephrase this sentence or remove it entirely. 

Alternatively, if there are reasons why it might not "go well" and we have troubleshooting documentation on this topic, we could point to that here.

reference/fleet/migrate-elastic-agent.md
```diff
@@ -224,27 +226,44 @@ This is best performed one policy at a time. For a given policy, you need to cap
...
```
karenzone:
```suggestion
```

Issue #3179, opened 2025-09-25
Additional improvements to time series data stream docs


marciw:
Not yet prioritized

specific:
- [ ] link to EDOT quickstart from TSDS quickstart + TSDS OTLP page (https://github.com/elastic/docs-content/pull/4219#discussion_r2613755272)
- [ ] OTLP: edit more thoroughly
- [ ] Reindex: explain `op_type` (add something like "ensures the operation is idempotent")
- [ ] Reindex: update to "better path" mentioned [here](https://github.com/elastic/docs-content/pull/3222#discussion_r2405376739) 
- [ ] Querying DS page title: change to `# Querying downsampled data in {{es}}` (see [comment](https://github.com/elastic/docs-content/pull/3461/files/268915f8a98fa06f10bdf96718109eaecad6ace8#r2432530573))
- [ ] Double-check deployment options and recommended approaches, and separate from true best practices

general:
- [ ] More glue (and consistency) with general data streams docs
- [ ] Other cross-linking opportunities
- [ ] More ES|QL TS command coverage (examples etc.)
- [ ] Any additional advanced operations

PR #3222, opened 2025-09-30
[WIP] Remaining TSDS edits

marciw:
# Work in progress

Status
🟢  Ready for PM/engineer review
🚧  Not ready for tech writer review

Changes
- Revised overview
- Revised setup
- New advanced section 

Notes for reviewers: We're going for "MVP" docs for now and tracking additional improvements in https://github.com/elastic/docs-content/issues/3179

TODO: 
- [ ] Reconcile with recent changes to general data stream docs
- [ ] Check docs patterns/style/etc.


manage-data/data-store/data-streams/reindex-tsds.md
```diff
@@ -9,208 +9,111 @@ products:
...
+2. Highest timestamp value in the old data stream
+3. Sum of the primary shards from all source backing indices
+4. Speed up reindexing
+5. Pause ILM
 
-## Reindex [tsds-reindex-op]
+### Create the destination data stream and reindex [tsds-reindex-op]
 
-Invoke the reindex api, for instance:
+Run the reindex operation using `op_type: create` to prevent overwrites:
```
kkrik-es, line 107: Maybe skip `to prevent overwrites` ? This is a new data stream, so no overwriting is possible?

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -11,63 +11,68 @@ products:
...
+Before setting up a time series data stream, make sure you're familiar with general [data stream](/manage-data/data-store/data-streams.md) concepts.
 
-## When to use a TSDS [when-to-use-tsds]
+## When to use a time series data stream [when-to-use-tsds]
 
-Both a [regular data stream](../data-streams.md) and a TSDS can store timestamped metrics data. Only use a TSDS if you typically add metrics data to {{es}} in near real-time and `@timestamp` order.
+Both a regular data stream and a time series data stream can store timestamped metrics data. 
 
-Use a time series data stream for metrics data only. For other timestamped data, such as logs or traces, use a [logs data stream](logs-data-stream.md) or regular data stream.
+Choose a time series data stream if you typically add metrics data to {{es}} in near real-time and in `@timestamp` order. For other timestamped data, such as logs or traces, use a [logs data stream](logs-data-stream.md) or [regular data stream](/manage-data/data-store/data-streams.md).
```
kkrik-es, line 24: Consider expanding what metrics data means. Here, we're looking for a sequence of data point-timestamp pairs, identified by one or more dimension fields that can be used for slicing in aggregation queries.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -11,63 +11,68 @@ products:
...
 
-## Differences from a regular data stream [differences-from-regular-data-stream]
+A time series data stream works like a regular data stream, with some key differences:
 
-A TSDS works like a regular data stream with some key differences:
+* **Time series index mode:** The matching index template for a TSDS must include a `data_stream` object with `index.mode` set to `time_series`. This option enables most TSDS-related functionality.
+* **Required fields:** In a TSDS, each document contains:
+  * A `@timestamp` field
+  * One or more [dimension fields](#time-series-dimension), set with `time_series_dimension: true`  
+  * One or more [metric fields](#time-series-metric) (not strictly required, but typical for a TSDS)
```
kkrik-es, line 34: Let's remove the `not strictly required` part. A time-series requires a metric field with non-null values.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -11,63 +11,68 @@ products:
...
+* **Document IDs:** Time series documents use two IDs: 
+    * An internal [`_tsid`](#tsid) metadata field, generated by {{es}} for each document in a TSDS and used for sorting and compression
```
kkrik-es, line 36: Still thinking whether we need to expose the `_tsid` so readily in our documentation.. If we do, we probably want to mention that it's calculated over all dimension values.

Another option is to have a section towards the end, shedding some light into how data gets structured internally.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -11,63 +11,68 @@ products:
...
+    * The document `_id`, a generated hash of the document's dimensions and `@timestamp` (custom `_id` values are not supported)
```
kkrik-es, line 37: Maybe mention that it's auto-generated - passing a doc id during indexing results to an error.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -11,63 +11,68 @@ products:
...
+* **Backing indices:** A TSDS uses [time-bound indices](/manage-data/data-store/data-streams/time-bound-tsds.md) to store data from the same time period in the same backing index.
+* **Dimension-based routing:** The matching index template for a TSDS must contain the `index.routing_path` index setting, which specifies dimensions for routing documents to shards.
+* **Sorting:** A TSDS uses internal [index sorting](elasticsearch://reference/elasticsearch/index-settings/sorting.md) to order shard segments by `_tsid` and `@timestamp`, for better compression. Time series data streams do not use `index.sort.*` settings.
+* **Synthetic source:** A TSDS uses [synthetic `_source`](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#synthetic-source), which has some [restrictions](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#synthetic-source-restrictions) and [modifications](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#synthetic-source-modifications).
```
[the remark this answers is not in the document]
kkrik-es: One difference, though, is that it's not possible to [disable source](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/mapping-source-field#disable-source-field) for a TSDS. It's either standard or synthetic source mode.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
kkrik-es: This is nicely described in `Time-series concepts` below. Maybe add a cross reference, or have this section follow that one?

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -11,63 +11,68 @@ products:
...
+- Temperature and humidity readings from a weather sensor
 
-:::{image} /manage-data/images/elasticsearch-reference-time-series-chart.svg
+:::{image} /manage-data/images/elasticsearch-reference-time-series-chart2.svg
 :alt: time series chart
 :title: Time series of weather sensor readings plotted as a graph
 :::
 
-In a TSDS, each {{es}} document represents an observation, or data point, in a specific time series. Although a TSDS can contain multiple time series, a document can only belong to one time series. A time series can’t span multiple data streams.
+In a TSDS, each {{es}} document represents an observation, or data point, in a specific time series. Although a TSDS can contain multiple time series, a document can belong to only one time series. A single time series can't span multiple data streams.
```
kkrik-es, line 61: That's not 100% correct. The proper definition of a time series includes the metric name. Since we can have multiple metric fields populated in a single doc, these map to multiple time series. @felixbarny too for thoughts here.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -11,63 +11,68 @@ products:
...
+
+### Time series fields
 
+Compared to a regular data stream, a TSDS uses some additional fields specific to time series:  dimension fields (required) and metric fields (optional but usually defined), plus an internal `_tsid` metadata field.
```
kkrik-es, line 65: Ditto, let's not call metric fields optional.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -78,146 +83,56 @@ You mark a field as a dimension using the boolean `time_series_dimension` mappin
...
+To work with a flattened field, use the `time_series_dimensions` parameter to configure an array of fields as dimensions. For details, refer to [`flattened`](elasticsearch://reference/elasticsearch/mapping-reference/flattened.md#flattened-params).
 
+You can also simplify dimension definitions by using [pass-through](elasticsearch://reference/elasticsearch/mapping-reference/passthrough.md#passthrough-dimensions) fields.
+:::
 
-### Metrics [time-series-metric]
+#### Metrics [time-series-metric]
 
-Metrics are fields that contain numeric measurements, as well as aggregations and/or downsampling values based off of those measurements. While not required, documents in a TSDS typically contain one or more metric fields.
+Metrics are numeric measurements that change over time. Although metrics are not required, documents in a TSDS typically contain one or more metric fields.
```
kkrik-es, line 94:
```suggestion
Metrics are numeric measurements that change over time. Documents in a TSDS typically contain one or more metric fields. 
```

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -78,146 +83,56 @@ You mark a field as a dimension using the boolean `time_series_dimension` mappin
...
 
-Metrics differ from dimensions in that while dimensions generally remain constant, metrics are expected to change over time, even if rarely or slowly.
+:::{tip}
+Metrics are expected to change (even if rarely or slowly), while dimensions generally remain constant.
+:::
 
-To mark a field as a metric, you must specify a metric type using the `time_series_metric` mapping parameter. The following field types support the `time_series_metric` parameter:
+To mark a field as a metric, use the `time_series_metric` mapping parameter. This parameter ensures data is stored in an optimal way for time series analysis. The following field types support the `time_series_metric` parameter:
 
 * [`aggregate_metric_double`](elasticsearch://reference/elasticsearch/mapping-reference/aggregate-metric-double.md)
```
kkrik-es, line 102: Nit: move this second, it's very rare that users populate it. It gets internally generated during downsampling, mostly.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -78,146 +83,56 @@ You mark a field as a dimension using the boolean `time_series_dimension` mappin
...
-
-TSDS documents don’t support a custom `_routing` value. Similarly, you can’t require a `_routing` value in mappings for a TSDS.
-
-
-### Index sorting [tsds-index-sorting]
-
-{{es}} uses [compression algorithms](elasticsearch://reference/elasticsearch/index-settings/index-modules.md#index-codec) to compress repeated values. This compression works best when repeated values are stored near each other — in the same index, on the same shard, and side-by-side in the same shard segment.
-
-Most time series data contains repeated values. Dimensions are repeated across documents in the same time series. The metric values of a time series may also change slowly over time.
+- You **can't** query or update the internal `_tsid` field.
```
kkrik-es, line 120: I'd skip these 3 points and just keep the last one, to highlight why it should not be used in queries.

manage-data/data-store/data-streams/set-up-tsds.md
```diff
@@ -86,28 +94,26 @@ PUT _ilm/policy/my-weather-sensor-lifecycle-policy
...
-* Define index settings:
+The structure of a time series data stream is defined by an index template. Create an index template with the following required elements and settings:
 
-    * Set `index.mode` setting to `time_series`.
-    * Your lifecycle policy in the `index.lifecycle.name` index setting.
-    * Optional: Other index settings, such as [`index.number_of_replicas`](elasticsearch://reference/elasticsearch/index-settings/index-modules.md#dynamic-index-number-of-replicas), for your TSDS’s backing indices.
+- **Index patterns:** One or more wildcard patterns matching the name of your TSDS, such as `weather-sernsors-*`. For best results, use the [data stream naming scheme](/reference/fleet/data-streams.md#data-streams-naming-scheme).
+- **Data stream object:** The template must include `"data_stream": {}`.
+- **Time series mode:** Set `index.mode: time_series`.
+- **Field mappings:** Define at least one `keyword` dimension field and typically one or more metric fields:
```
kkrik-es, line 109:
```suggestion
- **Field mappings:** Define at least one dimension field and typically one or more metric fields:
```

manage-data/data-store/data-streams/set-up-tsds.md
```diff
@@ -86,28 +94,26 @@ PUT _ilm/policy/my-weather-sensor-lifecycle-policy
...
+    - To define a dimension, set `time_series_dimension` to `true`. Dimension fields like `counter` only increase over time. For more details, refer to [Dimensions](/manage-data/data-store/data-streams/time-series-data-stream-tsds.md#time-series-dimension).
+    - To define a metric, use the `time_series_metric` mapping parameter. Metric fields like `gauge` can increase or decrease over time. For more details, refer to [Metrics](/manage-data/data-store/data-streams/time-series-data-stream-tsds.md#time-series-metric).
```
kkrik-es, line 111: I think we should either mention counters too, or just stick to the cross reference for more details.

manage-data/data-store/data-streams/set-up-tsds.md
```diff
@@ -150,79 +156,105 @@ PUT _index_template/my-weather-sensor-index-template
...
 
-    You can manually roll over the write index using the [rollover API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-rollover).
+### Convert an existing data stream to a TSDS [convert-existing-data-stream-to-tsds]
 
-    ```console
-    POST metrics-weather_sensors-dev/_rollover
-    ```
+You can convert an existing regular data stream to a TSDS. Follow these steps:
 
+1. Update your existing index template to include time series settings. Also update your index lifecycle policy (if any) and component templates (if any).
```
gmarouli, line 234: The second sentence here is a vague, when I read I have no idea what kind of updates it is talking about. Maybe something like:

> Update your existing index template and/or component templates (if any) to include time series settings.

It is reasonable to bundle the index template and component templates together, because depending on the setup it might be enough to update only one component template. 

About the ILM policy, I have no idea what updates it is referring to, that's why I would suggest to remove it.

manage-data/data-store/data-streams/time-bound-tsds.md
```diff
@@ -0,0 +1,86 @@
...
+### Accepted time range for adding data [tsds-accepted-time-range]
+
+A TSDS is designed to ingest current metrics data. When the TSDS is first created, the initial backing index has the following settings:
+
+- An `index.time_series.start_time` value set to `now - index.look_back_time`
+- An `index.time_series.end_time` value set to `now + index.look_ahead_time`
+
+Only data that falls within this range is indexed.
+
+To check the accepted time range for writing to a TSDS, use the [get data stream API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-get-data-stream):
```
gmarouli, line 40: We are missing a nuance here, this API responds with the time range supported by a TSDB, but the writes are not necessarily accepted in this time range, if a backing index is marked read-only for example, they will be rejected.

Not sure how to rephrase this, potentially it could but it's not a given.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
gmarouli: +1, here I would mention that the id is calculated by es and cannot be provided, and if we want to elaborate more it should be in an implementation section. The reason is that I see it as an implementation detail and not as part of the API, if that makes sense.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
marciw: reorder and rephrased

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
marciw: This was legacy content so I went ahead and deleted it

manage-data/data-store/data-streams/set-up-tsds.md
```diff
@@ -11,30 +11,38 @@ products:
...
+```{applies_to}
+stack: ga
+serverless: unavailable
+```
+In most cases, you can use a data stream lifecycle to manage your time series data stream. If you're using [data tiers](/manage-data/lifecycle/data-tiers.md) in {{stack}}, you can use index lifecycle management (ILM).
 
+:::{dropdown} Create an ILM policy
 
-## Create an index lifecycle policy [tsds-ilm-policy]
+If you're using {{stack}}, {{ilm-init}} can help you manage a time series data stream's backing indices. {{ilm-init}} requires an index lifecycle policy.
```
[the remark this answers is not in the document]
marciw: Thank you for explaining! Moved lifecycle mentions to a new "set up lifecycle management" section (with cross-reference to the full data lifecycle topic) and removed the step.

manage-data/data-store/data-streams/time-bound-tsds.md
marciw: I think the tip that comes after this explanation tries to capture this nuance. I added a more general sentence at the beginning of the tip:

`Writes within the accepted time range might still be rejected. The following actions can affect the writable time range, either because they make a backing index read-only or remove it:`
(etc.)

manage-data/data-store/data-streams/set-up-tsds.md
```diff
@@ -86,28 +94,26 @@ PUT _ilm/policy/my-weather-sensor-lifecycle-policy
...
 
-    * Set `index.mode` setting to `time_series`.
-    * Your lifecycle policy in the `index.lifecycle.name` index setting.
-    * Optional: Other index settings, such as [`index.number_of_replicas`](elasticsearch://reference/elasticsearch/index-settings/index-modules.md#dynamic-index-number-of-replicas), for your TSDS’s backing indices.
+- **Index patterns:** One or more wildcard patterns matching the name of your TSDS, such as `weather-sernsors-*`. For best results, use the [data stream naming scheme](/reference/fleet/data-streams.md#data-streams-naming-scheme).
+- **Data stream object:** The template must include `"data_stream": {}`.
+- **Time series mode:** Set `index.mode: time_series`.
+- **Field mappings:** Define at least one `keyword` dimension field and typically one or more metric fields:
+    - To define a dimension, set `time_series_dimension` to `true`. Dimension fields like `counter` only increase over time. For more details, refer to [Dimensions](/manage-data/data-store/data-streams/time-series-data-stream-tsds.md#time-series-dimension).
```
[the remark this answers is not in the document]
marciw: Added a sub-bullet with the cross-reference:

`- To define dimensions dynamically, you can use a pass-through object. For more details, see [Defining sub-fields as time series dimensions](elasticsearch://reference/elasticsearch/mapping-reference/passthrough.md#passthrough-dimensions).`

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -11,63 +11,68 @@ products:
...
-* A TSDS uses internal [index sorting](elasticsearch://reference/elasticsearch/index-settings/sorting.md) to order shard segments by `_tsid` and `@timestamp`.
-* TSDS documents only support auto-generated document `_id` values. For TSDS documents, the document `_id` is a hash of the document’s dimensions and `@timestamp`. A TSDS doesn’t support custom document `_id` values.
-* A TSDS uses [synthetic `_source`](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#synthetic-source), and as a result is subject to some [restrictions](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#synthetic-source-restrictions) and [modifications](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#synthetic-source-modifications) applied to the `_source` field.
+You can use the {{esql}} [`TS` command](elasticsearch://reference/query-languages/esql/commands/ts.md) to query time series data streams. The `TS` command is optimized for time series data. It also enables the use of aggregation functions that efficiently process metrics per time series, before aggregating results.
```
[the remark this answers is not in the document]
marciw: yes, thanks!

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -78,146 +61,68 @@ You mark a field as a dimension using the boolean `time_series_dimension` mappin
...
 
-### Metrics [time-series-metric]
+Metrics are numeric measurements that change over time. Documents in a TSDS typically contain one or more metric fields. 
 
-Metrics are fields that contain numeric measurements, as well as aggregations and/or downsampling values based off of those measurements. While not required, documents in a TSDS typically contain one or more metric fields.
-
-Metrics differ from dimensions in that while dimensions generally remain constant, metrics are expected to change over time, even if rarely or slowly.
+:::{tip}
+Metrics are expected to change (even if rarely or slowly), while dimensions generally remain constant.
+:::
```
kkrik-es, line 74: You can probably remove this note. Dimension values may also change, e.g. as nodes join and leave a cloud deployment.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -78,146 +61,68 @@ You mark a field as a dimension using the boolean `time_series_dimension` mappin
...
-
-When you create the matching index template for a TSDS, you must specify one or more dimensions in the `index.routing_path` setting. Each document in a TSDS must contain one or more dimensions that match the `index.routing_path` setting.
+:   A metric that represents a single numeric that can arbitrarily increase or decrease. For example, a temperature or available disk space. 
 
-The `index.routing_path` setting accepts wildcard patterns (for example `dim.*`) and can dynamically match new fields. However, {{es}} will reject any mapping updates that add scripted, runtime, or non-dimension fields that match the `index.routing_path` value.
+#### `_tsid` metadata field [tsid]
 
-[Pass-through](elasticsearch://reference/elasticsearch/mapping-reference/passthrough.md#passthrough-dimensions) fields may be configured as dimension containers. In this case, their sub-fields get included to the routing path automatically.
+The `_tsid` is an automatically generated object containing the document’s dimensions. It's intended for internal {{es}} use, so in most cases you won't need to work with it. The format of the `_tsid` field is subject to change.
```
kkrik-es, line 91:
```suggestion
The `_tsid` is an automatically generated object derived from the document’s dimensions. It's intended for internal {{es}} use, so in most cases you won't need to work with it. The format of the `_tsid` field is subject to change. 
```

kkrik-es: "containing" is not accurate (it used to contain parts of the dimension values but not any more).. it's calculated using all dimension values per doc.

manage-data/data-store/data-streams/query-downsampled-data.md
```diff
@@ -2,7 +2,7 @@
 applies_to:
   stack: ga
   serverless: ga
-navigation_title: "Query downsampled data"
+navigation_title: "Querying"
 products:
   - id: elasticsearch
 ---
```
kkrik-es, line 8: We can probably include a reference to the `TS` command here as tech preview.

kkrik-es (approved):
Amazing, thanks!

manage-data/data-store/data-streams/set-up-tsds.md
```diff
@@ -11,103 +11,38 @@ products:
...
+        - To define dimensions dynamically, you can use a pass-through object. For details, refer to [Defining sub-fields as time series dimensions](elasticsearch://reference/elasticsearch/mapping-reference/passthrough.md#passthrough-dimensions).
+    - To define a metric, use the `time_series_metric` mapping parameter. For more details, refer to [Metrics](/manage-data/data-store/data-streams/time-series-data-stream-tsds.md#time-series-metric).
+    - (Optional) Define a `date` or `date_nanos` mapping for the `@timestamp` field. If you don't specify a mapping, {{es}} maps `@timestamp` as a `date` field with default options.
+    * (Optional) Other index settings, such as [`index.number_of_replicas`](elasticsearch://reference/elasticsearch/index-settings/index-modules.md#dynamic-index-number-of-replicas), for the data stream's backing indices.
+- A priority higher than `200`, to avoid [collisions](/manage-data/data-store/templates.md#avoid-index-pattern-collisions) with built-in templates.
```
gmarouli, line 43: I think we need to add here a lifecycle management section. I would recommend adding the following in the `template`:
```
"lifecycle": {
  "enabled": true
}
```
The main reason we need this is rollover, if a user doesn't add this they are going to end up with a gigantic index. Everything else is optional, we can suggest that users can read further on lifecycle management if they want more features like retention or tiers.

manage-data/data-store/data-streams/reindex-tsds.md
```diff
@@ -9,208 +9,111 @@ products:
   - id: elasticsearch
 ---
 
-# Reindex a TSDS [tsds-reindex]
+# Reindex a time series data stream [tsds-reindex]
```
gmarouli, line 12: @kkrik-es if I understand correctly, this reindexing manual is suggesting to reindex the data of one data stream into a single backing index of another data stream. Right?

If this is true, then I think we need to add a disclaimer before a user gets unpleasantly surprised. 

We could also mention the reindex data stream API that was added for upgrades, I will check if it works.

manage-data/data-store/data-streams/time-series-data-stream-tsds.md
```diff
@@ -78,146 +61,68 @@ You mark a field as a dimension using the boolean `time_series_dimension` mappin
...
 
-{{es}} uses [compression algorithms](elasticsearch://reference/elasticsearch/index-settings/index-modules.md#index-codec) to compress repeated values. This compression works best when repeated values are stored near each other — in the same index, on the same shard, and side-by-side in the same shard segment.
+## Query time series data
+```{applies_to}
+stack: preview
+serverless: preview
+```
 
-Most time series data contains repeated values. Dimensions are repeated across documents in the same time series. The metric values of a time series may also change slowly over time.
+You can use the {{esql}} [`TS` command](elasticsearch://reference/query-languages/esql/commands/ts.md) (in technical preview) to query time series data streams. The `TS` command is optimized for time series data. It also enables the use of aggregation functions that efficiently process metrics per time series, before aggregating results.
```
felixbarny, line 113: Maybe remove the (in technical preview). The TS docs also contain that and it's another thing we need to remember to keep in sync when TS goes beta or GA.

manage-data/data-store/data-streams/set-up-tsds.md
felixbarny, line 1: Either in this page, or in the quickstart, we could document how to send data via the OTLP endpoint. WDYT @kkrik-es? I can follow up with a PR for that. This is probably more relevant than the API docs for OLTP

manage-data/data-store/data-streams/reindex-tsds.md
```diff
@@ -9,208 +9,111 @@ products:
...
   "template": {
     "settings": {
-      "index": {
-        "number_of_replicas": 2,
-        "number_of_shards": 2,
-        "mode": "time_series",
-        "routing_path": [ "metricset" ]
-      }
+      "index.mode": "time_series",
+      "index.routing_path": ["dimension_field"]
```
kkrik-es, line 47: This setting gets auto-populated, let's remove it.

yannis-roussos (approved):

PR #3234, opened 2025-09-30
[AutoOps] Update incoming HTTP port for OTLP

pickypg:
This updates the OTLP port to indicate that it is now exclusively `443` instead of the standard `4318` (which causes issues for some users).


arnonst:
@wajihaparvez We made a quick fix here as we release a fix to the OTel port

PR #3245, opened 2025-10-01
Clarifications added to remote clusters, self-managed, api key based auth

eedugon:
This is a preliminary / quick solution to address the issues and noise caused by the lack of details of the doc to configure remote clusters from a self-managed cluster towards ECK.

This PR improves:

- The note that was clearly vague, telling users that enabling the server interface and adding  `xpack.security.remote_cluster_client.ssl.certificate_authorities` is NOT needed when remote is ECH.
- Highlights at the beginning that the remote cluster server interface port is `9443`.
- Tells the user to NOT try `sniff` mode if the remote is on ECH, ECE or ECK, as that will never be an option.

In a next iteration we will refine the content to cover other remote possibilities, such as ECE or ECK, completely, as we have in other docs.

For example for self-manged --> ECE we will require the user to add the ECE proxy CA into the self-managed list of trusted certs, but that's not covered yet.

### DIFF
--- a/deploy-manage/remote-clusters/remote-clusters-api-key.md
+++ b/deploy-manage/remote-clusters/remote-clusters-api-key.md
@@ -16,7 +16,7 @@ All cross-cluster requests from the local cluster are bound by the API key’s p
 
 On the local cluster side, not every local user needs to access every piece of data allowed by the API key. An administrator of the local cluster can further configure additional permission constraints on local users so each user only gets access to the necessary remote data. Note it is only possible to further reduce the permissions allowed by the API key for individual local users. It is impossible to increase the permissions to go beyond what is allowed by the API key.
 
-In this model, cross-cluster operations use [a dedicated server port](elasticsearch://reference/elasticsearch/configuration-reference/networking-settings.md#remote_cluster.port) (remote cluster interface) for communication between clusters. A remote cluster must enable this port for local clusters to connect. Configure Transport Layer Security (TLS) for this port to maximize security (as explained in [Establish trust with a remote cluster](#remote-clusters-security-api-key)).
+In this model, cross-cluster operations use [a dedicated server port](elasticsearch://reference/elasticsearch/configuration-reference/networking-settings.md#remote_cluster.port) (remote cluster interface) for communication between clusters, which defaults to port `9443`. A remote cluster must enable this port for local clusters to connect. Configure Transport Layer Security (TLS) for this port to maximize security (as explained in [Establish trust with a remote cluster](#remote-clusters-security-api-key)).
 
 The local cluster must trust the remote cluster on the remote cluster interface. This means that the local cluster trusts the remote cluster’s certificate authority (CA) that signs the server certificate used by the remote cluster interface. When establishing a connection, all nodes from the local cluster that participate in cross-cluster communication verify certificates from nodes on the other side, based on the TLS trust configuration.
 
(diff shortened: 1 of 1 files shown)


deploy-manage/remote-clusters/remote-clusters-api-key.md
```diff
@@ -39,7 +39,11 @@ If you run into any issues, refer to [Troubleshooting](/troubleshoot/elasticsear
 ## Establish trust with a remote cluster [remote-clusters-security-api-key]
 
 ::::{note}
-If a remote cluster is part of an {{ech}} deployment, it has a valid certificate by default. You can therefore skip steps related to certificates in these instructions.
+If a remote cluster is part of an {{ech}} (ECH) deployment, the remote cluster server is enabled by default and it uses a publicly trusted certificate provided by the platform proxies. You can therefore skip the following steps in these instructions:
```
yetanothertw, line 42:
```suggestion
If a remote cluster is part of an {{ech}} (ECH) deployment, the remote cluster server is enabled by default and it uses a publicly trusted certificate provided by the platform proxies. Therefore, you can skip the following steps in these instructions:
```

deploy-manage/remote-clusters/remote-clusters-api-key.md
```diff
@@ -39,7 +39,11 @@ If you run into any issues, refer to [Troubleshooting](/troubleshoot/elasticsear
...
+
+**On the remote (ECH) cluster:** Skip steps 1-4 (enabling the service, generating certificates, configuring SSL settings, and restarting the cluster), and go straight to step 5 (create API key step).
+
+**On the local (self-managed) cluster:** Do not add the `xpack.security.remote_cluster_client.ssl.certificate_authorities` setting to the configuration file because ECH uses publicly trusted certificates that don't require custom CA configuration.
```
yetanothertw, line 46: I don't understand enough about this, but would it make sense to repeat this specific condition again in the corresponding step (step b of On the local cluster), so that it's reiterated in context? Just in case it was missed.

deploy-manage/remote-clusters/remote-clusters-api-key.md
```diff
@@ -149,6 +153,10 @@ To add a remote cluster from Stack Management in {{kib}}:
 2. Enter a name (*cluster alias*) for the remote cluster.
 3. Specify the {{es}} endpoint URL, or the IP address or host name of the remote cluster followed by the remote cluster port (defaults to `9443`). For example, `cluster.es.eastus2.staging.azure.foundit.no:9443` or `192.168.1.1:9443`.
 
+::::{note}
+If the remote cluster is part of an {{ech}}, {{ece}}, or {{eck}} deployment, ensure you configure the connection to use `proxy`, as the default `sniff` mode won't work in this type of environments. Refer to the [connection modes](/deploy-manage/remote-clusters/remote-clusters-self-managed.md#sniff-proxy-modes) description for more information.
```
yetanothertw, line 157:
```suggestion
If the remote cluster is part of an {{ech}}, {{ece}}, or {{eck}} deployment, configure the connection to use `proxy`. The default `sniff` mode doesn't work in these environments. Refer to the [connection modes](/deploy-manage/remote-clusters/remote-clusters-self-managed.md#sniff-proxy-modes) description for more information.
```

yetanothertw (approved):
Approving to unblock you, @eedugon!

I've left a few minor suggestions, but am afraid don't know enough about this subject to contribute more meaningfully. Many thanks!

deploy-manage/remote-clusters/remote-clusters-api-key.md
```diff
@@ -120,6 +124,10 @@ If a remote cluster is part of an {{ech}} deployment, it has a valid certificate
         xpack.security.remote_cluster_client.ssl.certificate_authorities: [ "remote-cluster-ca.crt" ]
         ```
 
+        ::::{tip}
+        If the remote cluster uses a publicly trusted certificate, you don't need to add the CA certificate to the `certificate_authorities` setting. This example assumes you're using the private certificates [created earlier on the remote cluster](#remote-clusters-security-api-key-remote-action), which require the CA to be added.
+        ::::
+
```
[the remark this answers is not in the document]
yetanothertw: It sounds like you don't need to add the CA certificate to the setting, but still include the setting in the file. I think we can say they can exclude this configuration setting instead?

```suggestion
        ::::{tip}
        If the remote cluster uses a publicly trusted certificate, don't include the `certificate_authorities` setting. This example assumes you're using the private certificates [created earlier on the remote cluster](#remote-clusters-security-api-key-remote-action), which require the CA to be added.
        ::::

```

shainaraskas (approved):
great quick wins. nice

PR #3269, opened 2025-10-02
[Security] Entity store historical views

natasha-moore-elastic:


solutions/security/advanced-entity-analytics/view-analyze-risk-score-data.md
```diff
@@ -13,14 +13,7 @@ products:
...
-* [Entity Analytics overview](#entity-analytics-overview)
-* [Alerts page](#alerts-page)
-* [Alert details flyout](#alert-details-flyout)
-* [Hosts and Users pages](#hosts-users-pages)
-* [Host and user details pages](#host-user-details-pages)
-* [Entity details flyouts](#entity-details-flyouts)
+The {{security-app}} provides several ways to monitor the change in the risk posture of entities in your environment.
 
 ::::{tip}
 We recommend that you prioritize [alert triaging](#alert-triaging) to identify anomalies or abnormal behavior patterns.
```
MikePaquette, line 19: This section is not consistent with our users intended workflow.
While it's possible to go "backwards" to go back to alert triage, the benefit of risk-based analysis is that the app has already done that for you by creating the risk score, which is distilled information from the alert history of the selected entity.  The normal next step in our workflow from here is to go to investigate the risky user in timeline. Our docs should emphasize this path as primary, not alert triage. 
image

MikePaquette (dismissed):

PR #3289, opened 2025-10-02
Update price change for Nov 1 update

demers-eric:


kilfoyle (approved):
LGTM! 🦕 

The new prices all look right.

Issue #2067, opened 2025-07-08
[Website]: elastic.co/docs, Fleet-managed APM


AnastasiiaMonika:
### Type of issue

Inaccurate

### What documentation page is affected

https://www.elastic.co/docs/solutions/observability/apm/get-started

### What happened?

Hi 

according to "help me decide" paragraf, I have to install centrally hoster APM server in I want to run RUM (see comment under "Where will you run APM server-- Note that central in required for RUM") 

BUT 

According to Configure RUM page 
https://www.elastic.co/docs/solutions/observability/apm/configure-real-user-monitoring-rum 

Fleed-managed APM server is supported now 

Where is the truth? :)

theletterf:
@AnastasiiaMonika You're indeed right: RUM can also be used with Fleet-managed APM Server. I'm going to correct the diagram to remove the note about RUM. Thanks again for your patience — and apologies for the delay.

PR #3308, opened 2025-10-03
Replace diagram

theletterf:
Fixes https://github.com/elastic/docs-content/issues/2067

Note that this uses an updated version of the diagram that adds ECH as the first node.


vishaangelova (approved):
LGTM

PR #3311, opened 2025-10-03
edot-sdks/python: document OTEL_LOG_LEVEL in troubleshoot

xrmx:


theletterf (approved):
Thanks! Fixed the `applies_to` section statements, as a lifecycle is always required before a version is stated: https://docs-v3-preview.elastic.dev/elastic/docs-builder/tree/main/syntax/applies#key-value-reference

PR #3312, opened 2025-10-03
ECH and ECE ccs template migration updated

eedugon:
I've updated the ECE and ECH docs to migrate from the legacy / removed CCS template to a standard template, as the dedicated template is no longer needed anymore.

@shainaraskas : For the ECE doc we have to take a decision, as the doc is NOT applicable to ECE 4.x at all.
I've added a note explaining that the doc is only valid for ECE 3.x installations running deployments in versions earlier than 8.0 with the mentioned template.

The reason of not being applicable is that before upgrading to ECE 4.x all deployments must be on 8.x, and to upgrade to 8.x this template must be not used by any deployment.

Our options are:
- Accept my changes and leave the doc in our docs V3.
- Remove the doc entirely and rely on the older version of the doc (ECE 3.x).

The doc for ECH is valid because in ECH we are still supporting 7.17 deployments and "maybe" this removed template, not sure about that.

PS - I have removed the `deprecated` statement as if the template is `removed` from the system that implies a previous deprecation.


deploy-manage/remote-clusters/ece-migrate-ccs.md
```diff
@@ -9,12 +9,15 @@ products:
...
+The legacy cross-cluster search deployment template was removed in {{ece}} 3.0. You no longer need a dedicated template to search across deployments. Instead, you can now use any template to [configure remote clusters](ece-enable-ccs.md) and search across them. Existing deployments created using this template are not affected, but they are required to migrate to another template before upgrading to {{stack}} 8.x.
 
-In order to migrate your existing CCS deployment using the CCS Deployment template to the new mechanism which supports CCR and cross-environment remote clusters you will need to migrate your data a new deployment [following these steps](#ece-migrate-ccs-deployment-using-snapshot).
+::::{important}
+This guide only applies to {{ece}} 3.x installations and to deployments on {{stack}} versions earlier than 8.0 that were created using the legacy cross-cluster search template.
+::::
 
+In order to migrate your existing CCS deployment using the CCS deployment template to the new mechanism which supports CCR and cross-environment remote clusters you will need to migrate your data a new deployment [following these steps](#ece-migrate-ccs-deployment-using-snapshot).
```
shainaraskas, line 20: this anchor linking down to the next line is not a great pattern. just say "follow these instructions to ..." (implying content is to follow)

shainaraskas (approved):
approved so you can make a choice to keep or get rid of this doc, considering the 7.17 sunset

note: if you do delete it, we'll need to ask the web team to redirect this url to 3.8 (instead of these docs), and remove the listing from https://github.com/elastic/docs/blob/master/resources/asciidoctor/lib/chunker/v3-mapping.json

eedugon:
@kunisen , I'd like to get your thoughts here (for the ECE doc):

Do you think it's beneficial to keep this doc in the new system? https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/3312/deploy-manage/remote-clusters/ece-migrate-ccs

Or would it be enough to leave the 3.8 doc as the reference? https://www.elastic.co/guide/en/cloud-enterprise/3.8/ece-migrate-ccs.html

The reason is that the doc is not applicable to ECE 4.x, as before upgrading to ECE 4.x all deployments should have been upgraded to 8.x, and that would imply to get rid of the ccs deployment template.

I don't have a clear opinion here, I'm ok removing the doc because it's probably not needed, but I'm ok leaving it also.

PR #3316, opened 2025-10-03
Fix "component template" link in ILM docs

kilfoyle:
Fixes a link to point to regular docs instead of the API.


yetanothertw (approved):
Great idea, thank you! 🚀

Issue #1564, opened 2025-05-30
Data lifecycle docs: "Skip rollover" improvements


kilfoyle:
Feedback on [Skip rollover](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/skip-rollover) (23:50):

 - [x] Explain why you'd want to do this.
 - [x] You can apply ILM policy and skip rollover (customers do this).Typically these are Search customers who manually create an index each month on the same day. They don't have continuous ingest. To disable: Go to ILM policy > Advanced settings > disable rollover. Index won't rollover, but you can set it to automatically move this to move to Warm phase.

kilfoyle:
Good points @yetanothertw!

For the use cases perhaps we don't need a lot of detail. We can say something like:

```
You might want to disable rollover if:

 - You're not continuously ingesting data, and instead simply want to move your indices to a different ILM phase at a certain interview (for example, on the first day of each month).
 - Something else...
```

The existing content sure does seem advanced. We can try to make it more basic, and I think also add the UI steps.

PR #3320, opened 2025-10-03
ILM docs: Update 'Skip rollover'

kilfoyle:
This updates the [Skip rollover](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/skip-rollover) page to include the UI steps and to provide some guidance about why one would do this.

@yetanothertw I tried to do this in a way that preserves the original (advanced) content. Please let me know what you think.

Closes: https://github.com/elastic/docs-content/issues/1564


manage-data/lifecycle/index-lifecycle-management/skip-rollover.md
```diff
@@ -9,17 +9,40 @@ products:
...
+1. Update the **Enable rollover** option to enable or disable automatic rollover.
+
+    ![Create policy page](/manage-data/images/elasticsearch-reference-create-policy.png "")
+
+
+## Skip rollover manually
+
+When the [`index.lifecycle.indexing_complete`](elasticsearch://reference/elasticsearch/configuration-reference/index-lifecycle-management-settings.md#_index_level_settings_2) setting is `true`, {{ilm-init}} won’t perform the rollover action on an index, even if it otherwise meets the rollover criteria. This setting is updated to `true` automatically by {{ilm-init}} when the rollover action completes successfully.
+
+You can update this setting manually to skip rollover if, for instance, you need to make an exception to your normal lifecycle policy and update the alias to force a rollover, but you want {{ilm-init}} to continue to manage the index. If you use the rollover API rather than an {{ilm-init}} policy to rollover indices, it is not necessary to configure this setting manually.
```
yetanothertw, line 38:
```suggestion
You can update this setting manually to skip rollover if, for instance, you need to make an exception to your normal lifecycle policy and update the alias to force a rollover, but you want {{ilm-init}} to continue to manage the index. If you use the rollover API rather than an {{ilm-init}} policy to roll over indices, it is not necessary to configure this setting manually.
```
Is this because you can't automate rollover via the API?

yetanothertw (approved):
This is a great update to this section. I think I understand better when or why one would consider skipping rollover.

I've left an optional suggestion at your discretion. 
🚢

manage-data/lifecycle/index-lifecycle-management/skip-rollover.md
kilfoyle: I'm not sure. I just moved this content from its original location. I gather that this is someone running the rollover API on a one-time basis, in which case it automatically updates the setting to "true".

PR #3327, opened 2025-10-04
Dedicated frozen

stefnestor:
👋 howdy, 

This important snippet was moved down only [under self-managed](https://www.elastic.co/docs/manage-data/lifecycle/data-tiers#configure-data-tiers-on-premise) but also applies to admins in ECK+ECE. (Only ECH manages it for users but it also still does technically apply to them.)

image


yetanothertw (approved):
Thank you for spotting and fixing this @stefnestor!

My only suggestion has to do with keeping the message consistent with the note. I'll leave that at your discretion.

PR #3361, opened 2025-10-07
ipv6 support for remote addresses added

eedugon:


deploy-manage/remote-clusters/_snippets/rcs-kibana-api-snippet-self.md
```diff
@@ -0,0 +1,29 @@
...
+    * **Remote cluster name**: This *cluster alias* is a unique identifier that represents the connection to the remote cluster and is used to distinguish local and remote indices.
+
+      When using API key authentication, this alias must match the **Remote cluster name** you configured when adding the API key in the Cloud UI.
+    * **Remote address**: Enter the endpoint of the remote self-managed cluster,  including the hostname, FQDN, or IP address, and the port. Both IPv4 and IPv6 ({applies_to}`stack: ga 9.2`) addresses are supported.
+
+      Make sure you use the correct port for your authentication method:
+      * **API keys**: Use the port configured in the remote cluster interface of the remote cluster (defaults to `9443`).  
+      * **TLS Certificates**: Use the {{es}} transport port (defaults to `9300`).
+
+      When using an IPv6 address ({applies_to}`stack: ga 9.2`), enclose it in square brackets followed by the port number. For example: `[2001:db8::1]:9443`.
```
shainaraskas, line 20: we generally don't want to use these tags inline like that. consider using a note instead. we should also direct people to use the api if they want to set a v6 address in 9.1 and earlier

deploy-manage/remote-clusters/_snippets/rcs-elasticsearch-api-snippet-self.md
```diff
@@ -0,0 +1,40 @@
+<!--
+This snippet is in use in the following locations:
+- ece-remote-cluster-self-managed.md
+- ec-remote-cluster-self-managed.md
+-->
+To configure a self-managed cluster as a remote cluster, use the [cluster update settings API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-put-settings). Configure the following fields:
+
+* `mode`: `proxy`
+* `proxy_address`: Enter the endpoint of the remote self-managed cluster, including the hostname, FQDN, or IP address, and the port. Both IPv4 and IPv6 ({applies_to}`stack: ga 9.2`) addresses are supported.
```
shainaraskas, line 9: this is not specific to 9.2 here is it?

shainaraskas (approved):
couple of comments for you. I think the applies tag needs to be removed from the api instructions

deploy-manage/remote-clusters/_snippets/rcs-elasticsearch-api-snippet-self.md
eedugon: badge removed

deploy-manage/remote-clusters/_snippets/rcs-kibana-api-snippet-self.md
```diff
@@ -0,0 +1,29 @@
...
+    * **Remote cluster name**: This *cluster alias* is a unique identifier that represents the connection to the remote cluster and is used to distinguish local and remote indices.
+
+      When using API key authentication, this alias must match the **Remote cluster name** you configured when adding the API key in the Cloud UI.
+    * **Remote address**: Enter the endpoint of the remote self-managed cluster,  including the hostname, FQDN, or IP address, and the port. Starting with {{kib}} 9.2, this field also supports IPv6 addresses.
+
+      Make sure you use the correct port for your authentication method:
+      * **API keys**: Use the port configured in the remote cluster interface of the remote cluster (defaults to `9443`).  
+      * **TLS Certificates**: Use the {{es}} transport port (defaults to `9300`).
+
+      When using an IPv6 address, enclose it in square brackets followed by the port number. For example: `[2001:db8::1]:9443`.
```
florent-leborgne, line 20:
```suggestion
    * **Remote address**: Enter the endpoint of the remote self-managed cluster,  including the hostname, FQDN, or IP address, and the port.

      Make sure you use the correct port for your authentication method:
      * **API keys**: Use the port configured in the remote cluster interface of the remote cluster (defaults to `9443`).  
      * **TLS Certificates**: Use the {{es}} transport port (defaults to `9300`).

      Starting with {{kib}} 9.2, this field also supports IPv6 addresses. When using an IPv6 address, enclose it in square brackets followed by the port number. For example: `[2001:db8::1]:9443`.
```
Since this content is no longer in notes/tips, I suggest we group the IPv6 bits together, after the most important part to read (like ports information)

deploy-manage/remote-clusters/_snippets/rcs-kibana-api-snippet-self.md
```diff
@@ -0,0 +1,29 @@
...
+
+    * **Configure advanced options** (optional): Expand this section if you need to customize additional settings.
+      * **TLS server name**: Specify a value if the certificate presented by the remote cluster is signed for a different name than the remote address.
+      * **Socket connections**: Define the number of connections to open with the remote cluster.
+
+    For a full list of available client connection settings, refer to [remote cluster settings reference](elasticsearch://reference/elasticsearch/configuration-reference/remote-clusters.md#remote-cluster-proxy-settings).
```
florent-leborgne, line 26: Is this relevant for the UI configuration exposed in the Kibana UI, or are we saying that there are more options if users do it through the API? If the latter, we can make it a little bit clearer as it becomes an alternative more than complimentary information

deploy-manage/remote-clusters/remote-clusters-api-key.md
```diff
@@ -148,21 +148,26 @@ If a remote cluster is part of an {{ech}} (ECH) deployment, the remote cluster s
...
+### Using {{kib}}
+
 To add a remote cluster from Stack Management in {{kib}}:
 
 1. Select **Remote Clusters** from the side navigation.
 2. Enter a name (*cluster alias*) for the remote cluster.
-3. Specify the {{es}} endpoint URL, or the IP address or host name of the remote cluster followed by the remote cluster port (defaults to `9443`). For example, `cluster.es.eastus2.staging.azure.foundit.no:9443` or `192.168.1.1:9443`.
+3. Specify the {{es}} endpoint URL, the IP address, or host name of the remote cluster, followed by the remote cluster port (defaults to `9443`). For example, `cluster.es.eastus2.staging.azure.foundit.no:9443` or `192.168.1.1:9443`.
+
+    Both IPv4 and IPv6 ({applies_to}`stack: ga 9.2`) addresses are supported.
```
florent-leborgne, line 161:
```suggestion
    Starting with {{kib}} 9.2, you can also specify IPv6 addresses.
```
To make it closer to your other edit where you no longer use a badge

florent-leborgne (approved):
I think this works. Few minor non blocking suggestions.
Thanks Edu for handling the Kibana part :)

Issue #3381, opened 2025-10-07
[Internal]: Add anonymization limitation


neptunian:
### Description

Changes to anonymization were introduced in https://github.com/elastic/kibana/pull/232319.  This includes not stringifying an entire tool response and walking the JSON tree and anonymizing leaf nodes for better accuracy and less errors.  It was decided to not "look" at non string types (more info here https://github.com/elastic/kibana/issues/237906).  We should clarify that. 

### Resources

Implemented in https://github.com/elastic/kibana/pull/232319
May change this behavior in the future https://github.com/elastic/kibana/issues/237906

### Which documentation set does this change impact?

Elastic On-Prem and Cloud (all)

### Feature differences

None

### What release is this request related to?

9.2

### Serverless release

Already there

### Collaboration model

The engineering team

### Point of contact.

**Main contact:** @neptunian 

**Stakeholders:** @neptunian @sorenlouv @emma-raffenne

PR #3382, opened 2025-10-07
Anonymization limitation update update

neptunian:
Resolves https://github.com/elastic/docs-content/issues/3381


mdbirnstiehl (commented):
Added the `applies_to` tag, but otherwise LGTM!

PR #3387, opened 2025-10-07
[Cases][Security][9.2 & Serverless]: Auto-extract observables when adding alerts to a Security case

nastasha-solomon:
Fixes https://github.com/elastic/docs-content-internal/issues/362 by adding information about the auto-extract observables feature for Security cases. 

Preview


florent-leborgne (approved):
LGTM

PR #3397, opened 2025-10-08
Update ingest-lag.md because of `event.put` to event.*P*ut

philippkahr:
This is a typo in the JavaScript, fixing it for better copy pasting. Thanks @morashwan for pointing it out


vishaangelova (approved):
LGTM

PR #3486, opened 2025-10-15
Update private-connectivity-aws.md

5hanker:
This looks like a typo. I see Azure instead of AWS in the AWS Privatelink setup doc.

https://www.elastic.co/docs/deploy-manage/security/private-connectivity-aws#ec-access-the-deployment-over-private-link


kilfoyle (approved):
LGTM! 🏎️ 
Nice catch. Thanks @5hanker!

PR #3490, opened 2025-10-15
switch to variables for stack docker images included in ECE

shainaraskas:
part of https://github.com/elastic/docs-content-internal/issues/27

Switching to two variables so this information is easier to update. will be documented in the ECE release coordinator tasks (see issue)


kilfoyle (approved):
LGTM! ⛵

PR #3398, opened 2025-10-08
WIP: upgrade from 7.17 guide

eedugon:
WIP

PR #3491, opened 2025-10-15
reindex to upgrade method removed

eedugon:
**Original scope of this PR**

This PR refines and scopes a bit the method that we have shared for ages as an alternative to an actual major upgrade: reindex to upgrade.

The method consists of creating a new cluster in the desired major version, and just reindex the data from the old cluster running a much older version.

I believe (and I could be wrong) that this method won't work today for system indices, Kibana data and probably a lot of other features of our stack.

I have proposed to consider this method only for use cases that do not rely heavily on stack features, or when the user just wants to migrate their own application data to the new version.

This PR relates to https://github.com/elastic/docs-content/pull/3398

@ppf2 , @shainaraskas , I'd like to get your thoughts on that. I discovered this when I was preparing the 7.17 -> 9.x upgrade guide, because we have always suggested in our docs the "reindex to upgrade" as an alternative to a real major upgrade, and when thinking about it I believe it's clearly not that simple and it's not going to work for users that rely heavily on our new Stack Solutions. Those use cases should follow a normal upgrade path and not consider a reindex to upgrade as an alternative approach if they want to keep all their kibana and features data.

cc: @georgewallace , it would also be good if a PM or someone from dev team validates this.

**PR scope UPDATE**

Based on the feedback below we have decided to remove the "reindex to upgrade" method and just refer to it as data migration (which we have already documented).
We have removed it from the "prepare to upgrade" document, as it's clearly not the place for it, and we have mentioned the possibility of migrating data to a new cluster instead of upgrading in the upgrade landing doc, also stating that when aiming to upgrade, data migration doesn't cover Kibana data, stack features data and system indices.


shainaraskas (approved):

georgewallace:
@eedugon are we still waiting on a review from the ES infra team on this one?

eedugon:
> @eedugon are we still waiting on a review from the ES infra team on this one?

ideally yes

mark-vieira:
I agree, this doesn't really feel like an "upgrade" method and these recommendation likely precedes system indicies. I see a couple options here. First, we could simply remove this entirely as it's not really a proper upgrade path. Second, we could instruct folks to use snapshots to move over the relevant feature states. @rjernst, curious on your thoughts here.

rjernst:
> we could instruct folks to use snapshots to move over the relevant feature states

Feature states still need to go through migration, system indices need to be reindexed internally. Realistically reindexing is no longer a viable "upgrade" method.

mark-vieira:
> Feature states still need to go through migration, system indices need to be reindexed internally. Realistically reindexing is no longer a viable "upgrade" method.

I concur. I'm apt to simply remove this entirely or rebrand it as a "data migration" method vs an "upgrade" but I suspect we already capture that use case in the reindex docs.

eedugon:
Thanks @mark-vieira and @rjernst , I completely agree with your comments, and I was also suspecting this was kind of deprecated.

I'm going to update this PR (or cancel + create a new one) to remove this as an official upgrade alternative method and ensure we are covering this in our data migration doc.

deploy-manage/upgrade.md
```diff
@@ -110,3 +110,5 @@ Upgrading from {{stack}} version 7.17 to the latest {{version.stack}} release in
 - From 8.19 to {{version.stack}}
 
 For detailed guidance on planning and executing these upgrades, refer to the [Upgrade from 7.17 guide](/deploy-manage/upgrade/deployment-or-cluster/upgrade-717.md).
+
+If you prefer to migrate your data to a new cluster instead of performing an in-place upgrade, refer to [Migrate your {{es}} data](/manage-data/migrate.md). In upgrade scenarios, this approach focuses on user and application data, and does not include {{kib}} data, system indices, or other {{stack}} feature data.
```
yetanothertw: We can make this a tad clearer:

```suggestion
If you prefer to migrate your data to a new cluster instead of performing an in-place upgrade, refer to [Migrate your {{es}} data](/manage-data/migrate.md). 
When migrating data to a new cluster, the primary focus is on moving user and application data, which differs from upgrade scenarios that include {{kib}} data, system indices, and other {{stack}} feature data.
```
OR 

```suggestion
If you prefer to migrate your data to a new cluster instead of performing an in-place upgrade, refer to [Migrate your {{es}} data](/manage-data/migrate.md).
Migrating data to a new cluster prioritizes the transfer of user and application data, leaving behind {{kib}} data, system indices, and other {{stack}} feature data that are usually part of the upgrade process.
```

mark-vieira (approved):
I think this is a much more honest description of the options. Looks good to me.

shainaraskas (approved):
this looks like a reasonable approach to me. thanks for circling back 🙏

PR #3496, opened 2025-10-16
Update instructions for connecting AutoOps to self-managed cluster

insukcho:
We should create the Elasticsearch API key or Role from the **source side self-managed cluster**. We don't need ECH deployment for setting up AutoOps for the self managed cluster and it is making a big confusion.


eedugon (approved):
Thanks for this, I hope we don't have similar errors in other autoops docs. If you know about anything else feel free to reach out directly to me.

PR #3502, opened 2025-10-16
EIS: remove limitations for GA

maxjakob:
The Elastic Inference Service (EIS) is going GA in `9.2`. Hence we remove the limitations that apply to tech preview.

We leave in the batch size limitation as it currently still exists.


maxjakob:
> Should we update this to indicate that this is becoming GA in 9.2?

Yes, thanks for showing me how to tag this!

> Are these [batch size] limitations still valid for users that are still on 9.1 and aren't upgrading to 9.2?

The limitation is in the EIS service, therefore it is not dependent on any Elasticsearch version. We will remove/change this restriction at some point in EIS at which point it will be changed for all Elasticsearch clusters of all versions.

> batch size limitation is still in effect [...] but the changes in the file are removing it

This was an oversight on my part.

Thanks @leemthompo, I will close this PR in favor of https://github.com/elastic/docs-content/pull/3014.

Issue #1567, opened 2025-05-30
Data lifecycle docs: Better guidance for Search users


kilfoyle:
Search customers create indices using the API (Search > Content > Indices > Project > Overview).

 - So how can I optimize my 5000 indices? Why don't we create a template or alias, or something for them? They have to rebuild everything.
 - They go to Index management to find the index they created. They “Apply a lifecycle policy”, and it doesn't work. Cluster is broken. Banners saying index can't roll over. Even if it did it can't move to next phase.
 - Problem is that index doesn't have the 00001 suffix.
 - Applying the policy from our docs for manually created indices won't work.
 - The only way is to skip the rollover, and move the indices manually. 
 - Aliases and rollover is not covered very well. 

(see around 43:00 or after)

PR #3509, opened 2025-10-16
Add note to 'Migrate to ILM' about existing indices

kilfoyle:
Just a tiny PR to help folks find our page about applying an ILM policy to existing indices, and also reiterating the warning about not applying a policy that includes rollover.

Closes: https://github.com/elastic/docs-content/issues/1567

manage-existing


shainaraskas (approved):
read this and was a little confused until I clicked through. if you felt like it, you might be able to clarify this further by saying when you apply the ilm policy it's matching on the specific index name or whatever the underlying reason for the disconnection is, and uplevel the "apply to template" recommendation. if you think this will all be obvious to readers at this stage feel free to ignore.

kilfoyle:
> read this and was a little confused until I clicked through. if you felt like it, you might be able to clarify this further by saying when you apply the ilm policy it's matching on the specific index name or whatever the underlying reason for the disconnection is, and uplevel the "apply to template" recommendation. if you think this will all be obvious to readers at this stage feel free to ignore.

Thanks for the suggestion. In this case I think I'll leave the warning where it is as it's a bit of an edge case, and a different scenario that just setting up an index and managing it (that is, this  is specific to folks who have a ton of indices set up and haven't named them in such a way to support rollover).

Issue #1557, opened 2025-05-30
Data lifecycle docs: Customize built-in policies tutorial


kilfoyle:
Feedback related to [Customize built-in policies](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/tutorial-customize-built-in-policies):

- [ ] Kubernetes logs are massive so we want a dedicated ILM policy for that, apart from the main one. Need to clone the existing policy, etc. The relevant tutorial is [Customize built-in policies](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/tutorial-customize-built-in-policies) so maybe we need to expand on that.

 - [ ] What is logs@custom that is mentioned in [Customize built-in policies](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/tutorial-customize-built-in-policies)?

kilfoyle:
I've opened https://github.com/elastic/docs-content/pull/3511 to help people find more detailed instructions for working with `@custom` deployment templates.

I think the "massive Kubernetes logs" scenario is covered nicely in the existing tutorial, so I've added a sentence at the top of that page just to give this as a specific example of why we'd want a custom ILM policy.

PR #3511, opened 2025-10-16
ILM docs: improve @custom component template usage/details

kilfoyle:
In the ILM docs, the [Customize built-in policies](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/tutorial-customize-built-in-policies#example-using-index-lifecycle-policy-apply-policy) tutorial mentions `@custom` component templates with a "learn more" link that points to the "Create or update component template" API docs. Not super helpful, since those API docs don't mention the `@custom` component templates.

Those templates are discussed in more detail in the Fleet docs, with lengthy examples of using them. This PR adds a link to that content and removes the API link.

I've also added a sentence at the top giving an example of creating a custom policy for K8s logs, since we're told that those can be giant.

Closes: https://github.com/elastic/docs-content/issues/1557

### DIFF
--- a/manage-data/data-store/data-streams/query-downsampled-data.md
+++ b/manage-data/data-store/data-streams/query-downsampled-data.md
@@ -29,7 +29,7 @@ For more details, refer to [Date histogram aggregation: Time zone](elasticsearch
 
 ## {{esql}} `TS` command
 ```{applies_to}
-stack: planned
+stack: preview 9.2
 serverless: preview
 ```
 
--- a/solutions/observability/get-started/opentelemetry/use-cases/llms/index.md
+++ b/solutions/observability/get-started/opentelemetry/use-cases/llms/index.md
@@ -23,7 +23,7 @@ Elastic currently supports LLM observability through the Elastic Distributions o
 
 The following LLM platforms are supported:
 
-| Technology | [EDOT Java](elastic-otel-java://reference/edot-java/supported-technologies.md#openai-client-instrumentation) | [EDOT Node.js](elastic-otel-node://reference/edot-node/supported-technologies.md#llm-instrumentations) | [EDOT Python](elastic-otel-python://reference/edot-python/supported-technologies.md#llm-instrumentations) |
+| Technology | [EDOT Java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/openai/openai-java-1.1/javaagent) | [EDOT Node.js](elastic-otel-node://reference/edot-node/supported-technologies.md#llm-instrumentations) | [EDOT Python](elastic-otel-python://reference/edot-python/supported-technologies.md#llm-instrumentations) |
 |:-----------|:----------|:-------------|:------------|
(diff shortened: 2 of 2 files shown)


manage-data/lifecycle/index-lifecycle-management/tutorial-customize-built-in-policies.md
```diff
@@ -11,6 +11,8 @@ products:
 
 {{es}} includes a set of built-in {{ilm-init}} policies that govern how managed indices transition as they age. This guide demonstrates how you can customize the lifecycle of a managed index, to adjust how the index transitions across [data tiers](/manage-data/lifecycle/data-tiers.md) and what [actions](/manage-data/lifecycle/index-lifecycle-management/index-lifecycle.md#ilm-phase-actions), such as downsampling or shrinking, are performed on the index during each lifecycle phase.
 
+Setting a custom {{ilm-init}} policy is useful when you have a specific set of indices, for example a set of Kubernetes logs which can grow to be quite large in volume, for which you don't want to use the default data retention duration and other {{ilm-init}} settings.
```
eedugon: beautiful :)

manage-data/lifecycle/index-lifecycle-management/tutorial-customize-built-in-policies.md
```diff
@@ -140,7 +144,7 @@ Copies of managed {{ilm-init}} policies are also marked as **Managed**. You can
 
 To apply your new {{ilm-init}} policy to the `logs` index template, create or edit the `logs@custom` component template.
 
-A `*@custom` component template allows you to customize the mappings and settings of managed index templates, without having to override managed index templates or component templates. This type of component template is automatically picked up by the index template. [Learn more](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-put-component-template).
+A `*@custom` component template allows you to customize the mappings and settings of managed index templates, without having to override managed index templates or component templates. This type of component template is automatically picked up by the index template.
```
eedugon: The sentence with `*@custom` component template feels a bit weird.

Apparently it implies that any random `*@custom` component template will be magically picked up by an index template. So if I create `mydata@custom` it will be picked up by a template.

I think what we want to mean here is that we can create `logs@custom`, `metrics@custom`, `traces@custom`, and others (but not `any`), and those will be `picked` because they belong to the default index templates created by the stack, right?

Maybe we can convert this list into a table and include for each ILM policy the associated stack index template that uses it?

image

And then in this section we can explain that we can create any `*@custom` component template where `*` is any of the default index templates, but not any random string....

eedugon: Sorry, maybe an easier solution just with a text suggestion :) Let me know your thoughts.

```suggestion
A `<name-of-managed-index-template>@custom` component template allows you to customize the mappings and settings of the managed index templates, without having to override them or their main component templates. This type of component template, which doesn't exist by default, is already included in all managed index templates to allow customizations.
```

Not sure if my goal is clear, if you understand it and you think there's a better way please go ahead!

PR #3543, opened 2025-10-20
[Security][9.x] Add 'search.allow_expensive_queries' to detections-requirements.md

sdesalas:
## Summary

De-facto requirement of Security Detections, (and much of Kibana) that has not been documented properly.

```
# elasticsearch.yml

search.allow_expensive_queries=false    # <=== causes errors across Kibana 🤯
```

Errors go as far back as `v8.18` (but likely much further), and not only affect Security app, but also many `Server Management` features such as Fleet, Saved Objects, Tags, etc.

For more info: https://github.com/elastic/kibana/pull/237496 -> Section **`Risks that were found acceptable`**

We're documenting it [under the correct location](https://www.elastic.co/docs/solutions/security/detect-and-alert/detections-requirements) to make sure that users know about it. There is an [internal slack thread](https://elastic.slack.com/archives/C02HA9E8221/p1760694975799469) for discussion.

Note also that this is documented as a [requirement of the 'alerting' plugin](https://www.elastic.co/docs/explore-analyze/alerts-cases/alerts/alerting-setup#alerting-prerequisites) that we're using under the hood.

image

### DIFF
--- a/solutions/security/advanced-entity-analytics/privileged-user-monitoring-setup.md
+++ b/solutions/security/advanced-entity-analytics/privileged-user-monitoring-setup.md
@@ -28,11 +28,24 @@ Privileged users typically include accounts with elevated access rights that all
 
 You can define privileged users in the following ways:
 
+* {applies_to}`stack: preview 9.2` [Add a supported integration](#privmon-integrations) with your organization’s user identities. If your environment is already ingesting data from a supported integration, the setup steps are skipped—you're taken directly to the Privileged user monitoring dashboard, where you can start [monitoring user activity](/solutions/security/advanced-entity-analytics/monitor-privileged-user-activitites.md).
 * [Select an existing index](#privmon-index) or create a new custom index with privileged user data.
 * [Bulk-upload](#privmon-upload) a list of privileged users using a CSV or TXT file. 
 * Use the Entity analytics APIs to [mark individual users as privileged]({{kib-apis}}/operation/operation-createprivmonuser) or [bulk-upload multiple privileged users]({{kib-apis}}/operation/operation-privmonbulkuploaduserscsv).
 
 To get started, find the **Privileged user monitoring** page in the navigation menu or use the [global search field](/explore-analyze/find-and-organize/find-apps-and-objects.md).
+
+### Add a supported integration [privmon-integrations]
+```yaml {applies_to}
+stack: preview 9.2
+```
+
+1. On the **Privileged user monitoring** page, select an integration. The supported integrations are:
+   * [Active Directory Entity Analytics](integration-docs://reference/entityanalytics_ad.md). Users in the following security groups will be automatically assigned as privileged:
+      * Domain Admins
+      * Enterprise Admins
+   * [Okta Entity Analytics](integration-docs://reference/entityanalytics_okta.md). Refer to [Standard administrator roles and permissions](https://help.okta.com/en-us/content/topics/security/administrators-admin-comparison.htm) for a list of Okta roles that will be automatically assigned as privileged users.
+2. Follow the steps to install the integration. 
 
 ### Select or create an index [privmon-index]
 
@@ -78,6 +91,7 @@ You can use multiple data source types, such as an index and a CSV file, at the 
 
 On this page, you can:
 
+* {applies_to}`stack: preview 9.2` Change which integrations you're using as data sources.
 * View, remove, and change indices after initially defining them.
 * Import a new supported file with a list of privileged users.
 


alaudazzi (approved):
LGTM

solutions/security/detect-and-alert/detections-requirements.md
```diff
@@ -28,11 +28,17 @@ stack:
...
-* In [`kibana.yml`](/deploy-manage/stack-settings.md), add the `xpack.encryptedSavedObjects.encryptionKey` setting with any alphanumeric value of at least 32 characters. For example:
+* In [`kibana.yml`](/deploy-manage/stack-settings.md):
+
+  Add the `xpack.encryptedSavedObjects.encryptionKey` setting with any alphanumeric value of at least 32 characters. For example:
 
     `xpack.encryptedSavedObjects.encryptionKey: 'fhjskloppd678ehkdfdlliverpoolfcr'`
 
+* In [`elasticsearch.yml`](/deploy-manage/deploy/self-managed/configure-elasticsearch.md):
+    1. We need to enable the `xpack.security.enabled` setting to `true`.
+    2. We need the `search.allow_expensive_queries` setting to be left on its default value of `true` for key detection features like [alerting rules](https://www.elastic.co/docs/explore-analyze/alerts-cases/alerts/alerting-setup#alerting-prerequisites) and rule exceptions to work. Please remove the line `search.allow_expensive_queries=false` if you find it.
```
nastasha-solomon: Couple of follow-up questions about this list item:
- What are the alerting rules being referenced here? Are they different from the detection rules that Security users work with? 
- Why are we asking users to remove the `search.allow_expensive_queries=false` line from their `elasticsearch.yml` file? Is there ever a scenario where they should leave it in?

sdesalas: Hi Natasha,

Thanks for prompt reply and suggestions. Here are some answers to your questions.

1. The difference between 'alerts' and 'alerting rules' is one of terminology between observability (where alerts = rules, so only one term is needed) and security (where 'alerting rules' are pattern-matching templates used to generate 'alerting events', so both terms are needed). Its all a bit nitpicky TBH. But including both terms means the user is more likely to understand that by alerting rules, we mean detection rules, even if we're using the alerting framework underneath the hood. 

2. The reason for asking users to remove `search.allow_expensive_queries=false` from their `elasticsearch.yml` is because this setting defaults to `true`. So the only way it would be set to `false` is because someone has manually changed it. It would be simpler (from a communication standpoint) to just ask users to set it `true` but then we'd be asking them to do something unnecessary since ES already comes prepackaged with the right value for this setting. Basically, if the value is set to `false` in the `yml` file, they need to remove it. If its not set, or set to `true`, they dont need to do anything. But then I am also trying to balance for cognitive load. A partial instruction that is easier to follow would be best than a complete explanation I think. 

Hope the above makes sense. Let me know your thoughts. Thanks! 😁

banderror: Alerting rules are simply rules of the Alerting Framework. It's a superset of detection rules.

nastasha-solomon: Gotcha, thanks both for the background. The current draft looks good, but I am going to suggest a minor formatting change.

solutions/security/detect-and-alert/detections-requirements.md
```diff
@@ -28,11 +28,16 @@ stack:
...
-* In [`kibana.yml`](/deploy-manage/stack-settings.md), add the `xpack.encryptedSavedObjects.encryptionKey` setting with any alphanumeric value of at least 32 characters. For example:
+* In [`kibana.yml`](/deploy-manage/stack-settings.md):
+
+  Add the `xpack.encryptedSavedObjects.encryptionKey` setting with any alphanumeric value of at least 32 characters. For example:
 
     `xpack.encryptedSavedObjects.encryptionKey: 'fhjskloppd678ehkdfdlliverpoolfcr'`
 
+* In [`elasticsearch.yml`](/deploy-manage/deploy/self-managed/configure-elasticsearch.md):
+    * Set the `xpack.security.enabled` setting to `true`. Refer to [General security settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#general-security-settings) for more information.
+    * Remove the line `search.allow_expensive_queries=false` if you find it. The `search.allow_expensive_queries` setting must be left on its default value of `true` for key detection features like [alerting rules](https://www.elastic.co/docs/explore-analyze/alerts-cases/alerts/alerting-setup#alerting-prerequisites) and rule exceptions to work.
```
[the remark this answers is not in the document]
sdesalas: Thanks @nastasha-solomon, good catch on the URL too. I'll make sure to look out for this in the future.

I've committed the changes but I'm thinking of taking out a few spare words to make it easier to read.

PR #3549, opened 2025-10-20
Change semantic query example to match query

charlotte-hoblik:
This PR updates an example to use the `match` query instead of the `semantic` query.
Connected to: https://github.com/elastic/developer-docs-team/issues/321


kderusso (commented):
LGTM with one piece of feedback

PR #3629, opened 2025-10-23
Break up explore-analyze code ownership

shainaraskas:
... for more targeted review requests and reduced noise

right now, we can't get away from global notifications for the top-level topics in each of these sections because they're not handled as index files in their respective folders. we could move to that model if we want to cut down on noise further.

Also looked at the other folders where I was familiar with the content, and didn't see any clear wins for breaking down ownership further.


.github/CODEOWNERS
```diff
@@ -9,7 +9,20 @@
 # The docs
 /cloud-account/                             @elastic/admin-docs
 /deploy-manage/                             @elastic/admin-docs
+
 /explore-analyze/                           @elastic/core-docs @elastic/experience-docs
+/explore-analyze/alerts-cases/              @elastic/experience-docs
+/explore-analyze/dashboards/                @elastic/experience-docs
+/explore-analyze/discover/                  @elastic/experience-docs
+/explore-analyze/elastic-inference/         @elastic/developer-docs
```
florent-leborgne:
```suggestion
/explore-analyze/elastic-inference/         @elastic/developer-docs @elastic/experience-docs
```
Will impact experience too

PR #3668, opened 2025-10-27
Update remote clusters examples

eedugon:
Closes https://github.com/elastic/elasticsearch/issues/78216

@shainaraskas : I've used `ADDRESS` (it's generic for a hostname, IP address or FQDN) or `FQDN` instead of `URL` because we don't include a protocol prefix in these fields. URL could be felt misleading or technically incorrect.


shainaraskas (approved):
1 comment! otherwise lgtm ty

PR #3707, opened 2025-10-28
Update upgrade-elastic-agent.md

nathandh22:
Elastic Agents can be upgraded to a maintenance release, as long as it doesn't exceed the Fleet Servers current minor version. Updating the restrictions list to clarify this point and add an example.


karenzone (approved):
Nice improvements both!  

Industry standard (Apple, IBM, Microsoft) is to use "earlier or later" rather than either "higher" or "newer."  If the Elastic standard is different--and it probably shouldn't be--we need to codify that. 

Otherwise, LGTM

PR #3709, opened 2025-10-29
Device control updates

### LINKED ISSUE #3690: [Internal]: Update Device control documentation
### Description

Elastic 9.2.0. introduced the device control function. We have had some discussion on what this function does or doesn't do. Based on the docs's, it's not entirely clear.

We suggest adding the following to the public facing documentation:

- The device control feature works only on MacOS and Windows. Linux is not supported at this time.
- The device control feature only targets USB storage devices. Other USB peripherals (such as Yubikeys, webcams, keyboards, etc) are not impacted by this feature.

I think we also need to document at bit more in-depth what the feature actually protects and scans. I'll tag some members to fill this in.

### Resources

https://www.elastic.co/docs/solutions/security/manage-elastic-defend/trusted-devices
https://www.elastic.co/docs/solutions/security/configure-elastic-defend/configure-an-integration-policy-for-elastic-defend#device-control

### Feature differences

The feature is identical on all deployments.

9.2

### Serverless release

It's already released.

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @WiegerElastic

**Stakeholders:**

@ricardo-estc, @matthewscherer, @szwarckonrad: can you review my statement and correct/add to what I wrote?

benironside:
Fixes #3690 — adds some caveats to the device control docs.


mdbirnstiehl (approved):
LGTM!

PR #3736, opened 2025-10-30
Draft page for Choose your chart

alaudazzi:
Fixes https://github.com/elastic/platform-docs-team/issues/469.

### Note for the reviewers:

- This is an early draft, mainly to test the table structure.
- Content is WIP
- We might include a small image for each type of chart

### DIFF
--- a/explore-analyze/visualize/choose-your-chart.md
+++ b/explore-analyze/visualize/choose-your-chart.md
@@ -11,8 +11,6 @@ products:
 # Choose your chart [choose-your-chart]
 
 Use this guide to pick the most effective visualization. 
-
-## How do I choose?
  
 ### OPTION A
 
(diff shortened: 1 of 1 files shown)


explore-analyze/visualize/choose-your-chart.md
```diff
@@ -0,0 +1,30 @@
...
+mapped_pages:
+  - https://www.elastic.co/guide/en/kibana/current/chart-types.html
+applies_to:
+  stack: ga
+  serverless: ga
+products:
+  - id: kibana
+---
+
+# Choose your chart [choose-your-chart]
```
KOTungseth: In this context, “visualizations” is the correct term.

We use visualizations as the umbrella term for all dashboard elements that display data (bar charts, line charts, tables, metrics, gauges, etc.). The word charts is more limited and should only refer to graph-based visuals like bar or line charts.

You could phrase it like this in the docs:

“You can add visualizations such as bar charts, line charts, tables, metrics, and gauges to your dashboard to analyze your data.”

This keeps the language consistent with how Elastic and similar tools (like Power BI and Tableau) describe dashboard elements.

PR #3758, opened 2025-10-31
Add footnote about the indexing performance differences

yetanothertw:
Adding more information to explain why single-document indexing in Serverless may appear to be slower than in an ECH environment.

Fixes [#385](https://github.com/elastic/docs-content-internal/issues/385).

[Preview of the doc](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/3758/deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings#elasticsearch)

The footnote: 
image
image


deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings.md
```diff
@@ -86,7 +86,7 @@ This table compares Elasticsearch capabilities between {{ech}} deployments and S
 | [**AI Assistant**](/solutions/observability/observability-ai-assistant.md) | ✅ | ✅ | |
 | **Behavioral analytics** | ❌ (deprecated in 9.0) | ❌ | Not available in Serverless |
 | [**Clone index API**](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-clone) | ✅ | **Planned** | Anticipated in a future release |
-| [**Bulk indexing**](/deploy-manage/production-guidance/optimize-performance/indexing-speed.md#_use_bulk_requests) |  ✅ | ✅ | The maximum bulk request response time in {{serverless-short}} is 200ms |
+| [**Bulk indexing**](/deploy-manage/production-guidance/optimize-performance/indexing-speed.md#_use_bulk_requests) |  ✅ | ✅ | The maximum bulk request response time in {{serverless-short}} is 200ms [^1^](#footnote-1) |
```
Tim-Brooks: I would not exactly say this is accurate. It is certainly possible to get higher ACK latency in both Serverless and Hosted. It is not uncommon for ACK latency to be several hundred milliseconds in Hosted if it is on a very loaded cluster.

It's more that you cannot expect write acknowledgement latency to be much lower than 200 milliseconds in Serverless because of the batching period. I would say it is more accurate to say it is the "baseline latency". Or something like that.

yetanothertw: Thanks, Tim! I've applied your feedback in a new commit

PR #3786, opened 2025-11-03
Typo in Fleet reference

lcawl:
This fixes a very minor typo in https://www.elastic.co/docs/reference/fleet/migrate-auditbeat-to-agent

It also removes a link to the stack overview, since it doesn't seem helpful IMO, but that's not mandatory if folks want to keep it.


theletterf (approved):
Thanks!

PR #3789, opened 2025-11-03
Adding clarification to deploy-on-macos-with-mdm.md

nathandh22:
Highlighting an important 5th step under each section for `Enable Full Disk Access`. While the screenshot shows this, it's hard to read and easy to miss.


benironside (approved):
LGTM

PR #3816, opened 2025-11-05
[Obs AI Assistant] Add GPT-5 ratings to the LLM performance matrix

viduni94:
Closes https://github.com/elastic/obs-ai-assistant-team/issues/374

This PR adds the performance ratings of `GPT-5` to the Observability AI Assistant LLM performance matrix.

image


viduni94:
The bug with the Kibana connector for GPT-5 has been solved for the OpenAI provider (only Azure OpenAI is still pending.) 
Therefore, added evaluation results for `GPT-5` with `OpenAI` as the provider instead of `GPT-5-Chat` as it's not recommended for API usage.
cc: @pmoust

PR #3822, opened 2025-11-05
[Cloud Connect] Fix login link

pickypg:
Updated the login link for existing Elastic Cloud accounts to direct users to the correct page after login.

### DIFF
--- a/deploy-manage/deploy/elastic-cloud/create-an-organization.md
+++ b/deploy-manage/deploy/elastic-cloud/create-an-organization.md
@@ -41,9 +41,9 @@ A deployment lets you explore Elastic solutions for Search, Observability, and S
 
 For more information, check the [{{ech}} documentation](cloud-hosted.md).
 
-**One serverless project**
+**Three {{serverless-short}} project**
 
-Serverless projects package {{stack}} features by type of solution:
+{{serverless-short}} projects package {{stack}} features by type of solution:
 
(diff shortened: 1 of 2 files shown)


deploy-manage/monitor/autoops/cc-connect-self-managed-to-autoops.md
```diff
@@ -70,7 +70,8 @@ If you don’t have an existing {{ecloud}} account:
 :sync: existing
 
 If you already have an {{ecloud}} account:
-1. Log in to [{{ecloud}}](https://cloud.elastic.co?page=docs&placement=docs-body).
+1. Log in to [{{ecloud}}](https://cloud.elastic.co/login?redirectTo=%2Fconnect-cluster-services).
+    - The link provided should take you directly to the **Connect your self-managed cluster** page
 2. On your home page, in the **Connected clusters** section, select **Connect self-managed cluster**.
```
yetanothertw: As part of this update, should we remove step 2 if we say that the link in step 1 brings the user to the  **Connect your self-managed cluster** page?

shainaraskas: we should probably keep both, as some people will just want to follow instructions instead of using our link

yetanothertw (approved):
Thank you for updating the link! 🙂

PR #3823, opened 2025-11-05
[Streams] Add LLM note for partition suggestions

mdbirnstiehl:
Partition suggestions require an LLM which was not previously noted.


florent-leborgne (commented):
Can we check what from the security page can be reused here in observability? it doesn't feel right to link to a security page that starts with "Elastic Security", even if the content inside is relevant.

By using snippets, or by moving some of that content to a more central place?

PR #3824, opened 2025-11-05
[Synthetics] Add namespace option docs

emilioalvap:
### Summary

Add missing namespace option docs added on https://github.com/elastic/synthetics/pull/1021.


florent-leborgne (commented):
Very small styling suggestion.

I also wonder, is this option only available from a certain stack version (9.1 or 9.2 for example)? If so, we'll need to specify it. Let me know and I can guide you on how

PR #3830, opened 2025-11-05
Fix JQ command for processing mapping.json

octavioranieri:
Updated command to process mapping.json output with JQ.


kilfoyle (approved):
Awesome. Thanks @octavioranieri!

PR #3842, opened 2025-11-06
Clarify GA status and technical preview in warning

nimarezainia:
Updated warning message to clarify GA status and technical preview for Active Directory interfacing.


vishaangelova (approved):
Thanks @nimarezainia! LGTM 🚀 Just a tiny suggestion for your consideration.

vishaangelova:
I really like your solution, @karenzone! It makes perfect sense. @nimarezainia, with this solution, the GA status of the feature is implied and we only highlight the part that is in technical preview. I’m happy to take over this, make the changes, and then we can merge this, if you agree.

PR #3847, opened 2025-11-07
Add initial docs for `bfloat16` and `on_disk_rescore`  for dense vectors

thecoop:
Add docs for bfloat16 and on_disk_rescore for changes in https://github.com/elastic/elasticsearch/pull/138492

### DIFF
--- a/deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
+++ b/deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
@@ -71,10 +71,19 @@ Here are estimates for different element types and quantization levels:
 
 If you're using HNSW, the graph must also be in memory. To estimate the required bytes, use the following formula below. The default value for the HNSW `m` parameter is `16`.
 
+For `element_type: float`:
 ```{math}
 \begin{align*}
 estimated\ bytes &= num\_vectors \times 4 \times m \\
 &= num\_vectors \times 4 \times 16
+\end{align*}
+```
+
+For `element_type: bfloat16`:
+```{math}
+\begin{align*}
+estimated\ bytes &= num\_vectors \times 2 \times m \\
+&= num\_vectors \times 2 \times 16
 \end{align*}
 ```
 
--- a/solutions/search/vector/knn.md
+++ b/solutions/search/vector/knn.md
@@ -327,8 +327,7 @@ POST quantized-image-index/_search
 ```
 
 ### BFloat16 vector encoding [knn-search-bfloat16]
-```
-{applies_to}
+```{applies_to}
 stack: ga 9.3
 ```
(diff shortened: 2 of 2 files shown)


deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
```diff
@@ -122,13 +123,11 @@ You can check the current value in `KiB` using `lsblk -o NAME,RA,MOUNTPOINT,TYPE
 ::::
 
 
-## Use Direct IO when the vector data does not fit in RAM
+## Use on-disk rescoring when the vector data does not fit in RAM
```
shainaraskas: is the old direct IO guidance still valid in 9.1? did your team plan to not move forward with it (i.e. it is going from preview to removed)? Wonder if we need to keep it visible, but if it is going from preview to removed, your approach of just editing it out is ok.

solutions/search/vector/knn.md
```diff
@@ -325,6 +325,15 @@ POST quantized-image-index/_search
 }
 ```
 
+### BFloat16 vector encoding [knn-search-bfloat16]
+```
+{applies_to}
+stack: ga 9.3
+```
+Instead of storing raw vectors as 4-byte values, you can use `element_type: bfloat16` to store each dimension as a 2-byte value. This can be useful if your indexed vectors are at bfloat16 precision already, or if you wish to reduce the disk space required to store vector data. Elasticsearch will automatically truncate 4-byte float values to 2-byte bfloat16 values when indexing vectors.
```
shainaraskas:
```suggestion
Instead of storing raw vectors as 4-byte values, you can use `element_type: bfloat16` to store each dimension as a 2-byte value. This can be useful if your indexed vectors are at bfloat16 precision already, or if you want to reduce the disk space required to store vector data. When this element type is used, {{es}} automatically truncates 4-byte float values to 2-byte bfloat16 values when indexing vectors.
```

solutions/search/vector/knn.md
```diff
@@ -1227,6 +1236,16 @@ This example will:
...
 
+#### The `on_disk_rescore` option
+```{applies_to}
+stack: preview 9.3
+serverless: unavailable
+```
+
+By default, Elasticsearch will read raw vector data into memory to perform rescoring. This may have an effect on performance if the vector data is too large to all fit in off-heap memory at once. By specifying the `on_disk_rescore: true` index setting, Elasticsearch will read vector data from disk directly during rescoring.
+
+Note that this setting will only apply to newly indexed vectors; to apply the option to all vectors in the index, the vectors must be re-indexed or force-merged after changing the setting.
```
shainaraskas:
```suggestion
This setting only applies to newly indexed vectors. To apply the option to all vectors in the index, the vectors must be re-indexed or force-merged after changing the setting.
```

shainaraskas (approved):
generally looks good. provided some recommended style edits for you.

PR #3860, opened 2025-11-09
Revise TCP port requirements for ECE remote clusters

eedugon:
Updated TCP port requirements for ECE proxies and load balancers based on the selected security model.


yetanothertw (approved):
LGTM! 👍

PR #3921, opened 2025-11-12
AutoOps for SM: update storage regions list + remove list from Regions page

wajihaparvez:
This PR:
- updates the list of regions where self-managed AutoOps metrics are stored. The list is in a snippet so the change will be visible on two pages.
- removes the same list from the Regions page because technically AutoOps for SM clusters can be used anywhere. 
- Adds a colon before a list + fixes an anchor link.

Closes: https://github.com/elastic/opex-product/issues/668, https://github.com/elastic/opex-product/issues/664


yetanothertw (approved):
LGTM 🌻

Issue #3924, opened 2025-11-13
Few issues with the data views page


florent-leborgne:
- Issue with link from the data views UI not jumping straight to the correct section

- noticed a couple other things in these docs:
   - The "data view field cache" title isn't capitalized
   - "Format data fields" should probably be "Format data view fields" (plus other references to "data fields" in those docs)

PR #3925, opened 2025-11-13
Fix duplicated anchors and a few terms in data views docs

florent-leborgne:
Fixes: https://github.com/elastic/docs-content/issues/3924


davismcphee (approved):
Thanks!

PR #3927, opened 2025-11-13
Fix typos in the GROK patterns

gbamparop:
There were some instances where extra curly braces were used. For example,`%{{PATTERN}}` was displayed instead of `%{PATTERN}`.


yetanothertw (approved):
Thank you for fixing that, @gbamparop 🙏

PR #3932, opened 2025-11-13
Move own CA transport layer mtls guidance to security docs

shainaraskas:
Ports over and lightly edits the content from https://github.com/elastic/elasticsearch/pull/137688

also interlinks from every place where we introduce opportunities to BYO cert for transport mTLS.

core change:  [deploy-manage/security/external-ca-transport.md](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/3932/deploy-manage/security/external-ca-transport) (placed so it applies to both ECK and fully self-managed deployments)

### DIFF
--- a/release-notes/elastic-cloud-serverless/index.md
+++ b/release-notes/elastic-cloud-serverless/index.md
@@ -71,21 +71,15 @@ Review the changes, fixes, and more to {{serverless-full}}.
 * Extracts `AbstractGeoIpDownloader` to share concurrency logic across GeoIP downloaders [#137660]({{es-pull}}137660)
 * Iterates directly over `RoutingNode` contents to reduce allocation overhead [#137694]({{es-pull}}137694)
 * Speeds up sorts that use secondary sort fields [#137533]({{es-pull}}137533)
-* Updates HDFS version references in the documentation [#137576]({{es-pull}}137576)
 * Reduces worst-case Inference API latency by removing an additional 50 ms delay for non–rate-limited requests [#136167]({{es-pull}}136167)
-* Updates {{esql}} documentation to cover newly supported data types [#137726]({{es-pull}}137726)
 * Uses the `DEFAULT_UNSORTABLE` topN encoder for `TSID_DATA_TYPE` in {{esql}} to improve sorting behavior [#137706]({{es-pull}}137706)
 * Transitions Elastic Indexing Service auth polling to a single-node persistent task for improved reliability [#136713]({{es-pull}}136713)
-* Documents {{esql}} decay functions using the `math` directive [#137369]({{es-pull}}137369)
-* Adds an `applies_to` label to the logsdb message default sort setting [#137767]({{es-pull}}137767)
 * Makes {{esql}} field fusion generic so it can be reused across more field types [#137382]({{es-pull}}137382)
 * Releases the {{esql}} `decay` function [#137830]({{es-pull}}137830)
 * Adds additional APM attributes to coordinator-phase duration metrics for richer tracing [#137409]({{es-pull}}137409)
 * Adds telemetry to track CPS usage [#137705]({{es-pull}}137705)
 * Introduces simple bulk loading for binary doc values to improve indexing throughput [#137860]({{es-pull}}137860)
 * Uses IVF_PQ for GPU-based index builds on large datasets to improve vector indexing performance [#137126]({{es-pull}}137126)
-* Updates `semantic_text` documentation to link to the authoritative chunking settings guide [#137963]({{es-pull}}137963)
-* Refines `semantic_text` documentation based on user feedback [#137970]({{es-pull}}137970)
 * Aligns match-phase shard APM metrics with the originating search request context [#137196]({{es-pull}}137196)
 * Improves {{serverless-short}} filtering behavior when creating resources from existing configurations [#137850]({{es-pull}}137850)
 * Refactors model field parsing in `AnthropicChatCompletionStreamingProcessor` to better handle model variants [#137926]({{es-pull}}137926)
@@ -94,7 +88,6 @@ Review the changes, fixes, and more to {{serverless-full}}.
 * Adds additional DEBUG-level logging for authentication failures [#137941]({{es-pull}}137941)
 * Adds support for an extra output field in the {{esql}} `TOP` function [#135434]({{es-pull}}135434)
 * Introduces the `INDEX_SHARD_COUNT_FORMAT` setting for index shard count formatting [#137210]({{es-pull}}137210)
-* Documents RCS Strong Verification configuration and usage [#137822]({{es-pull}}137822)
 * Implements an OpenShift AI integration for chat completion, embeddings, and reranking workloads [#136624]({{es-pull}}136624)
 * Adds `first()` and `last()` aggregation functions to {{esql}} [#137408]({{es-pull}}137408)
 * Adds support for the `project_routing` parameter on `_search` and `_async_search` requests [#137566]({{es-pull}}137566)
@@ -104,8 +97,6 @@ Review the changes, fixes, and more to {{serverless-full}}.
 * Adds timezone support to {{esql}} `DATE_TRUNC`, `BUCKET`, and `TBUCKET` functions [#137450]({{es-pull}}137450)
 * Further improves bulk loading performance for binary doc values [#137995]({{es-pull}}137995)
 * Updates the Gradle wrapper to version `9.2.0`
-* Adds a synthetics test pipeline for UIAM
-* Improves the resolution for authenticating SAML realm in UIAM
 * Improves logging for the sampled metrics provider
 * Updates `BlobCacheIndexInput` to use `sliceDescription` as the resource description when available, improving diagnostics
 * Switches APM trace detection to use `hasApmTraceContext` and its variant APIs
@@ -131,7 +122,6 @@ Review the changes, fixes, and more to {{serverless-full}}.
 * Fixes OTLP responses to return the correct response type for partial successes [#137718]({{es-pull}}137718)
 * Fixes the get data stream API when a data stream's index mode has been changed to `time_series` [#137852]({{es-pull}}137852)
 * Ensures `include_execution_metadata` in {{esql}} always returns data, including for local-only queries [#137641]({{es-pull}}137641)
-* Fixes the DiskBBQ example in the release highlights documentation [#137960]({{es-pull}}137960)
 * Fixes an {{esql}} vector similarity concurrency issue affecting byte vectors [#137883]({{es-pull}}137883)
 * Reverts a previous change to `statsByShard` that regressed performance for very large shard counts [#137984]({{es-pull}}137984)
 * Fixes scalability issues when updating {{ml}} calendar events [#136886]({{es-pull}}136886)
@@ -458,7 +448,6 @@ Review the changes, fixes, and more to {{serverless-full}}.
 * Automatically copies source data into the alerts-as-data documents for other ES Query rule types [#230010]({{kib-pull}}230010)
 * Replaces the dashboard editor toolbar with the **Add** menu [#230324]({{kib-pull}}230324)
 * Adds support for package spec v3.5 [#235942]({{kib-pull}}235942)
-* Adds in-product documentation for the {{esql}} `FORK` command [#236494]({{kib-pull}}236494)
 * Adds **View in discover** button in alert details page for SLO burn rate and ES query rules [#233855]({{kib-pull}}233855)
 
 ### Fixes [serverless-changelog-09292025-fixes]


deploy-manage/security/_snippets/own-ca-warning.md
```diff
@@ -0,0 +1,5 @@
+:::{warning}
+Transport connections between {{es}} nodes are security-critical and you must protect them carefully. Malicious actors who can observe or interfere with node-to-node transport traffic can read or modify cluster data. A malicious actor who can establish a transport connection might be able to invoke system-internal APIs, including APIs that read or modify cluster data.
```
DaveCTurner: I think we need to qualify that this means the _raw_ (i.e. plaintext, unencrypted) traffic. Without that, "node-to-node transport traffic" might also mean the encrypted traffic.

deploy-manage/security/set-up-basic-security.md
```diff
@@ -28,6 +28,14 @@ In this guide, you will learn how to:
 
 Refer to [Transport TLS/SSL settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#transport-tls-ssl-settings) for the complete list of available settings in {{es}}.
 
+## Considerations for using an external CA
+
+You might choose to use an external CA to generate transport certificates for node-to-node connections.
```
DaveCTurner: The word "external" suggests to me a third-party CA, but we also care about users using a CA that is private to their organization (but widely used within the organization) which they might consider to be "internal". Could we find some wording that covers both of these cases? Really it's any CA other than the one you can manage using the `elasticsearch-certutil` tool.

deploy-manage/security/k8s-transport-settings.md
```diff
@@ -75,6 +75,12 @@ spec:
 
 ## Issue node transport certificates with third-party tools [k8s-transport-third-party-tools]
 
+:::{warning}
+Transport connections between {{es}} nodes are security-critical and you must protect them carefully. Malicious actors who can observe or interfere with node-to-node transport traffic can read or modify cluster data. A malicious actor who can establish a transport connection might be able to invoke system-internal APIs, including APIs that read or modify cluster data.
```
DaveCTurner: This first paragraph might be better-placed at the top of the page, since the whole page is about securing transport connections.

deploy-manage/security/k8s-transport-settings.md
```diff
@@ -75,6 +75,12 @@ spec:
...
+
+If you choose to issue node transport certificates using third-party tools, then carefully review [](/deploy-manage/security/self-tls-considerations.md) to ensure that the certificates that you provide meet the security requirements for transport connections.
+:::
```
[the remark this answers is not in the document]
shainaraskas: good catch. forgot to replace it (this was the first place I added it before I decided to make it a snippet)

deploy-manage/security/self-tls-considerations.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Transport certificates vs. HTTP certificates
+
+Transport certificates have different security requirements than [HTTP certificates](/deploy-manage/security/secure-cluster-communications.md#encrypt-http-communication). HTTP connections don't typically use mTLS because HTTP has its own authentication mechanisms. Because of this, HTTP certificates usually don't need to include the `clientAuth` value in their Extended Key Usage extension. HTTP certificates can come from public or organization-wide certificate authorities, while transport certificates should use a cluster-specific private CA. In most cases, you should not use the same certificate for both HTTP and transport connections.
+
+## Turning off mTLS for transport connections [turn-off-mtls]
+
+If your environment has some other way to prevent unauthorized node-to-node connections, you can disable mTLS by setting `xpack.security.transport.ssl.client_authentication: none`. You can still use non-mutual TLS for encryption by setting `xpack.security.transport.ssl.enabled: true`. With non-mutual TLS, transport certificates don't require the `clientAuth` value in the Extended Key Usage extension.
```
DaveCTurner: Likewise here could we spell out that non-mutual TLS still guarantees confidentiality and integrity?

deploy-manage/security/self-tls-considerations.md
```diff
@@ -0,0 +1,37 @@
...
+    self:
+    eck:
+products:
+  - id: elasticsearch
+navigation_title: External CA considerations
+---
+
+# Considerations for using an external CA for transport layer security
+
+By default, {{es}} uses mutual TLS (mTLS) to secure node-to-node transport connections. With mTLS, data is encrypted in transit and both nodes must present valid certificates when connecting. Each node requires that certificates be issued by a trusted certificate authority, ensuring that only authorized nodes can connect. Configure trusted certificate authorities using settings in the [`xpack.security.transport.ssl.*`](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#transport-tls-ssl-settings) namespace, such as `xpack.security.transport.ssl.certificate_authorities` and `xpack.security.transport.ssl.truststore.path`.
```
shainaraskas:
```suggestion
By default, {{es}} uses mutual TLS (mTLS) to secure node-to-node transport connections. Mutual TLS means that data is encrypted in transit, ensuring confidentiality and integrity, and also that both nodes in a connection must present a valid certificate to the other node when establishing the connection. Each node requires that certificates be issued by a trusted certificate authority, ensuring that only authorized nodes can connect. Configure trusted certificate authorities using settings in the [`xpack.security.transport.ssl.*`](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#transport-tls-ssl-settings) namespace, such as `xpack.security.transport.ssl.certificate_authorities` and `xpack.security.transport.ssl.truststore.path`. 
```

deploy-manage/security/set-up-basic-security.md
```diff
@@ -59,7 +78,7 @@ For the transport layer, we recommend using a separate, dedicated CA instead of
         1. Enter the password for your CA, or press **Enter** if you did not configure one in the previous step.
         2. Create a password for the certificate and accept the default file name.
 
-            The output file is a keystore named `elastic-certificates.p12`. This file contains a node certificate, node key, and CA certificate.
+            The output file is a keystore named `elastSic-certificates.p12`. This file contains a node certificate, node key, and CA certificate.
```
DaveCTurner:
```suggestion
            The output file is a keystore named `elastic-certificates.p12`. This file contains a node certificate, node key, and CA certificate.
```

deploy-manage/security/self-tls-considerations.md
```diff
@@ -0,0 +1,37 @@
...
+    eck:
+products:
+  - id: elasticsearch
+navigation_title: Private or 3P CA considerations
+---
+
+# Considerations for using an private or third-party CA for transport layer security
+
+By default, {{es}} uses mutual TLS (mTLS) to secure node-to-node transport connections. Mutual TLS means that data is encrypted in transit, ensuring confidentiality and integrity, and also that both nodes in a connection must present a valid certificate to the other node when establishing the connection. Each node requires that certificates be issued by a trusted certificate authority, ensuring that only authorized nodes can connect. Configure trusted certificate authorities using settings in the [`xpack.security.transport.ssl.*`](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#transport-tls-ssl-settings) namespace, such as `xpack.security.transport.ssl.certificate_authorities` and `xpack.security.transport.ssl.truststore.path`. 
+
```
DaveCTurner: Maybe we should add here some more context about the different kinds of CA: Elasticsearch comes with the `elasticsearch-certutil` tool that allows you to create a properly-configured private CA and issue certificates for the nodes in the cluster, but it is also compatible with certificates obtained from a different CA managed using other tools. This page is about the latter case.

deploy-manage/security/set-up-basic-security.md
```diff
@@ -17,26 +17,45 @@ Configuring TLS between nodes is the basic security setup to prevent unauthorize
...
+
+### Provide certificates from a private or third-party CA [private-3p]
+
+You might choose to use a private or third-party CA to generate transport certificates for node-to-node connections.
+
+Transport connections between {{es}} nodes are security-critical and you must protect them carefully. Malicious actors who can observe or interfere with unencrypted node-to-node transport traffic can read or modify cluster data. A malicious actor who can establish a transport connection might be able to invoke system-internal APIs, including APIs that read or modify cluster data.
+
+Carefully review [](/deploy-manage/security/self-tls-considerations.md) to ensure that the certificates that you provide meet the security requirements for transport connections.
+
+After you obtain your certificate, place the certificate file in the `$ES_PATH_CONF` directory on **every** node in your cluster.
```
DaveCTurner: ... and configure Elasticsearch to use these files. Also it's not just the certificate you obtained, you also need to tell Elasticsearch which CA certificates it can trust, and it needs the private key too. TBH I think it'd be simpler to drop this sentence, it is adequately covered elsewhere.

deploy-manage/security/set-up-basic-security.md
```diff
@@ -17,26 +17,45 @@ Configuring TLS between nodes is the basic security setup to prevent unauthorize
...
 You can add as many nodes as you want in a cluster but they must be able to communicate with each other. The communication between nodes in a cluster is handled by the transport module. To secure your cluster, you must ensure that internode communications are encrypted and verified, which is achieved with mutual TLS.
 
 In a secured cluster, {{es}} nodes use certificates to identify themselves when communicating with other nodes.
 
 The cluster must validate the authenticity of these certificates. The recommended approach is to trust a specific certificate authority (CA). When nodes are added to your cluster they must use a certificate signed by the same CA.
 
-For the transport layer, we recommend using a separate, dedicated CA instead of an existing, possibly shared CA so that node membership is tightly controlled. Use the `elasticsearch-certutil` tool to generate a CA for your cluster.
+For the transport layer, we recommend using a separate, dedicated CA instead of an existing, possibly shared CA so that node membership is tightly controlled.
+
+When you manually set up transport TLS, you can choose from the following CA options:
```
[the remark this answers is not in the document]
DaveCTurner: works for me 👍

DaveCTurner (commented):
Looks good, my only outstanding concern is around the "private or third-party" terminology. I wonder if we can come up with a neater term for this.

deploy-manage/security/set-up-basic-security.md
shainaraskas: 🤦

deploy-manage/security/self-tls-considerations.md
```diff
@@ -0,0 +1,39 @@
...
+applies_to:
+  deployment:
+    self:
+    eck:
+products:
+  - id: elasticsearch
+navigation_title: External CA considerations
+---
+
+# Considerations for using an external CA for transport layer security
```
DaveCTurner:
```suggestion
# Considerations for using an external certificate authority for transport layer security
```

 I'd prefer we avoid the `CA` abbreviation, it's too overloaded, but if we do use it we should spell it out until we've defined it.

Or maybe just

```suggestion
# Using an external certificate authority for transport layer security
```

Also "transport layer security" is what TLS stands for, but that applies to all secure network connections (HTTPS, LDAPS etc). Here we're only talking about "transport" in the sense of Elasticsearch node-to-node connections which is a confusing overload of terminology but we should at least avoid the phrase "transport layer security".

How about these?

```suggestion
# Using an external certificate authority to secure node-to-node connections
```

```suggestion
# Securing node-to-node connections using an external certificate authority
```

deploy-manage/security/self-tls-considerations.md
```diff
@@ -0,0 +1,39 @@
...
+
+{{es}} comes with a built-in tool called [`elasticsearch-certutil`](/deploy-manage/security/set-up-basic-security.md), which you can use to create and manage a dedicated certificate authority for each of your clusters, and to issue TLS certificates from this certificate authority. If you prefer not to use `elasticsearch-certutil`, then you must obtain the certificates from another certificate authority using standard TLS tools. This page explains the requirements and best practices to ensure that these certificates work correctly and protect your cluster properly.
```
DaveCTurner: Can we define the `external certificate authority` term in this paragraph?

deploy-manage/security/secure-cluster-communications.md
```diff
@@ -57,36 +57,43 @@ Securing this layer prevents unauthorized nodes from joining your cluster and pr
...
+::::{tab-item} ECH and Serverless
 :sync: ech
 {{es}} transport security is fully managed by Elastic, and no configuration is required.
-:::
+::::
 
-:::{tab-item} ECE
+::::{admonition}{tab-item} ECE
 :sync: ece
 {{es}} transport security is fully managed by {{ece}} platform, and no configuration is required.
```
wajihaparvez:
```suggestion
{{es}} transport security is fully managed by the {{ece}} platform, and no configuration is required.
```

deploy-manage/security/set-up-basic-security.md
```diff
@@ -17,26 +17,35 @@ Configuring TLS between nodes is the basic security setup to prevent unauthorize
...
 This document focuses on the **manual configuration** of TLS for [{{es}} transport protocol](./secure-cluster-communications.md#encrypt-internode-communication) in self-managed environments. Use this approach if you want to provide your own TLS certificates, generate them with Elastic’s tools, or have full control over the configuration. Alternatively, {{es}} can [automatically generate and configure HTTPS certificates](./self-auto-setup.md) for you.
 
-::::{note}
-For other deployment types, such as {{ech}}, {{ece}}, or {{eck}}, refer to [](./secure-cluster-communications.md).
-::::
-
 In this guide, you will learn how to:
 
-* [Generate a Certificate Authority (CA) and a server certificate using the `elasticsearch-certutil` tool](#generate-certificates).
+* [Either provide or generate security certificates](#obtain-certificates).
```
wajihaparvez:
```suggestion
* [Generate or provide security certificates](#obtain-certificates).
```
To align with the positioning of those sections on the page

deploy-manage/security/_snippets/own-ca-warning.md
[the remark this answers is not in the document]
shainaraskas: @DaveCTurner can you answer this? not sure if these two statements technically mean two different things

deploy-manage/security/set-up-basic-security.md
```diff
@@ -61,14 +70,23 @@ For the transport layer, we recommend using a separate, dedicated CA instead of
 
             The output file is a keystore named `elastic-certificates.p12`. This file contains a node certificate, node key, and CA certificate.
 
-3. On **every** node in your cluster, copy the `elastic-certificates.p12` file to the `$ES_PATH_CONF` directory.
+
+### Provide certificates from an external CA [external-ca]
+
+You might choose to use an external CA to generate transport certificates for node-to-node connections. An external CA is any CA that is not managed using `elasticsearch-certutil`.
+
+Transport connections between {{es}} nodes are security-critical and you must protect them carefully. Malicious actors who can observe or interfere with unencrypted node-to-node transport traffic can read or modify cluster data. A malicious actor who can establish a transport connection might be able to invoke system-internal APIs, including APIs that read or modify cluster data.
```
wajihaparvez: If you end up editing the [own-ca-warning.md](https://github.com/elastic/docs-content/pull/3932/files#diff-7795a6e345ac3daae67ac87ebfac13765e8320c1a95b73ad824ef823128bc10c) snippet based on my suggestions, apply the same changes here

Also, shouldn't this also be in a warning block? That way you can even reuse the snippet here

deploy-manage/security/set-up-basic-security.md
```diff
@@ -17,26 +17,35 @@ Configuring TLS between nodes is the basic security setup to prevent unauthorize
...
 
-* [Generate a Certificate Authority (CA) and a server certificate using the `elasticsearch-certutil` tool](#generate-certificates).
+* [Either provide or generate security certificates](#obtain-certificates).
 * [Configure your {{es}} nodes to use the generated certificate for the transport layer](#encrypt-internode-communication).
 
 Refer to [Transport TLS/SSL settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#transport-tls-ssl-settings) for the complete list of available settings in {{es}}.
 
-## Generate the certificate authority [generate-certificates]
+::::{note}
+For other deployment types, such as {{ech}}, {{ece}}, or {{eck}}, refer to [](./secure-cluster-communications.md).
```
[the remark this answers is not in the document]
shainaraskas: good catch. there are no procedures for ech and serverless in the child pages for this page, but we do explain who does what for those deployment types in the overview. we also explain who does what for self-managed- I'll add ech/serverless here because there is applicable content.

deploy-manage/security/_snippets/own-ca-warning.md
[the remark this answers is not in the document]
DaveCTurner: They're separate things. Establishing a connection is different from interfering with traffic on existing connections.

DaveCTurner (changes requested):
Sorry a couple of the recent edits need adjusting

deploy-manage/security/external-ca-transport.md
```diff
@@ -0,0 +1,40 @@
...
+
+::::{warning}
+Transport connections between {{es}} nodes are security-critical and you must protect them carefully. Malicious actors who can observe or interfere with unencrypted node-to-node transport traffic can read or modify cluster data. A malicious actor who can establish a transport connection might be able to invoke system-internal APIs, including APIs that read or modify cluster data.
+::::
+
+## mTLS transport certificate requirements for external CAs
+
+Obtain your transport certificates from a certificate authority that only issues certificates to {{es}} nodes permitted to connect to your cluster. Do not use a public certificate authority or an organization-wide private certificate authority, because these issue certificates to entities beyond your authorized cluster nodes. Use a dedicated private certificate authority for each {{es}} cluster.
+
+Certificates used for mTLS must either have no Extended Key Usage extension, or include both `clientAuth` and `serverAuth` values in the extension. Public certificate authorities typically omit the `clientAuth` value in the Extended Key Usage extension, making them unsuitable for mTLS.
```
eedugon:
```suggestion
Certificates used for transport mTLS must either have no Extended Key Usage extension, or include both `clientAuth` and `serverAuth` values in the extension, because the same certificate is used by the node when acting as client and server. Public certificate authorities typically omit the `clientAuth` value in the Extended Key Usage extension, making them unsuitable for {{es}} transport mTLS. 
```

A public certificate without clientAuth is perfectly valid for mTLS in general, but not for ES transport mTLS.

deploy-manage/security/external-ca-transport.md
```diff
@@ -0,0 +1,40 @@
...
+
+This page explains the requirements and best practices to ensure that certificates generated using an external CA work correctly and protect your cluster properly.
+
+::::{warning}
+Transport connections between {{es}} nodes are security-critical and you must protect them carefully. Malicious actors who can observe or interfere with unencrypted node-to-node transport traffic can read or modify cluster data. A malicious actor who can establish a transport connection might be able to invoke system-internal APIs, including APIs that read or modify cluster data.
+::::
+
+## mTLS transport certificate requirements for external CAs
+
+Obtain your transport certificates from a certificate authority that only issues certificates to {{es}} nodes permitted to connect to your cluster. Do not use a public certificate authority or an organization-wide private certificate authority, because these issue certificates to entities beyond your authorized cluster nodes. Use a dedicated private certificate authority for each {{es}} cluster.
```
eedugon: This first sentence should probably be considered or used for comparison in the section "Transport vs HTTP".

We probably want to explain that this is not that relevant for HTTP, and for HTTP, for operational purposes (and because we have extra authentication and authorization mechanisms), it's common to not use dedicated CAs per cluster, and even public / organizational CAs that are automatically trusted by the clients.

Of course this will depend on the use case, as in certain use cases it might have sense to have the HTTP layer also super-protected at TLS level.

We should end up with the recommendation that as minimum, if they use private CAs, they should create a private CA to generate transport certs, and another private CA to generate HTTP certs.

This relates with the comment in the "transport vs HTTP" section.

deploy-manage/security/external-ca-transport.md
DaveCTurner: I disagree, it depends whether you read "mTLS" as "everything about setting up a mutually-authenticated TLS connection" or "client-certificate authentication". The general practice seems to be the latter, and that's how we're using it here.

deploy-manage/security/external-ca-transport.md
```diff
@@ -0,0 +1,40 @@
...
+
+Certificates used for mTLS must either have no Extended Key Usage extension, or include both `clientAuth` and `serverAuth` values in the extension. Public certificate authorities typically omit the `clientAuth` value in the Extended Key Usage extension, making them unsuitable for mTLS. 
+
+### Transport certificates vs. HTTP certificates
+
+Transport certificates have different security requirements than [HTTP certificates](/deploy-manage/security/secure-cluster-communications.md#encrypt-http-communication). HTTP connections don't typically use mTLS because HTTP has its own authentication mechanisms. Because of this, HTTP certificates usually don't need to include the `clientAuth` value in their Extended Key Usage extension. HTTP certificates can come from public or organization-wide certificate authorities, while transport certificates should use a cluster-specific private CA. In most cases, you should not use the same certificate for both HTTP and transport connections.
```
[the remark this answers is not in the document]
DaveCTurner: As [above](https://github.com/elastic/docs-content/pull/3932#discussion_r2537611080) I would rather not split hairs about whether mTLS refers just to the client-certificate authentication or the whole connection process.

Likewise, you're right that there are many other concerns (mostly addressed elsewhere in this text) about how to obtain these certificates. Yet, a very common mistake is for users to use the same certificate for both uses, and it is vitally important to have a sentence calling out this case.

deploy-manage/security/external-ca-transport.md
```diff
@@ -20,15 +20,17 @@ This page explains the requirements and best practices to ensure that certificat
...
+Transport certificates have different security requirements than [HTTP certificates](/deploy-manage/security/secure-cluster-communications.md#encrypt-http-communication). HTTP server certificates don't require the `clientAuth` Extended Key Usage extension because they are used solely for server authentication, regardless of whether mTLS is enabled. In practice, HTTP connections don't typically use mTLS because HTTP has its own authentication mechanisms. 
+
+HTTP certificates can come from public or organization-wide certificate authorities, while transport certificates should use a cluster-specific private CA. In most cases, you should not use the same CA or certificate for both HTTP and transport connections.
```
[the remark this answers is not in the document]
eedugon: I like it, a lot. Thanks

deploy-manage/security/external-ca-transport.md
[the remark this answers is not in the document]
eedugon: > a very common mistake is for users to use the same certificate for both uses, and it is vitally important to have a sentence calling out this case.

yes, totally agree!

I realize now that all the misunderstanding and what I was considering "not accurate" in the writing it's because I have in my mind a standard `client to server` communication model, and I believe you are thinking of a `server - server` communication flow.

With that a `server-to-server` communication model in mind (typical in clusters and used in Elasticsearch transport), then the statement about mTLS requiring both `clientAuth` and `serverAuth` extension in the server certificate becomes accurate and valid. Sorry I wasn't finding it accurate because I was thinking in a client-to-server basic flow.

Anyway I think latest changes by @shainaraskas already improve this.

deploy-manage/security/external-ca-transport.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Turning off mTLS for transport connections [turn-off-mtls]
+
+If your environment has some other way to prevent unauthorized node-to-node connections, you can disable mTLS by setting `xpack.security.transport.ssl.client_authentication: none`. You can still use non-mutual TLS to ensure the confidentiality and integrity of node-to-node traffic by setting `xpack.security.transport.ssl.enabled: true`. With non-mutual TLS, transport certificates don't require the `clientAuth` value in the Extended Key Usage extension.
+
+::::{warning}
+Turning off mTLS by setting `xpack.security.transport.ssl.client_authentication` to `optional` or `none` allows anyone with network access to establish transport connections. Malicious actors can use these connections to invoke system-internal APIs that may read or modify cluster data. Use mTLS to
```
shainaraskas:
```suggestion
Turning off mTLS by setting `xpack.security.transport.ssl.client_authentication` to `optional` or `none` allows anyone with network access to establish transport connections. Malicious actors can use these connections to invoke system-internal APIs that might read or modify cluster data. Use mTLS to 
```

deploy-manage/security/external-ca-transport.md
```diff
@@ -0,0 +1,42 @@
...
+
+Certificates used for transport mTLS must either have no Extended Key Usage extension, or include both `clientAuth` and `serverAuth` values in the extension. Public certificate authorities typically omit the `clientAuth` value in the Extended Key Usage extension, making them unsuitable for mTLS. 
+
+### Transport certificates vs. HTTP certificates
+
+Transport certificates have different security requirements than [HTTP certificates](/deploy-manage/security/secure-cluster-communications.md#encrypt-http-communication). HTTP server certificates don't require the `clientAuth` Extended Key Usage extension because they are used solely for server authentication, regardless of whether mTLS is enabled. In practice, HTTP connections don't typically use mTLS because HTTP has its own authentication mechanisms. 
+
+HTTP certificates can come from public or organization-wide certificate authorities, while transport certificates should use a cluster-specific private CA. In most cases, you should not use the same CA or certificate for both HTTP and transport connections.
```
shainaraskas:
```suggestion
HTTP certificates can come from public or organization-wide certificate authorities, while transport certificates should use a cluster-specific private CA. In most cases, you should not use the same certificate for both HTTP and transport connections.
```

Issue #3914, opened 2025-11-12
[Internal]: "Set up a TSDS" contains invalid query


pquentin:
### Description

See https://github.com/elastic/docs-content/blob/126dfecc5ce6f3392cfc032541a0506fd2672704/manage-data/data-store/data-streams/set-up-tsds.md

The page suggests that `lifecycle.enabled` is an index setting, whereas it is actually at the `template` level.

Current documented query (edited for brevity):

```
PUT _index_template/my-weather-sensor-index-template
{
  "template": {
    "settings": {
      "lifecycle": { 
        "enabled": true
      }
    }
  }
}
```

This fails with `"reason": "index_template [my-weather-sensor-index-template] invalid, cause [Validation Failed: 1: unknown setting [index.lifecycle.enabled] did you mean [index.lifecycle.name]?;]"`.

Here's a correct query:

```
PUT _index_template/my-weather-sensor-index-template
{
  "template": {
    "lifecycle": { 
      "enabled": true
    }
  }
}
```

This text could also be clarified: https://github.com/elastic/docs-content/blob/126dfecc5ce6f3392cfc032541a0506fd2672704/manage-data/data-store/data-streams/set-up-tsds.md?plain=1#L88-L89

### Resources

N/A

### Which documentation set does this change impact?

Elastic On-Prem and Cloud (all)

### Feature differences

N/A

### What release is this request related to?

N/A

### Serverless release

N/A

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @pquentin 

**Stakeholders:** @nielsbauman @masseyke

PR #3945, opened 2025-11-14
Fix index template step in Set up a TSDS

marciw:
Fixes https://github.com/elastic/docs-content/issues/3914

- Move lifecycle object to the right place in the example
- Clarify intro text


kilfoyle (approved):
LGTM! 🦘

pquentin (approved):
Thanks! LGTM.

Issue #3220, opened 2025-09-29
Restructure AutoOps for self-managed clusters troubleshooting page


wajihaparvez:
Too much info for a definition list format. Either move things into new pages/sections, or reformat (dropdowns?)

PR #3955, opened 2025-11-17
Enhance and restructure AutoOps pages

wajihaparvez:
In this PR, I have:
  - added a child to the **Troubleshooting** page because it had too much info. More child pages will be added in the future.
  - made various general fixes.
  - ~- moved the **Views** and **Events** pages under the **AutoOps for ECH** section and linked to them from the **AutoOps for SM clusters** section because they don't apply to **AutoOps for Serverless**. I'm open to feedback if anyone has a better idea for organizing this info.~

Closes https://github.com/elastic/docs-content/issues/3220


shainaraskas:
>Plus, it feels like we're reshuffling all the pages and potentially introducing more confusion just because we want the Views and Events pages to show up in both the ech and self-managed sections. 

I think ultimately it's still something that should do - nesting this info inside of one deployment type when it applies to many (especially because it's actually the most "active" content - stuff people who use autoops all the time need to refer to) is burying it. 

I think something like this that pulls the insights into buckets so serverless people aren't confused is a lighter touch, and keeps info about day to day usage front and center. I also think that explaining that serverless analyzes vcu vs hosted/ccm more about resource utilization is nice content for the autoops overview, and the guts of the billing dimension info could go into the serverless insights intro

>There is no setting up required for ECH and Serverless

sounds like we could go even lighter here then - this is just wayfinding - why do we have dedicated pages? your self-managed cluster access instructions are also buried pretty deeply for day-to-day usage.

here's kind of where i'm going based on looking at the content a second time

```
* autoops (overview + comparison info, with a setup section that explains no setup is needed for serverless/ech)
	* set up for ccm
	* insights (access instructions for all deployment types could go here)
		* ech/ccm views and events
		* serverless views (billing dimensions info goes here)
	* regions
	* faq
```

ultimately, I won't block your approach because this isn't really my content, and I realize I'm throwing a lot at you. however, I think designing the info hierarchy based on modes of usage and info people need at each stage would likely be a worthy effort. let me know if you want to talk synchronously about this or just leave it alone and let you do your thing.

deploy-manage/monitor/autoops/autoops-sm-troubleshoot-firewalls.md
```diff
@@ -0,0 +1,150 @@
...
+},
+"tagline" : "You Know, for Search"
+}
+```
+
+If you did not receive a similar response, your system returns an error indicating one or more reasons for failure as outlined in the following table. Use the corresponding proposed solution to fix the issue.
+
+| Reason for failure | Proposed solution |
+| :--- | :--- |
+| The {{es}} endpoint URL you specified is incorrect/not reachable | - Make sure you are using `http` or `https` correctly. <br> - Make sure you are providing the correct port. The default port is **9200**. <br> - If you have verified that the URL is correct, your network team might need to open the firewall to allow-list this URL. |
```
kilfoyle:
```suggestion
| The {{es}} endpoint URL you specified is incorrect/not reachable | - Make sure you are using the correct protocol in the cluster URL: `http` or `https`. <br> - Make sure you are providing the correct port. The default port is **9200**. <br> - If you have verified that the URL is correct, your network team might need to open the firewall to allow-list this URL. |
```

wajihaparvez:
Not at all! Thanks for your suggestions, David 🙏

PR #3957, opened 2025-11-17
Restructure contrib tools docs

theletterf:
Fixes https://github.com/elastic/docs-content-internal/issues/486

Also reorganizes the Tools section. "Tools" might be a bit dry as a name, so the renaming is open to suggestions.


leemthompo (approved):
LGTM ✨

PR #3966, opened 2025-11-17
Fix typos in Azure billing documentation

mmahacek:
Fix misspelled "acccount"


yetanothertw (approved):
Thanks for fixing that! 🚢

PR #3972, opened 2025-11-18
Update differences-from-other-elasticsearch-offerings.md

theletterf:
Updates and adds several EDOT features.

Fixes https://github.com/elastic/opentelemetry/issues/491


yetanothertw (approved):
Thanks for reviewing the issue and fixing it so quickly, Fabri! 🚀

PR #3984, opened 2025-11-18
Add password unset information

shainaraskas:
Explain the cause of a password unset in a situation where both social/password based login is set up

@alxchalkias is there a "why" we might want to provide here?


cloud-account/add-a-login-method.md
```diff
@@ -19,6 +19,10 @@ When you create your account, you can choose one of the the following methods to
 
 After your account is created, you and can an additional login method. You can’t remove a login method after it’s added.
 
+:::{caution}
```
shainaraskas:
```suggestion
:::{warning}
```

PR #3998, opened 2025-11-19
Update kibana-api-docs-quickstart.md

lcawl:
This PR updates the guidance around the use of the capture_oas_snapshot command, since omitting `include-path` means the command outputs all "public" Kibana APIs but that's not the same list we're currently publishing.


szabosteve (approved):
Thanks, LGTM!

Issue #2235, opened 2025-07-23
revisit usage of version.stack


shainaraskas:
inspired by https://github.com/elastic/docs-content/pull/2222

we are using `{{version.stack}}` in the same way we used it in the old versioned docs in some cases.

this doesn't make sense in the new system, because it implies that something is only true in the latest version, when it might have been true since the _base_ version

example: https://github.com/elastic/docs-content/pull/2222/files#diff-07f5858632ade331374d8096f51946affacd54bd741f04658a5288db4c879a13

most areas for concern I believe are in the dev docs zone but we probably all need to audit this ... might apply to other version variables as well

shainaraskas:
admin can verify that the work is complete here

georgewallace:
I have reviewed `deploy-manage/deploy/self-managed/tutorial-self-managed-install.md` and all instances of `{{version.stack}}` are valid as it is a tutorial having a person run through the install of Elasticsearch so latest should be fine. Plus we need to put version variables in the commands.

PR #4008, opened 2025-11-20
[E&A][REF] Removes version.stack attributes

szabosteve:
## Overview

Related to https://github.com/elastic/docs-content/issues/2235

This PR removes the `version-stack` attributes from the Explore and analyze and Reference sections as using those attributes could be misleading. Some of these attributes are used as part of a download bash command, which always points to the latest stack release. I left those occurrences intact.


florent-leborgne (approved):
Thanks! LGTM

leemthompo (approved):
✅

PR #4022, opened 2025-11-20
Add Details About username_rdn_name and username_rdn_oid to PKI Security Doc

ebarlas:
This change adds one section to `deploy-manage/users-roles/cluster-or-deployment-auth/pki.md` with details about a pair of new settings:

* `username_rdn_name` - RDN attribute name to use for principal extraction from client certificate, e.g. `CN`
* `username_rdn_oid` - RDN attribute OID to use for principal extraction from client certificate, e.g. `2.5.4.3` (CN)

These changes have been merged to ES `main` (9.3.0): https://github.com/elastic/elasticsearch/pull/137230

Back-porting is planned for ES 9.2.2, 9.1.8, and 8.19.8.


ebarlas:
> if you don't have a preference (or prefer the alternative), consider the following. the bullets and tabs should be ordered in order of recommendation.

Great suggestion! I'm applying it now.

deploy-manage/users-roles/cluster-or-deployment-auth/pki.md
```diff
@@ -44,24 +44,51 @@ To use PKI in {{es}}, you configure a PKI realm, enable client authentication on
...
-            pki:
-              pki1:
-                order: 1
-                username_pattern: "EMAILADDRESS=(.*?)(?:,|$)"
-    ```
-
-    ::::{note}
-    If the regex is too restrictive and does not match the Subject DN of the client’s certificate, then the realm does not authenticate the certificate.
-    ::::
+2. Optional: The username is defined by the [username_pattern](https://docs-v3-preview.elastic.dev/elastic/elasticsearch/tree/main/reference/elasticsearch/configuration-reference/security-settings#ref-pki-settings). If you want to use something other than the CN of the Subject DN as the username, you can use one of the following methods to extract the username:
```
shainaraskas: reading back over this, I think I might have missed a nuance. the description here

>The username is defined by the [username_pattern](https://docs-v3-preview.elastic.dev/elastic/elasticsearch/tree/main/reference/elasticsearch/configuration-reference/security-settings#ref-pki-settings). If you want to use something other than the CN of the Subject DN as the username, you can use one of the following methods to extract the username:

actually applies to the regex option only, not some sort of default method. providing some edits

```suggestion
Optional: If you want to use something other than the CN of the Subject DN as the username, you can use one of the following methods to extract the username:
```

deploy-manage/users-roles/cluster-or-deployment-auth/pki.md
```diff
@@ -44,24 +44,51 @@ To use PKI in {{es}}, you configure a PKI realm, enable client authentication on
...
+         realms:
+           pki:
+             pki1:
+               order: 1
+               username_rdn_name: "CN"
+   ```
+   ::::
+
+	::::{tab-item} Regex
+	Specify a regex to extract the desired username. The regex is applied on the Subject DN.
```
shainaraskas:
```suggestion
	Using the [[username_pattern](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#ref-pki-settings) setting, specify a regex to extract the desired username. The regex is applied on the Subject DN. The regex is applied on the Subject DN.
```

deploy-manage/users-roles/cluster-or-deployment-auth/pki.md
ebarlas: The `username_pattern` regular expression _is_ used by default. But I think your edit is an improvement. It looks like you removed the `2. ` leading text. Was that intended?

shainaraskas: no, it wasn't - that was a bad edit (fixed)

when it's used by default, does that mean that the person doesn't need to set the setting for it to be applied, but it doesn't appear in any config files?

under the hood, is it more like "we technically use something that uses the same technology as username_pattern but it's invisible to the user, and the user can override it by setting their own username using one of these two methods"? if that's the case I think my edit still makes sense because using username_pattern behind the scenes is kind of an implementation detail

shainaraskas (approved):
one small typo then good to go. thanks for adding this in 🙏

shainaraskas:
@ebarlas given that this is patching older releases, I suggest holding the PR until the patches go out. otherwise, people will think that this functionality is currently available. 

for what it's worth, the engineering policy is not to release features/enhancements in patches like this (unfortunately I don't have a written source for this policy on hand). our docs system accommodates features coming out in minors by subbing tags for future unreleased minor versions as `planned`, so we can release those whenever, but due to the complexity of handling features introduced in patches, we just have to hold PRs like this :)

PR #4068, opened 2025-11-25
Service targetPort guidance when customizing service's port

eedugon:
Added notes on changing service ports and potential disruptions.

Addresses https://github.com/elastic/cloud-on-k8s/issues/7198


deploy-manage/deploy/cloud-on-k8s/accessing-services.md
```diff
@@ -68,7 +68,9 @@ You can expose services in [different ways](https://kubernetes.io/docs/concepts/
...
 
 ::::{warning}
-When you change the `clusterIP` setting of the service, ECK will delete and re-create the service as `clusterIP` is an immutable field. Depending on your client implementation, this might result in a short disruption until the service DNS entries refresh to point to the new endpoints.
+When you change the `clusterIP` setting of the service, ECK deletes and re-creates the service, as `clusterIP` is an immutable field. Depending on your client implementation, this might result in a short disruption until the service DNS entries refresh to point to the new endpoints.
+
+If you change the service’s `port`, set `targetPort` to the container’s default listening port. Otherwise, Kubernetes uses the same value for both, resulting in failed connections. For reference, default ports are `9200` for the {{es}} HTTP interface, and `5601` for {{kib}}.
```
eedugon: We could also include a list of reference default ports, like:
* 9200 for ES HTTP
* 9300 for ES Transport
* 9443 for ES RCS (Remote cluster server)
* 5601 for Kibana

But I didn't want to make the note too big. Ports 9300 and 9443 are mentioned in their relevant documents.

deploy-manage/security/k8s-transport-settings.md
```diff
@@ -30,7 +30,9 @@ spec:
 Check the [Kubernetes Publishing Services (ServiceTypes)](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) that are currently available.
 
 ::::{note}
-When you change the `clusterIP` setting of the service, ECK deletes and re-creates the service, as `clusterIP` is an immutable field. This will cause a short network disruption, but in most cases it should not affect existing connections as the transport module uses long-lived TCP connections.
+When you change the `clusterIP` setting of the service, ECK deletes and re-creates the service, as `clusterIP` is an immutable field. This causes a short network disruption, but it should not affect existing connections as the transport module uses long-lived TCP connections.
+
+If you change the service’s `port`, set `targetPort` to `9300`, which is the default {{es}} transport interface port. Otherwise, Kubernetes uses the same value for both, resulting in failed connections.
```
shainaraskas: consider adding bullets here, or making this stacked notes. when I read the first paragraph of a note, I make the assumption that the second paragraph is related to the first. one of these methods will help people to understand that these are two notes that are not related

deploy-manage/deploy/cloud-on-k8s/accessing-services.md
shainaraskas: I like "for example" - this makes me wonder if we have some sort of central resource we can link to for all of these ports though 🤔

shainaraskas: same note about overloading the warning (use bullets or split the notes)

deploy-manage/security/k8s-transport-settings.md
eedugon: Interesting, @shainaraskas , thanks for the hint.
In this case both paragraphs are related in the way that it's a note related with customizing service fields. First paragraph warns the user about what happens if they change `clusterIP` and the second warns the user about implications of changing the `port`.

How would you present this instead? 2 separates notes don't feel necessary here. Maybe an intro sentence + 2 bullets?

deploy-manage/deploy/cloud-on-k8s/accessing-services.md
barkbay: > this makes me wonder if we have some sort of central resource we can link to for all of these ports though 🤔

No, unfortunately we don't have that kind of reference documentation, we can create a doc issue though, I think it would make sense to have it in https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/accessing-services

PR #4134, opened 2025-11-27
refine remote cluster connection modes refinement

eedugon:
Small refinement to clarify when to use sniff mode and when to use proxy mode and reduce ambiguity.


kilfoyle (approved):
LGTM! 🦕 
Very nice!

Issue #2930, opened 2025-09-12
[Website]: Deployment comparison missing ece and eck


shainaraskas:
### Type of issue

Missing information

### What documentation page is affected

https://www.elastic.co/docs/deploy-manage/deploy/deployment-comparison

### What happened?

not all deployment types are represented on this doc and they need to be

PR #4154, opened 2025-11-28
fix the deployment comparison page (add ECE, ECK)

shainaraskas:
## Summary
Fixes https://github.com/elastic/docs-content/issues/2930

It was driving me insane. Added separate columns where needed (there are some infra differences between the deployment types, e.g. autoscaling and snapshot management)

## Generative AI disclosure

- [ ] Yes  
- [x] No  

Although I didn't use AI to compile the info, I had AI tell me if there were any categories missing. We might consider adding these in the future:

- Cost/Pricing: Since the parent page mentions cost considerations, you might want a comparison row about pricing models
- Support and SLA: Different deployment types have different support models
- Upgrade process: How upgrades are handled differs significantly across deployment types
- API access: Availability of REST APIs, management APIs, etc.
- Networking/Connectivity: VPC peering, private link, IP filtering, etc.
- Backup/Recovery: Currently only "Snapshot management" - could expand on disaster recovery procedures
- Performance tuning: Control over JVM settings, index settings, etc.

doing a sweep of this would be hard, but we should consider adding to this page as we discover key points of comparison. this was just about correcting the things that were absolutely wrong. asking @eedugon for a gut check please (and anything you think absolutely should go on this page ASAP)

### DIFF
--- a/deploy-manage/deploy/deployment-comparison.md
+++ b/deploy-manage/deploy/deployment-comparison.md
@@ -47,8 +47,8 @@ For more details about feature availability in {{serverless-short}}, refer to []
 | Custom plugins and bundles | Available | Available | No |
 | [Self-managed connectors](elasticsearch://reference/search-connectors/self-managed-connectors.md) | Available | Limited | Limited |
 | [{{es}}-Hadoop integration](elasticsearch-hadoop://reference/index.md) | Available | Available | No |
-| [Cross cluster search (CCS)](/solutions/search/cross-cluster-search.md) | Available | Available | [Planned](/deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings.md#elasticsearch-differences-serverless-feature-planned) |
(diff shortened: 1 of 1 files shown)


shainaraskas:
> And the link why in (Alerts ([why?](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/4154/deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings#elasticsearch-differences-serverless-features-replaced))) doesn't feel helpful or intuitive at all. There's no real explanation in the destination.

Good catch - we [lost an old section](https://github.com/elastic/docs-content/commit/061f12530ebbe0ae970a231877af16497bbb9e8c#diff-645c71b472d290b11c8d77331631de31391beb0f9e303b70cbb53cd3bcd6e05fL123) that explained what features we had removed and why. patching it up now

shainaraskas:
>Available + through ecloud are not complementary IMO, maybe through ecloud UI only would look better, but I might be missing background.

I'm going to leave this one alone. it's the tip of the iceberg of the whole "serverless doesn't use elasticsearch auth realms" - I think saying that auth realms are only functional through the elastic cloud layer makes sense. it's complex enough that people will need to click through anyway to understand

PR #4244, opened 2025-12-09
Fleetserver connectivity requirments when using an elasticsearch remote output

mjmbischoff:
In addition to connectivity check the fleet server also creates apikeys using the service token. This is currently not documented well, moreover we currently lead users astray telling them to ignore the output health.

This PR only addresses the note, a more elaborate section should be added to guide the user.

## Generative AI disclosure

- [ ] Yes  
- [x] No


bmorelli25:
Thanks for the contribution! @vishaangelova can you please give this an editorial review.

mjmbischoff:
> I hope I understood this correctly, but let me know if not or if I’m missing something.

Yeah, I think that description is fine. I'm not too particular about the location. What does need some sort of update is the "In some cases, the remote {{es}} output used for {{agent}} data can be reached by the {{agent}}s but not by {{fleet-server}}. In those cases, you can ignore the resulting unhealthy state of the output and the associated `Unable to connect` error on the UI." part as it gives the illusion that there is a scenario where we can ignore the health, but authentication is always  required.

> Would you mind opening a docs issue with some more details about this?

That's actually quite tricky as I'm unsure who on the dev side of things is responsible for this part.

vishaangelova (approved):
I’ve updated the PR to include the suggested changes from the tech writer review ([commit](https://github.com/elastic/docs-content/pull/4244/commits/cf763be4d115d6c9723213b8fd31db6082e77711)), so LGTM.

PR #4295, opened 2025-12-11
Change vm.max_map_count default to 1048576 in All ES docs for version 8.16+

kunisen:
## Summary

This is to address https://github.com/elastic/docs-content-internal/issues/592.

I am raising this as the first PR to decide the wording with docs, and ES dev / PM team. Thanks!

## Generative AI disclosure

- [ ] Yes  
- [x] No  

### Before PR merge

[deploy-manage/deploy/self-managed/vm-max-map-count.md](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/4295/deploy-manage/deploy/self-managed/vm-max-map-count)

### After PR merge

https://www.elastic.co/docs/deploy-manage/deploy/self-managed/vm-max-map-count


deploy-manage/deploy/self-managed/vm-max-map-count.md
```diff
@@ -12,10 +12,14 @@ products:
 
 {{es}} uses a [`mmapfs`](elasticsearch://reference/elasticsearch/index-settings/store.md#mmapfs) directory by default to store its indices. The default operating system limits on mmap counts is likely to be too low, which may result in out of memory exceptions.
 
+:::{admonition} Verify `vm.max_map_count` Configuration
+If the Operating System's default `vm.max_map_count` value is `1048576` or higher, no configuration change is necessary. If the default value is lower than `1048576`, configure `vm.max_map_count` to `1048576`.
```
yetanothertw:
```suggestion
If the operating system's default `vm.max_map_count` value is `1048576` or higher, no configuration change is necessary. If the default value is lower than `1048576`, configure the `vm.max_map_count` parameter to `1048576`.
```

deploy-manage/deploy/self-managed/vm-max-map-count.md
```diff
@@ -12,10 +12,14 @@ products:
 
 {{es}} uses a [`mmapfs`](elasticsearch://reference/elasticsearch/index-settings/store.md#mmapfs) directory by default to store its indices. The default operating system limits on mmap counts is likely to be too low, which may result in out of memory exceptions.
```
lkts: We say here "The default operating system limits on mmap counts is likely to be too low" but that is no longer true. I think your addition below supersedes this statement and we can remove it.

lkts: Maybe we want to say "could be too low"?

shainaraskas:
is this something that should be checked on upgrade, too? that's what I gathered from a couple of the slack conversations

kunisen:
Thanks @shainaraskas 

> is this something that should be checked on upgrade, too? that's what I gathered from a couple of the slack conversations

Yes. We should check this value when ES upgrade to 8.16+. 

However, it's OS setting so I am unsure if we can let ES API ([migration APIs](https://www.elastic.co/docs/api/doc/elasticsearch/group/endpoint-migration)) or Kibana [upgrade assistant](https://www.elastic.co/docs/deploy-manage/upgrade/prepare-to-upgrade/upgrade-assistant) detect it and do the suggestion.

I think from docs perspective, we can add one section like "review your prerequisite settings", and this `vm.max_map_count` belongs to [Increase virtual memory](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/vm-max-map-count) which is also part of the [Important system configuration](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/important-system-configuration), which can be considered as "prerequisites". 

From support and doc perspective, I think we may want to have this as an essential section, to make sure we can cover the things we need. (Another essential section we'd think could be important, is a reference to known issue. We may want to add that to help customer skip some version or take some prevention steps before things become too late after they upgrade)

---

But it's not my area (cloud) so I'd like to defer to ES friends (@bytebilly @lkts) for insights on this.

PR #4312, opened 2025-12-12
Remove old action

theletterf:
With https://github.com/elastic/docs-builder/pull/2343 we no longer need separate Vale actions.


.github/workflows/docs-build.yml
```diff
@@ -9,6 +9,7 @@ on:
 jobs:
   preview:
     uses: elastic/docs-builder/.github/workflows/preview-build.yml@main
+    enable-vale-linting: true
     with:
```
reakaleek:
```suggestion
    with:
      enable-vale-linting: true
```

reakaleek (changes requested):

PR #4323, opened 2025-12-12
Fix the HNSW memory estimate

benwtrent:
The HNSW graph memory estimate is the same no matter the quantization level or element type. 

Additionally, to further drive it home, I put down a full estimation combining the vector values and the graph.


yetanothertw (approved):
Thank you for your contribution to the docs! 🪴 

LGTM, I've left a minor rewording suggestion at your discretion.

PR #4351, opened 2025-12-15
Clarify that data streams are not supported by Playground

ppf2:
Added note about playground not supporting data stream indices.


leemthompo (approved):
thanks @ppf2! ❤️

PR #4393, opened 2025-12-17
EIS: remove batch size limitation

maxjakob:
This limitation on the `_bulk` API does not exist at this moment.


maxjakob:
> Thanks @maxjakob! Just checking to be sure to be sure: we don't need to tag this with a specific version?

@leemthompo no, this applies to all versions. Thanks for checking

PR #4394, opened 2025-12-17
Add Overview content type

theletterf:
## Summary

This PR adds the Overview content type and its template.

Fixes https://github.com/elastic/docs-team/issues/112

## Generative AI disclosure

- [X] Yes  
- [ ] No  

Tool(s) and model(s) used: Claude Opus 4.5 in Cursor 2 (Agent mode).

### DIFF
--- a/explore-analyze/elastic-inference/eis.md
+++ b/explore-analyze/elastic-inference/eis.md
@@ -82,8 +82,6 @@ The service enforces rate limits on an ongoing basis. Exceeding a limit results 
 | ELSER                 | 6,000           | 6,000,000               | 600,000                 | Limits are applied to both requests per minute and tokens per minute, whichever limit is reached first.  |
 | `jina-embeddings-v3`  | 6,000           | 6,000,000               | 600,000                 | Limits are applied to both requests per minute and tokens per minute, whichever limit is reached first.  |
 
-*Previously named *Elastic Managed LLM* (9.0–9.2).
-
 ## Pricing
 
 All models on EIS incur a charge per million tokens. The pricing details are available on our [Pricing page](https://www.elastic.co/pricing/serverless-search).


contribute-docs/content-types/overviews.md
shainaraskas: wondering how we handle for overviews that are a group of concepts under one umbrella - for example, this [security overview](https://www.elastic.co/docs/deploy-manage/security), which explains security by level. we might give people some doorways into breaking up concepts into multiple H2s if needed for "compound" concepts. in these contexts, next steps don't make a lot of sense, because your next step depends on a selection that you make further up on the page based on what you've learned.

we might want to hint that overviews might be very broad or very shallow, focused more on wayfinding or more on explanation. for example, on the opposite side of the spectrum, we basically just have an intro section and then some links: https://www.elastic.co/docs/deploy-manage/production-guidance 

a limitations section might also be useful

theletterf: I think the Overview as presented stands well on its own — I would argue that it's a good type to have. If we feel the need to deviate, we would discuss — it might be that we've been cramming several overviews into one in the past.

shainaraskas: > Would it be useful to have way-finding overview pages (or even call them something like "landing pages") as a standalone type? Or maybe that can be handled in one page here too...

I think that we generally skew a little shallow in terms of explanation on our wayfinding pages, so I like presenting them as overviews. However, at least hinting that the sections can be shorter (maybe with examples of the two ends of the spectrum) feels valuable so people don't feel like they're doing something "wrong" if their pages are at the 10th or 110th percentile of completeness.

contribute-docs/content-types/overviews.md
```diff
@@ -0,0 +1,94 @@
...
+
+Readers of overview pages are typically in a learning state rather than a goal-oriented or problem-solving state. They are often first-time users who want to get up and running quickly. Having effective overview content builds trust and confidence in our products.
+
+Overviews serve several purposes:
+
+- Explain what something is and why it matters.
+- Help users navigate to the right content for their needs.
+- Clarify how components, features, or concepts relate to each other.
+- Help users choose between options or understand trade-offs.
+
```
KOTungseth: Also: Inform users how Elastic features and capabilities bring value to their organizations and make their lives easier/improve their workflows.

contribute-docs/content-types/overviews.md
```diff
@@ -0,0 +1,94 @@
...
+## Best practices
+
+When you create overview pages, follow these best practices:
+
+- **Focus on a single concept:** Each overview should be dedicated to one concept, feature, or topic. If you find yourself explaining a second concept in depth, create a separate overview and link to it.
+- **Lead with user value:** Start by explaining why the feature or concept matters to the user, not just what it does.
```
KOTungseth: Here it is. We need to get better at this very important piece.

contribute-docs/content-types/_snippets/templates/overview-template.md
```diff
@@ -0,0 +1,115 @@
...
+<!-- REQUIRED
+
+A concise, descriptive name for the feature or concept.
+
+Example: Inference processing
+-->
+
+<!-- REQUIRED
+
+Introduction
```
KOTungseth: How prescriptive do we want to be? 

I was taught that intros should contain 50 words or fewer in one to two sentences. Best practice is to keep the first sentence to about 25 words to provide just enough information so that users know whether to read on, and so that advanced users can get what they need from the intro and not need to read more. 

Also, never repeat the title.

KOTungseth (approved):
Really great work on our overview content type! I left only a few comments and suggestions. We really do need to get better at including the value prop on all docs pages.

contribute-docs/content-types/overviews.md
theletterf: Yes!

Issue #3993, opened 2025-11-19
[Website]: Provide a list of URLs a user has to open when using AutoOps for self managed


jakommo:
### Type of issue

Missing information

### What documentation page is affected

https://www.elastic.co/docs/deploy-manage/monitor/autoops/ec-autoops-faq#collected-metrics-and-data-in-autoops-for-self-managed-clusters

### What happened?

We already list the regions in " Where are metrics stored in AutoOps for self-managed clusters?" and we mention which ports and protocols are used in "How does AutoOps gather data from my cluster and ensure its security?" to send the AutoOps data to Elastic, but it's not obvious for the user which URLs they need to allow list in their firewall/proxy to allow the Agent to send data.

It would be good to merge or clone the two sections into one that explains which URLs and ports/protocols need to be allow listed in a firewall/proxy. 
We have some details about this in https://www.elastic.co/docs/deploy-manage/monitor/autoops/cc-cloud-connect-autoops-troubleshooting#troubleshoot-issues, but it's not very obvious to the user.

So basically use the table with regions we have in https://www.elastic.co/docs/deploy-manage/monitor/autoops/ec-autoops-faq#collected-metrics-and-data-in-autoops-for-self-managed-clusters and explain that one has to open the firewall/proxy for HTTPS/443 :
* api.elastic-cloud.com 
* https://otel-collector.auto-ops.${REGION}.${CSP}.cloud.elastic.co (where REGION and CSP have to be filled in according to the region used and details from the table )

arnonst:
@jakommo We have a dedicated page that talks about firewall issues https://www.elastic.co/docs/deploy-manage/monitor/autoops/autoops-sm-troubleshoot-firewalls
We can add the information there and reference and firewall settings in the [page](https://www.elastic.co/docs/deploy-manage/monitor/autoops/cc-connect-self-managed-to-autoops) under the [Configure agent](https://www.elastic.co/docs/deploy-manage/monitor/autoops/cc-connect-self-managed-to-autoops#configure-agent) section.
Would this work?

wajihaparvez:
Thanks @jakommo! The changes are live [here](https://www.elastic.co/docs/deploy-manage/monitor/autoops/cc-connect-self-managed-to-autoops#configure-agent) and [here](https://www.elastic.co/docs/deploy-manage/monitor/autoops/ec-autoops-faq#data-gathering).

jakommo:
Looks great, thank you!

PR #4416, opened 2025-12-18
Add note about allowlisting port and URLs for AutoOps self-managed

wajihaparvez:
## Summary
This PR adds a note to specify which port and URLs need to be allowlisted if an org is using firewalls that block Elastic Agent from sending data to Elastic Cloud.

Closes https://github.com/elastic/docs-content/issues/3993


shainaraskas (approved):
one small suggestion for ya

Issue #4340, opened 2025-12-15
Clarify context of disabling data collection


wajihaparvez:
https://www.elastic.co/docs/deploy-manage/monitor/autoops/autoops-disable-metrics-collection
Should offer this option before installation

wajihaparvez:
Waiting on details about required config changes. See linked PR.

PR #4434, opened 2025-12-19
Add disable metrics instructions to setup and FAQ

wajihaparvez:
## Summary
This PR links the disable data collection page to relevant spots in the docs.

Closes: https://github.com/elastic/docs-content/issues/4340

## Generative AI disclosure

- [ ] Yes  
- [x] No

### DIFF
--- a/deploy-manage/monitor/autoops/cc-connect-self-managed-to-autoops.md
+++ b/deploy-manage/monitor/autoops/cc-connect-self-managed-to-autoops.md
@@ -91,6 +91,10 @@ If you don’t have an existing {{ecloud}} account:
 
 ### Select installation method
 
+:::{tip}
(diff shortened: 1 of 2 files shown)


pickypg (approved):
LGTM. I don't think we need to duplicate this to the other page as that page can be exactly how to do it and you get taken there if you have a reason.

It does make me wonder a little if we should add a table of the supported metricsets and whether they are optional or not. Only tasks_management and the templates are optional as we wrote, but would a table be better?

wajihaparvez:
@pickypg we haven't explicitly mentioned anywhere that tasks_management and templates are the only optional ones. They just happen to be the ones in our example. We do have this note on the page though:
Screenshot 2025-12-19 at 1 33 52 PM

If you think we should add the table/make any other changes, I'm happy to do it

PR #4447, opened 2025-12-23
Revise vm.max_map_count recommendation in bootstrap checks

shainaraskas:
## Summary

Updated the recommended value for vm.max_map_count to 1048576 in the bootstrap check docs

## Generative AI disclosure

- [ ] Yes  
- [x] No


kunisen:
Thank you @shainaraskas for opening this one. I missed to find we also have one mention in bootstrap. 

But this is not true yet - I am asking in [this internal comment](https://elasticco.atlassian.net/browse/ES-13805?focusedCommentId=305152) to ES dev about whether we should change our bootstrap code to set it higher. 

Currently it's still `262144` in ES code, so technically it means our bootstrap value is not bumped yet. 

* 8.19: [elasticsearch/server/src/main/java/org/elasticsearch/bootstrap/BootstrapChecks.java at 8.19 · elastic/elasticsearch](https://github.com/elastic/elasticsearch/blob/8.19/server/src/main/java/org/elasticsearch/bootstrap/BootstrapChecks.java#L451) 
* 9.2: [elasticsearch/server/src/main/java/org/elasticsearch/bootstrap/BootstrapChecks.java at 9.2 · elastic/elasticsearch](https://github.com/elastic/elasticsearch/blob/9.2/server/src/main/java/org/elasticsearch/bootstrap/BootstrapChecks.java#L451)

We are waiting on @mark-vieira and other ES team to confirm they want to carry out this change too, and if so, then we will likely have the value reflected to latest 8.18, 8.19 (as IIUC both minors are under maintenance) and 9.x docs.

---

I don't have a good idea about what we should do to this PR, but maybe we could wait until ES dev friend comes back after holiday season and discuss together? WDYT?

shainaraskas:
@kunisen the change I made doesn't change the bootstrap check value. instead, it only states that the bootstrap check value is different from the suggested value, and leaves the bootstrap check value the same. let me know what you think.

kunisen:
Thank you @shainaraskas I am good with your change. Let's change it for now. 
When ES team makes the change (as they should) in a future version, we may need to revisit the wording. 
But let's make our current docs and 8.16~8.19 ones covered first, and take actions if needed, in the future.

Issue #4117, opened 2025-11-26
Clean up conflicting/duplicative applies tags


shainaraskas:
We have a bunch of pages that have both a `stack` tag and one or more `deployment` tags, neither of which specifying a version. because `stack` should be applied in contexts where a page applies to all deployment types, we need to audit these pages and their procedures to determine which tags should actually be there.

Internal spreadsheet of pages with issues: https://docs.google.com/spreadsheets/d/1XZsifW-RNltrmAulkeKuD_2WAnglbfjpJtnzvDSLuyM/edit?gid=0#gid=0 
## Action
* Review the pages. Adjust the tags to reflect whether a page applies to all deployment types, or specify which deployment types something applies to. might involve testing in different contexts.
	*  section level? 

e.g. [shard allocation awareness](https://www.elastic.co/docs/deploy-manage/distributed-architecture/shard-allocation-relocation-recovery/shard-allocation-awareness) does not apply to ECH deployments because the setting is on the ECH denylist. Does it apply to ECK/ECE?

**fixes track**
- [x] @colleenmcginnis comms and doc updates
- [x] @shainaraskas fix spreadsheet to share load
- [x] TBD execute fixes (likely largely admin docs team)
- [ ] @colleenmcginnis? validate that no others have crept in

**other tracks**
- [ ] section-level problems?
		- look for section-level tags where deployment / stack / product coexist 
		- @colleenmcginnis to gather data please
- [ ] do we want to slip in some linting for the short term? @colleenmcginnis to investigate

PR #4449, opened 2025-12-23
Applies tag cleanup pt 1 (mostly troubleshooting topics)

shainaraskas:
## Summary

part of https://github.com/elastic/docs-content/issues/4117

Fixes applies information on several troubleshooting pages. Did the following additional labor:

* Restructured some tabbed content: some ECH/self tab splits were actually Kibana/ES API splits. Others were ECE+ECH/ECK+self splits
* Added some hints around applicability for "shoulder" docs (e.g. discovery is much more automated in ECE/ECH/ECK, but the docs are still helpful)
* Fixed some linting stuff

Applicability was determined based on supporting documentation (e.g. availability of snapshot/restore plugins) and accessibility of the troubleshooting paths provided. Will get a double-check probably from Edu on my gut feelings

## Generative AI disclosure

- [x] Yes  
- [ ] No  

Confirmed assumptions about troubleshooting applicability with chatgpt

### DIFF
--- a/deploy-manage/deploy/elastic-cloud/restrictions-known-problems.md
+++ b/deploy-manage/deploy/elastic-cloud/restrictions-known-problems.md
@@ -27,6 +27,7 @@ When using {{ecloud}}, there are some limitations you should be aware of:
 * [Regions and Availability Zones](#ec-regions-and-availability-zone)
 * [Node count and size](#ec-node-count-size)
 * [Repository analysis API is unavailable in {{ecloud}}](#ec-repository-analyis-unavailable)
+* [Service status](#ec-service-status)
 
 For limitations related to logging and monitoring, check the [Restrictions and limitations](../../monitor/stack-monitoring/ece-ech-stack-monitoring.md#restrictions-monitoring) section of the logging and monitoring page.
 
@@ -151,3 +152,7 @@ To make a seamless migration, after restoring from a snapshot there are some add
 ## Repository analysis API is unavailable in {{ecloud}} [ec-repository-analyis-unavailable]
 
 * The {{es}} [Repository analysis API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-snapshot-repository-analyze) is not available in {{ecloud}} due to deployments defaulting to having [operator privileges](../../users-roles/cluster-or-deployment-auth/operator-privileges.md) enabled that prevent non-operator privileged users from using it along with a number of other APIs.
+
+## Service status [ec-service-status]
+
+* To ensure we can continue evolving our status page to best serve our customers, we cannot guarantee consistency of API implementation or component API identifiers. However, we communicate changes which might impact status page subscribers on a best-effort basis. Review [Service status](../../cloud-organization/service-status.md#service-status-support-limitations) for more guidance.
--- a/deploy-manage/users-roles/cluster-or-deployment-auth/pki.md
+++ b/deploy-manage/users-roles/cluster-or-deployment-auth/pki.md
@@ -4,7 +4,6 @@ mapped_pages:
 applies_to:
   deployment:
     self:
-    ece:
     eck:
 products:
   - id: elasticsearch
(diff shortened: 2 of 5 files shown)


troubleshoot/elasticsearch/decrease-disk-usage-data-node.md
```diff
@@ -4,18 +4,13 @@ mapped_pages:
...
-    ece:
-    self:
 products:
   - id: elasticsearch
 ---
 
 # Decrease the disk usage of data nodes [decrease-disk-usage-data-node]
 
-In order to decrease the disk usage in your cluster without losing any data, you can try reducing the replicas of indices.
+To decrease the disk usage in your cluster without losing any data, you can try reducing the replicas of indices.
```
eedugon: Unrelated question.... should we mention users (in the next note for example) about more permanent solutions, such as using searchable snapshots for example to reduce the disk usage? Searchable snapshot do not rely on replicas for high availability, hence the disk usage is much lower than standard indices. Of course they shouldn't be used for frequently accessed data, but that's explained in the searchable snapshots guide (I hope).

A second comment:
- should we also mention the obvious method to reduce disk usage which consist of deleting unused / unneeded data? I mention this because many times when a user has an out of disk situation, the quickest solution is to find unneeded data and delete it (for example old indices that are no longer needed). Definitely that's obvious but it's not mentioned in this guide at all.

Sorry for the off-topic :) feel free to ignore it!

troubleshoot/elasticsearch/diagnose-unassigned-shards.md
```diff
@@ -22,36 +17,17 @@ products:
 
 There are multiple reasons why shards might get unassigned, ranging from misconfigured allocation settings to lack of disk space.
 
-In order to diagnose the unassigned shards in your deployment use the following steps:
+To diagnose the unassigned shards in your deployment, use the following steps.
```
eedugon: Would it make sense to add a small description explaining what's an unassigned shard and what's the impact?

Just in case it helps I can add some background:

An unassigned shard is a shard that exists in the cluster metadata but is not currently allocated to any node, which means its data is unavailable for both search and indexing operations.

Unassigned shards can occur for many reasons, such as node failures, cluster or indices configuration, insufficient resources, or allocation rules that prevent Elasticsearch from placing the shard on any available node.

The presence of unassigned shards directly affects the cluster health status:
- If at least one replica shard is unassigned, the cluster health becomes yellow. The cluster can still serve all data, but redundancy is reduced.
- If at least one primary shard is unassigned, the cluster health becomes red. In this state, some data is unavailable, and affected indices cannot fully operate.

troubleshoot/elasticsearch/diagnosing-corrupted-repositories.md
```diff
@@ -4,47 +4,33 @@ mapped_pages:
...
 
-Multiple {{es}} deployments are writing to the same snapshot repository. {{es}} doesn’t support this configuration and only one cluster is allowed to write to the same repository. See [Repository contents](../../deploy-manage/tools/snapshot-and-restore.md#snapshot-repository-contents) for potential side-effects of corruption of the repository contents, which may not be resolved by the following guide. To remedy the situation mark the repository as read-only or remove it from all the other deployments, and re-add (recreate) the repository in the current deployment:
+Multiple {{es}} deployments are writing to the same snapshot repository. {{es}} doesn’t support this configuration and only one cluster is allowed to write to the same repository. Refer to [Repository contents](../../deploy-manage/tools/snapshot-and-restore.md#snapshot-repository-contents) for potential side-effects of corruption of the repository contents, which might not be resolved by the following guide. To remedy the situation mark the repository as read-only or remove it from all the other deployments, and re-add (recreate) the repository in the current deployment.
```
eedugon: Starting the content of this doc with the sentence `Multiple deployments are writing to the same snapshot repository` feels weird, but not sure how to correct it.

Considering the doc is called "Diagnose corrupted repositories", I would probably start saying that a possible cause for a corrupted repo is that multiple clusters or deployments are writing to the same repository, which is unsupported.

I'd recomment to find a way to improve reading flow, but I understand the original doc looked like this already.

eedugon (approved):
Looks great!

I shared some comments that you can ignore if you feel unneeded. More over the comments that suggest changes on the troubleshooting procedure itself which might not be the objective of this PR :) 

btw, are you planning to fix the applies tags of https://www.elastic.co/docs/troubleshoot/elasticsearch/troubleshooting-unstable-cluster and its siblings in a different PR?

I think those would also need to remove deployment types and keep stack only.

troubleshoot/elasticsearch/decrease-disk-usage-data-node.md
shainaraskas: added!

troubleshoot/elasticsearch/diagnose-unassigned-shards.md
shainaraskas: that's super helpful and I just stole most of it

Issue #4392, opened 2025-12-17
Elasticsearch PKI realm doc mislabeled with ECE


rseldner:
The `current`/merged PKI doc is incorrectly mislabeled with ECE.

PKI realms are not currently supported in ECE. 

https://www.elastic.co/docs/deploy-manage/users-roles/cluster-or-deployment-auth/pki

rseldner:
On a related note, the bundle references in the TIP boxes further down are not applicable as PKI is not useable on neither ECE nor ECH

https://github.com/elastic/docs-content/blob/f992ed6111e697a1608dc3ad52fafc7cb1181888/deploy-manage/users-roles/cluster-or-deployment-auth/pki.md?plain=1#L131

https://github.com/elastic/docs-content/blob/f992ed6111e697a1608dc3ad52fafc7cb1181888/deploy-manage/users-roles/cluster-or-deployment-auth/pki.md?plain=1#L178

PR #4459, opened 2025-12-29
Clarify PKI authentication limitations for deployments

eedugon:
Updated documentation to clarify that PKI authentication cannot be configured on both {{ech}} and {{ece}} deployments.

Closes https://github.com/elastic/docs-content/issues/4392

This PR assumes that PKI authentication is NOT supported on ECH and ECE deployments (@shainaraskas , should we confirm this first?).


yetanothertw (approved):

rseldner:
btw also discussed [here](https://elastic.slack.com/archives/CA6N3HDF1/p1765840527475749)

PR #4485, opened 2025-12-31
Update applies_to section in SSL handshake document

eedugon:
Removed deployment types from the applies_to section.


yetanothertw (approved):
👍

PR #4605, opened 2026-01-13
Update security-trb-roles.md applies_to tags

eedugon:
I'm leaving deployment.self only because the docs addresses the file realm and `elasticsearch-users` command that is only available in self-managed clusters.

It also mentions LDAP which is not available in ECH.

If we improve this doc in the future we could make it valid for `stack` and all deployment types.


yetanothertw (approved):
LGTM, thank you! 🚢

PR #4629, opened 2026-01-14
docs: update security url to link to central policy

ismisepaul:
This PR replaces security URLs with a link to the central security policy.


ismisepaul:
@yetanothertw sorry for the noise I thought I cleaned up the previous PRs. 

The link should be https://github.com/elastic/.github/blob/main/SECURITY.md as this will remain a constant.

PR #4703, opened 2026-01-20
Clarify resource allocation note in ECE documentation

eedugon:
Updated note on resource allocation to include disk.

I think the note is useful in any case, but we'll wait for @psanz-estc confirmation as he is working in a use case where a user has hot-expanded the ECE host disk (LVM volume) and the UI doesn't show the new size without a restart.


deploy-manage/deploy/cloud-enterprise/ece-manage-capacity.md
```diff
@@ -27,7 +27,7 @@ You should plan your deployment size based on the amount of data you ingest. Mem
 During installation, the allocator capacity defaults to 85% of the host physical memory, as the rest is reserved for ECE system services.
 
 ::::{note}
-ECE does not support hot-adding of resources to a running node.  When increasing CPU/memory allocated to a ECE node, a restart is needed to utilize the additional resources.
+ECE does not support hot-adding of resources to a running node.  When increasing CPU, memory, or disk allocated to an ECE node, a restart is needed to utilize the additional resources.
```
eedugon:
```suggestion
ECE does not support hot-adding of resources to a running node.  When increasing CPU, memory, or disk allocated to an ECE node, a restart is needed to use the additional resources.
```

kunisen:
> What if we move the note to the beginning of document at the end of the intro? (Just before the memory section) The note is generic about resources (it covered memory and CPU and we are just adding the word disk).

Your idea is much better indeed Edu!
I also had the concern you mentioned and yeah putting it on top would solve all the unbalance.

kilfoyle (approved):
LGTM! 🦘

kunisen:
Operational (ops) update: Add label. Please ignore this message.

PR #4784, opened 2026-01-23
Add Painless debugging section

kilfoyle:
## Summary
This replaces the Painless section that was [proposed for the Troubleshooting section of the docs](https://github.com/elastic/docs-content/pull/3649), to instead be a [Debugging Painless scripts](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/4784/explore-analyze/scripting/painless-debugging) section in the Explore & Analyze Painless docs.

To my mind, this has two advantages:

 - These Painless pages aren't true "troubleshooting" docs, and I think would not fit well into the Symptoms / Diagnosis / Resolution format that is now standard for that section. Debugging, here, is something that people need to be mindful of as they're composing their scripts, rather than a spontaneous error that needs diagnosis and recovery.
 - Embedding this content into the existing Explore and Analyze section means that we have our Painless info in two locations (narrative + reference), rather than split across three separate parts of the docs. Much better as a user experience, IMHO.

Preview: Please see [Debug Painless scripts in Elasticsearch](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/4784/explore-analyze/scripting/painless-debugging) and all sub-topics.

---

@yetanothertw I made changes as we talked about, except that I didn't move the "root cause" sections. As I read through I realized that these are specific to each example, so they wouldn't make sense where the description is. But, I have moved the "Notes" sections to be part of the descriptions at the top of each page (above the fold).

## Generative AI disclosure

- [ ] Yes  
- [x] No

### DIFF
--- a/deploy-manage/deploy/cloud-on-k8s/manage-compute-resources.md
+++ b/deploy-manage/deploy/cloud-on-k8s/manage-compute-resources.md
@@ -293,6 +293,8 @@ If `resources` is not defined in the specification of an object, then the operat
 | Elastic Agent | `400Mi` | `400Mi` |
 | Elastic Maps Server | `200Mi` | `200Mi` |
 | Logstash | `2Gi` | `2Gi` |
+| Elastic Package Registry | `1Gi` | `1Gi` |
+| AutoOps Agent | `400Mi` | `400Mi` |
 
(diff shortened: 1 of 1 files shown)


florent-leborgne (approved):
LGTM, just a few nits (or questions since I'm not really sure of how painless works 😅)

Thank you it looks really great to have this type of troubleshooting section and pages

PR #4804, opened 2026-01-26
Update JVM heap size recommendations applies_to

eedugon:
For your consideration @stefnestor , maybe this was missing in your previous and great PR.

I think in ECH and ECE, users don't really have ways to configure JVM heaps in unexpected ways, so the section where we recommend not exceeding 1/2 of RAM and link to the main doc with guidelines and recommendations should also be applicable to self managed and ECK users.


stefnestor (approved):
I did totally miss, thanks for catching! 🙏

PR #4881, opened 2026-01-29
Remove Author Terms and add Stack Version column from EIS model table config

szabosteve:
## Summary

This PR removes the "Author Terms" column from the models.csv as the Model Card column contains enough informatin which makes Author Terms redundant.

## Generative AI disclosure

- [ ] Yes  
- [x] No


seanhandley:
@jonathan-buttner @maxjakob FYI we should think about stack version for models in EIS as a compatibility issue rather than a release date proxy i.e. we're going to need to prevent some models from showing up on older stack versions.

For the purposes of documentation, I'd expect the models we see in the docs to be a hint to users that they need to upgrade stack version for some newer models - for example, multimodal chat support.

PR #4992, opened 2026-02-04
Mark deprecated S3 settings for removal

DaveCTurner:
Uses the new `applies_to` syntax to express that these settings are
slated for removal in the next major version.


eedugon (approved):
LGTM!

Issue #5002, opened 2026-02-04
Default limit for xpack.alerting.rules.maxScheduledPerMinute is defined as to 10000 although this page indicates the default is default is 32,000


maggieghamry:
### Type of issue

Inaccurate

### What documentation page is affected

https://www.elastic.co/docs/deploy-manage/production-guidance/kibana-alerting-production-considerations

### What happened?

- The default setting for `xpack.alerting.rules.maxScheduledPerMinute` is shown as `32,000` [in this page's paragraph](https://www.elastic.co/docs/deploy-manage/production-guidance/kibana-alerting-production-considerations#_rules_with_very_short_intervals): 

<img width="828" height="162" alt="Image" src="https://github.com/user-attachments/assets/d34376f4-f49b-4283-a302-e82d55863355" />

- whereas the _actual_ default is [defined](https://www.elastic.co/docs/reference/kibana/configuration-reference/alerting-settings#:~:text=xpack.alerting.rules.-,maxScheduledPerMinute,-Specifies%20the%20maximum) as `10000`:
The default limit for `xpack.alerting.rules.maxScheduledPerMinute` is `10000` (except for Serverless which has a default of `400`); _not_ `32000` as indicated by [this paragraph](https://www.elastic.co/docs/reference/kibana/configuration-reference/alerting-settings#:~:text=xpack.alerting.rules.-,maxScheduledPerMinute,-Specifies%20the%20maximum) 
<img width="470" height="123" alt="Image" src="https://github.com/user-attachments/assets/91627f86-44f8-484b-9cb7-19f30848cb44" />

### Additional info

The advice on this page indicates:
>Default value of this config is 32,000, you should change this if you want to run more than 32,000 rules per minute.

This results in many users doubling and results in ~3-6x the actual default limit, which can create unexpected instability. Are we able to identify the source of "_Default value of this config is 32,000_" and resolve to the correct value (with link to the page that shows the default definition)?

eedugon:
Per conversation with @ersin-erdal , it looks we updated the default on 8.16.0 with [this PR](https://github.com/elastic/kibana/pull/190057).

I'll make sure to fix new and legacy asciidoc docs to show the right value in the reference docs.

Thanks a lot Ersin for your help and Maggie for reporting this!

PR #5016, opened 2026-02-05
Fix formatting and clarify maxScheduledPerMinute setting

eedugon:
Removing thousands separator in the default value to avoid confusion.

Relates to https://github.com/elastic/docs-content/issues/5002


deploy-manage/production-guidance/kibana-alerting-production-considerations.md
```diff
@@ -68,7 +68,7 @@ There are several scenarios where running alerting rules and actions can start t
...
+Another related setting is `xpack.alerting.rules.maxScheduledPerMinute`, which limits the number of rules that can run per minute. For example if it’s set to `400`, you can have 400 rules with one minute check intervals or 2000 rules with 5 minute check intervals. You cannot create or edit a rule if its check interval would cause this setting to be exceeded. To stay within this limit, delete or disable some rules or update the check intervals so that your rules run less frequently. Default value of this config is `32000`, you should change this if you want to run more than 32000 rules per minute.
```
eedugon:
```suggestion
Another related setting is `xpack.alerting.rules.maxScheduledPerMinute`, which limits the number of rules that can run per minute. For example if it’s set to `400`, you can have 400 rules with one minute check intervals or 2,000 rules with 5 minute check intervals. You cannot create or edit a rule if its check interval would cause this setting to be exceeded. To stay within this limit, delete or disable some rules or update the check intervals so that your rules run less frequently. Default value of this config is `32000`, you should change this if you want to run more than 32,000 rules per minute.
```

Issue #4978, opened 2026-02-04
[Website]: docker compose multi node instructions not working


segevfiner:
### Type of issue

Inaccurate

### What documentation page is affected

https://www.elastic.co/docs/deploy-manage/deploy/self-managed/install-elasticsearch-docker-compose

### What happened?

Kibana doesn't seem to start. And the setup container health check looks weird. The container just exits at the if all succeeds so the health check with the depends_on doesn't seem like it can logically work...

eedugon:
Thanks for sharing this @segevfiner , and @georgewallace for your quick test!

We were able to reproduce the issue, or at least a similar issue where the 3 instances of the Elasticsearch cluster (es01, es02, es03 ) are up but Kibana is not, and raises continuous authentication errors with the `kibana_system` user.

The `setup` container (the container that set up the certificates and afterwards configures the kibana_system password) is stucked in an endless loop trying to set up kibana_system password, because the command is continuously failing. We can't see it in the logs, but it's returning this:

```
curl -s -X POST --cacert config/certs/ca/ca.crt -u "elastic:${ELASTIC_PASSWORD}" -H "Content-Type: application/json" https://es01:9200/_security/user/kibana_system/_password -d "{\"password\":\"${KIBANA_PASSWORD}\"}"
{"error":{"root_cause":[{"type":"validation_exception","reason":"Validation Failed: 1: passwords must be at least [6] characters long;"}],"type":"validation_exception","reason":"Validation Failed: 1: passwords must be at least [6] characters long;"},"status":400}
```

Was your password shorter than 6 characters?
I've tried again with a longer password (`1234567890`) and everything worked fine.

The problems I see are:

- Password needs to be longer than 6 characters, and that's not documented.

- The loops of the `setup` container don't provide any log and could be infinite if the associated command always fails (like in this case setting up the `kibana_system` password). We can forward this information to the product team.

- The `healthcheck` of the `setup` container considers only the certificates generation and not the `kibana_system` password setup, but that's needed because:
  - `es01/2/3` shouldn't start if the certificates are not ready (that's what the healthcheck and depends_on is for).
  -  The command to set up `kibana_system` password needs the cluster to be up, so there's mutual dependency between setup and the cluster (cluster nodes depend on setup, but the last command of setup depend on the cluster being up).

[...]

PR #5037, opened 2026-02-05
Revise password requirements in docker compose installation guide

eedugon:
Updated password requirements for `ELASTIC_PASSWORD` and `KIBANA_PASSWORD` to specify minimum length.

The requirement only applies to `KIBANA_PASSWORD` actually in case we prefer a better wording.

Note that if the password is shorter than 6 characters Kibana never gets fully operational and there's no way to see why, but...

> curl -s -X POST --cacert config/certs/ca/ca.crt -u "elastic:${ELASTIC_PASSWORD}" -H "Content-Type: application/json" https://es01:9200/_security/user/kibana_system/_password -d "{\"password\":\"${KIBANA_PASSWORD}\"}"
{"error":{"root_cause":[{"type":"validation_exception","reason":"Validation Failed: 1: passwords must be at least [6] characters long;"}],"type":"validation_exception","reason":"Validation Failed: 1: passwords must be at least [6] characters long;"},"status":400}

This was discovered while taking a look at https://github.com/elastic/docs-content/issues/4978


deploy-manage/deploy/self-managed/install-elasticsearch-docker-compose.md
```diff
@@ -30,7 +30,7 @@ Use Docker Compose to start a three-node {{es}} cluster with {{kib}}. Docker Com
 
 4. In the `.env` file, specify a password for the `ELASTIC_PASSWORD` and `KIBANA_PASSWORD` variables.
 
-    The passwords must be alphanumeric and can’t contain special characters, such as `!` or `@`. The bash script included in the `docker-compose.yml` file only works with alphanumeric characters. Example:
+    The passwords must be at least 6 characters long, alphanumeric, and can’t contain special characters, such as `!` or `@`. The bash script included in the `docker-compose.yml` file only works with alphanumeric characters. Example:
 
     ```txt
     # Password for the 'elastic' user (at least 6 characters)
```
shainaraskas: hmm this makes me think this is a secret requirement for elasticsearch too

if we can confirm this we should ship. if we confirm otherwise we should remove this comment / be precise

(or just ship ... min password req relatively low risk)

shainaraskas (approved):
approved with a comment

deploy-manage/deploy/self-managed/install-elasticsearch-docker-compose.md
shainaraskas: ok, so it sounds like both require it, it's just not going to _break_ on anything but kibana. so i'd keep it as you have it now.

PR #5149, opened 2026-02-16
Fix instructions on how to find Cloud IDs

szabosteve:
## Summary

Relates to https://github.com/elastic/docs-content-internal/issues/776

This PR updates the following pages to fix the instructions on how to find the Cloud ID:

* Using Cohere with Elasticsearch
* Text embedding and semantic search
* Named entity recognition

## Generative AI disclosure

- [ ] Yes  
- [x] No


florent-leborgne (approved):
LGTM but I'd like to call out 2 things (probably to treat separately):

- "{{es}} Cloud" doesn't sound like something we should use. Does this mean {{ech}}?
- "Cloud ID" may not be the one you need on serverless; I had a case where I uploaded data from a script using the Elasticsearch Endpoint; that should be clarified (or if both work, that should be clarified too, though in serverless the ES Endpoint is easier to find)

PR #5203, opened 2026-02-18
Update privileges required for mcp-server.md

joeafari:
## Summary

Based on https://www.elastic.co/docs/explore-analyze/ai-features/agent-builder/permissions#es-cluster-privileges it looks like `monitor_inference` is required in some scenario. End users might end up having failures related to privileges while using the provided examples from the docs. It would be good to have an example of the API call in one article without the need search various places.

## Generative AI disclosure

- [ ] Yes  
- [x] No


leemthompo (approved):
thanks!

leemthompo:
> Previously we had: "cluster": ["all"], - I'm not opposed to bringing this back if you'd prefer this rather than specifying the monitor_inference in the cluster, up to you. I do like the idea of these examples including the least privilege necessary so what you've added makes sense.

Could add an annotation footnote for this maybe?

PR #767, opened 2025-03-13
Adds applies tags to Security pages

natasha-moore-elastic:
Adds page-level 'applies' tags to Solutions > Security pages and the Reference > Security pages that didn't already have them.

Previews:

- [Solutions > Security](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/767/solutions/security)
- [Reference > Security](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/767/reference/security/)

PR #5293, opened 2026-02-25
Add note based on the KB promotion process

yetanothertw:
## Summary

Fixes [#767](https://github.com/elastic/docs-content-internal/issues/767)

The PR adds a note immediately after the three field descriptions and before the `Operator` step. This seems like a decision point which is why I placed it there. I know the extra admonition is not ideal, but didn't feel like it slotted into the one above it.

## Generative AI disclosure

- [ ] Yes  
- [x] No


florent-leborgne (commented):
Thanks for dealing with this issue! I'd like to suggest using simpler wording if that makes sense (with my suggestion or anything else you'd prefer)

PR #5316, opened 2026-02-27
rum: fix the exporter endpoint configuration for all signals

vigneshshanmugam:
+ URL configuration incorrectly recommends adding the path to all Log, Metrics and Trace exporter which already adds this based on the provided `OTEL_EXPORTER_OTLP_ENDPOINT`. This redundancy would incorrectly causes issues as the RUM agent would be exporting to `v1/logs/v1/logs` instead of correct path.

OTLP exporters already handle this - https://github.com/open-telemetry/opentelemetry-js/blob/dcbead54f6e5f451d2635a1264028897dc976ef7/experimental/packages/exporter-logs-otlp-http/src/platform/browser/OTLPLogExporter.ts#L27


theletterf (approved):
+CC @alexandra5000

PR #5325, opened 2026-03-02
Rework "Rolling upgrade considerations" text

DaveCTurner:
A "network malfunction" isn't really the only way we can get into these
problems, it can also be general cluster instability. This commit
reworks the section to be more accurate.


yetanothertw (approved):
Looks great, thanks for making these changes! 🪴

PR #5417, opened 2026-03-10
EIS: increase rerank rate limits

maxjakob:
We increased the EIS rate limits for rerank models.


szabosteve (approved):
Thank you for the update! LGTM!

PR #5602, opened 2026-03-24
[ECE 4.2] IPv6 on AWS tutorial, ECE IPv6 Support, installation instructions, load balancers and OS guides

eedugon:
**DO NOT MERGE UNTIL ECE 4.2 IS RELEASES**

## Summary

Adding new doc from https://github.com/elastic/cloud/pull/153120 into docs-content as a tutorial.

Original seed by @brendandw refined in [this commit](https://github.com/elastic/docs-content/pull/5602/commits/b62c3477dc8d41425d1c5cc1975f3ebf463b7331) (for your review):

This PR address now Proxy Protocol V2 and IPv6 at the following levels:

- "IPv6 support on ECE" --> New document describing the current supported architecture.
- "Set up IPv6 for ECE on AWS" --> New doc - End to end example tutorial to implement IPv6 on AWS environments.
- "Proxy Protocol V2" --> New document that describes how to enable Proxy protocol V2 on ECE proxies for deployments HTTP traffic (ports 9200 and 9243), through the new flags added at https://github.com/elastic/cloud/pull/153120

The following documents have also been refined to properly support IPv6 and PPv2 when applicable:
- ECE Installation guides
- OS preparation guides (RedHat, SuSE, Ubuntu).

## Generative AI disclosure
- [x] Yes  : cursor with GPT model & Claude
- [ ] No


brendandw (commented):
Thanks for the thorough work on this tutorial, Eduardo! I've reviewed the current version and have a few technical suggestions based on my end-to-end testing.

I'll add specific line comments below, but here's a summary:

1. **Health check configuration** - The NLB target group health check should use HTTPS with `/_health` rather than TCP for more accurate health status
2. **Podman network for egress** - A note clarifying the dependency on dual-stack network configuration (which I understand you plan to add to "Prepare your environment" docs)
3. **RHEL NetworkManager config** - For existing installations, RHEL 8/9 often needs explicit nmcli configuration for IPv6
4. **Container recreation note** - A brief explanation of why Support guidance is needed for the Container Sets API approach

I've also created a tracking issue for the KB article covering the Container Sets API PATCH workflow: https://github.com/elastic/cp-hosted-team/issues/4183

Happy to sync on Slack if easier to discuss any of these.

brendandw (commented):
Technical review based on end-to-end testing. See inline comments for specific suggestions.

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,546 @@
...
+
+# Test IPv6 connectivity
+ping6 -c 3 ipv6.google.com
+```
+
+## Step 3: Prepare the Host and Install ECE [step-install-ece]
+
+Prepare the host according to the official ECE documentation: [Prepare your environment](/deploy-manage/deploy/cloud-enterprise/prepare-environment.md).
+
+::::{note}
```
brendandw: **Suggestion: Use HTTPS health check with `/_health` endpoint**

TCP health checks only verify the port is open, not that the proxy service is functioning correctly. The ECE proxy exposes a dedicated `/_health` endpoint for health checking.

Suggested change:

```markdown
3. **Health checks**:
   - **Health check protocol**: HTTPS
   - **Health check path**: `/_health`
   - **Health check port**: Traffic port
   - **Healthy threshold**: 3
   - **Unhealthy threshold**: 3
   - **Interval**: 30 seconds
   - **Timeout**: 10 seconds
   - **Success codes**: `200`
```

Also suggest updating the "Verify Proxy NLB" curl commands below to use `/_health`:
```bash
curl -4 -k -s -o /dev/null -w "IPv4: %{http_code}\n" "https://${NLB_DNS}/_health"
curl -6 -k -s -o /dev/null -w "IPv6: %{http_code}\n" "https://${NLB_DNS}/_health"
```
Both should return `200`.

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,546 @@
...
+
+## Step 2: Set up EC2 instances [step-vm-launch]
+
+Create one or more RHEL instances to host ECE. Skip this step if ECE is already installed.
+
+::::{note}
+This tutorial uses a single-host deployment for simplicity and testing purposes. For production environments, you should deploy multiple hosts across availability zones to ensure high availability and resilience. Refer to [Identify the deployment scenario](/deploy-manage/deploy/cloud-enterprise/identify-deployment-scenario.md) for guidance on recommended architectures.
+::::
+
+1. **EC2** → **Launch Instance**
```
brendandw, line 144: **Suggestion: Add note about dual-stack network dependency for egress**

The architecture diagram shows a dual-stack Podman network (`ece-network`), but the creation steps aren't included here. I understand from your earlier comment that you plan to add IPv6/dual-stack support to the "Prepare your environment" docs.

For now, could we add a note here clarifying this dependency? Something like:

```markdown
::::{note}
If you require IPv6 egress (outbound container connections to IPv6 endpoints), you must configure a dual-stack Podman network before installing ECE. Refer to [Prepare your environment](/deploy-manage/deploy/cloud-enterprise/prepare-environment.md) for container runtime IPv6 configuration.
::::
```

This sets the expectation without duplicating content that belongs in the preparation docs.

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,546 @@
...
+   - Add an IPv6 CIDR with a unique suffix (e.g., `2a05:d018:xxx:xx20::/64`)
+5. **Associate with route table:**
+   - Go to **VPC → Route Tables** → select your route table
+   - **Subnet associations → Edit** → Add the new subnet
+   - Ensure the route table has routes for `0.0.0.0/0` and `::/0` to the Internet Gateway
+
+### Create the ECE coordinators target group
+
+1. Navigate to **EC2** → **Target Groups** → **Create target group**
+
```
brendandw, line 365: **Suggestion: Add RHEL NetworkManager configuration for existing installations**

For existing RHEL 8/9 hosts, assigning an IPv6 address in AWS doesn't automatically configure the OS to use it. NetworkManager often needs explicit configuration.

Suggest adding after "Reconfigure host network interfaces...":

```markdown
#### Configure NetworkManager for IPv6 (RHEL 8/9)

After assigning the IPv6 address in AWS, RHEL may not automatically configure it:

```bash
# Find your connection name
nmcli con show

# Configure IPv6 auto (SLAAC) - replace "System eth0" with your connection name
sudo nmcli con mod "System eth0" ipv6.method auto

# Restart the connection
sudo nmcli con down "System eth0" && sudo nmcli con up "System eth0"

# Verify
ip -6 addr show scope global
ping6 -c 3 ipv6.google.com
```

If `ping6` returns "Network unreachable", verify:
- Your AWS route table has a route for `::/0` pointing to the Internet Gateway
- NetworkManager IPv6 method is set to `auto`
```

File: deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md [hunk 4]
```diff
@@ -0,0 +1,546 @@
...
+   - **IP address type**: IPv4
+   - **VPC**: Select your dual-stack VPC
+
+3. **Health checks**:
+   - **Health check protocol**: HTTPS
+   - **Health check path**: `/`
+   - **Health check port**: Traffic port
+   - **Healthy threshold**: 2
+   - **Unhealthy threshold**: 2
+   - **Timeout**: 5 seconds
```
brendandw, line 380: **Suggestion: Add note explaining why container recreation is needed**

This helps users understand what the Support guidance will involve:

```markdown
::::{note}
After updating the proxy configuration via the Container Sets API, running containers retain their previous network state. For the configuration changes to take effect on containers with exposed ports (such as the proxy on port 9243), those containers must be recreated. The Support team can guide you through this process safely.
::::
```

I've created a tracking issue for a KB article with the detailed procedure: https://github.com/elastic/cp-hosted-team/issues/4183

eedugon:
Awesome, @brendandw ! Thanks a lot for all your comments, I'll process them soon

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
brendandw: **Note:** This was updated in my latest PR commit (elastic/cloud#153120, commit `b787b62246f`). Your version pulled from an earlier iteration where I had TCP health checks. The change from TCP to HTTPS with `/_health` came from my end-to-end testing - TCP health checks were passing even when the proxy service wasn't fully functional.

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,546 @@
...
+
+| Component | Requirement |
+|-----------|-------------|
+| **VPC** | An associated IPv6 CIDR block (Amazon-provided or BYOIP) |
+| **Subnets** | Dual-stack subnets (IPv4 + IPv6) in at least two Availability Zones (required for Application Load Balancers) |
+| **Internet Gateway** | Attached to the VPC with routes for both `0.0.0.0/0` and `::/0` |
+| **Route Tables** | IPv4 and IPv6 routes to the Internet Gateway |
+| **Security Groups** | Inbound rules allowing traffic on ports `443`, `9243`, and `12443` for both `0.0.0.0/0` and `::/0` |
+| **EC2 instances** | ECE hosts running in dual-stack subnets (required only for IPv6 egress) |
+
```
brendandw, line 524: **Suggestion: Add detailed EC2 IPv6 address assignment steps**

The AWS requirements table mentions EC2 instances need to be in dual-stack subnets, but existing instances don't automatically get IPv6 addresses. Suggest adding this section before "IPv6 ingress in existing IPv4 environments":

```markdown
### Assigning IPv6 addresses to existing EC2 instances

Existing EC2 instances do not automatically receive IPv6 addresses when you enable IPv6 on their subnet. You must assign them manually.

**For each ECE host:**

1. Go to **EC2 → Instances → Select your instance**
2. **Actions → Networking → Manage IP addresses**
3. Expand the network interface section
4. Under **IPv6 addresses**, click **Assign new IP address**
5. Click **Assign** to auto-assign an IPv6 address from your subnet's range
6. Click **Save**

Alternatively, enable auto-assignment for future instances:
- **VPC → Subnets → Select subnet → Actions → Edit subnet settings**
- Enable **Auto-assign IPv6 address**
```

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,546 @@
...
+1. Reconfigure host network interfaces so each ECE host has working dual-stack connectivity.
+2. Reconfigure Podman or Docker bridge/network settings to enable IPv6 for container traffic.
+
+Because these are host-level networking changes, the recommended approach is to reinstall or rebuild ECE hosts one by one, following the official host maintenance procedure: [Perform ECE hosts maintenance](/deploy-manage/maintenance/ece/perform-ece-hosts-maintenance.md).
+
+This rolling approach reduces platform risk and helps preserve service availability while introducing IPv6 egress support.
+
+After completing these changes, refer to [Step 6: Verify IPv6 egress from containers](#step-ipv6-egress) to validate outbound IPv6 connectivity from ECE workloads.
+
```
brendandw, line 552: **Suggestion: Add Podman network commands for IPv6 egress on existing installations**

The current text says to "reconfigure Podman or Docker bridge/network settings" but doesn't show how. Suggest adding concrete commands:

```markdown
#### Configure dual-stack Podman network for IPv6 egress

Create the dual-stack network and connect existing containers:

```bash
# Create the dual-stack network
sudo podman network create \
    --subnet 10.89.0.0/24 \
    --subnet fd00:10:89::/64 \
    --ipv6 \
    ece-network

# Set as default for future containers
sudo tee -a /etc/containers/containers.conf > /dev/null <<EOF

[network]
default_network = "ece-network"
EOF

# Connect existing containers to the new network (for IPv6 egress)
for container in $(sudo podman ps -q); do
    sudo podman network connect ece-network "$container" 2>/dev/null || true
done
```

::::{note}
The `podman network connect` command enables IPv6 **egress** (outbound connections) for existing containers. However, it does not update iptables rules for **ingress** (inbound connections). For IPv6 ingress, containers with exposed ports may need to be recreated.
::::

Verify a container has IPv6 connectivity:

```bash
sudo podman exec <container_id> curl -6 -s -o /dev/null -w "%{http_code}" https://ipv6.google.com
```
```

This could alternatively go in the "Prepare your environment" docs as you mentioned - let me know your preference.

File: deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md [hunk 7]
```diff
@@ -0,0 +1,546 @@
...
+## Related Documentation
+
+- [ECE Hardware requirements](/deploy-manage/deploy/cloud-enterprise/ece-hardware-prereq.md)
+- [Prepare your environment](/deploy-manage/deploy/cloud-enterprise/prepare-environment.md)
+- [Install ECE](/deploy-manage/deploy/cloud-enterprise/install.md)
+- [ECE installation procedures](/deploy-manage/deploy/cloud-enterprise/install-ece-procedures.md) (memory settings for different deployment sizes)
+- [ECE Load balancers](/deploy-manage/deploy/cloud-enterprise/ece-load-balancers.md)
+- [Networking prerequisites](/deploy-manage/deploy/cloud-enterprise/ece-networking-prereq.md)
+- [{{aws}} Elastic Load Balancing documentation](https://docs.aws.amazon.com/elasticloadbalancing/)
```
brendandw, line 723: **Suggestion: Add final verification checklist**

After all the configuration steps, it would help to have a verification checklist. Suggest adding before "Related Documentation":

```markdown
## Final verification

After completing the configuration for existing installations, verify IPv6 is working end-to-end:

1. **Check target health** in {{aws}} Console: **EC2 → Target Groups → Targets** - all registered targets should show "Healthy"

2. **Test NLB connectivity** from a client with IPv6:
   ```bash
   NLB_DNS="your-nlb-dns-name.elb.region.amazonaws.com"
   
   # Test IPv4
   curl -4 -k -s -o /dev/null -w "IPv4: %{http_code}\n" "https://${NLB_DNS}/_health"
   
   # Test IPv6
   curl -6 -k -s -o /dev/null -w "IPv6: %{http_code}\n" "https://${NLB_DNS}/_health"
   ```
   Both should return `200`.

3. **Verify client IP propagation** by checking proxy logs after making a request:
   ```bash
   sudo podman exec frc-proxies-proxyv2 tail -10 /app/logs/proxy.requests.log | grep client_ip
   ```
   The `client_ip` field should show your actual client IP (IPv4 or IPv6), not an internal IP like `10.89.0.1`.
```

brendandw (commented):
Additional suggestions for the 'Existing Installations' appendix based on my end-to-end testing. These are new sections/content to add.

brendandw:
One gap to flag: **Docker IPv6 configuration**.

The tutorial currently says Docker users should refer to "Prepare your environment" docs, but those docs don't currently include Docker IPv6/dual-stack configuration.

Alex Piggott raised this in my PR (elastic/cloud#153120) - for Docker, the `/etc/docker/daemon.json` needs IPv6 settings:

```json
{
    "ip6tables": true,
    "experimental": true,
    "ipv6": true,
    "fixed-cidr-v6": "fd3f:25ad:fef0::/64"
}
```

**Context:** The original requirement was for STIG-hardened servers, which are "99.99% of the time RHEL+Podman" (per https://github.com/elastic/cp-hosted-team/issues/3531#issuecomment-3821179387). That's why the tutorial focuses on Podman - Docker IPv6 wasn't in scope.

Options:
1. Leave as-is - tutorial focuses on Podman (the hardened server path), Docker users can refer to Docker docs
2. Add Docker config alongside Podman
3. Separate issue to update "Prepare your environment" docs with Docker IPv6

I'm leaning toward option 1 given the original scope, but wanted to flag it. Thoughts?

cc @AlexP-Elastic

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
eedugon: final verification steps look great. I'll also check how to ensure they don't feel repetitive considering in each section we have verifications steps focused on each individual goal.

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
eedugon: Adding it, great catch!

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
eedugon: @brendandw : 

> The current text says to "reconfigure Podman or Docker bridge/network settings" but doesn't show how.

It was clearly a "placeholder" so we can decide what to do :) Thanks for sharing that procedure!

One thing that we probably must do is to ensure our "OS host preparation" official documents include the necessary steps to prepare a dual-stack host (for both docker and podman based hosts), for new installations.

Would they be similar to the ones you have proposed or would a new / uninstalled host would be configured with a dual-stack default podman network instead of creating a custom `ece-network` network?

For the procedure you are suggesting it looks really nice and smooth, but I'd like you to double check some concerns and potential risks to be totally sure this is ok:

- With the provided steps we are tweaking podman networks in a way that probably a new system won't have, although maybe that's fine.

- We are also adding the containers to an extra net (`ece-network`), besides the original /default bridge network they alrady belong and are connected to, and future containers would belong to `ece-network` only probably. I don't know the implications of that in future ECE upgrades, existing stunnels and things like that.

- The text saying that "containers with exposed ports may need to be recreated." should be clarified a bit more, as for ingress we are actually suggesting to use the original ipv4 adress of the hosts, so I don't think ingress should have any impact except if the user wants ingress towards the ipv6 address to the host (i've added an extra sentence in the note for your consideration).

Of course if you are confident about this, including upgrades, maintenance and support of a production system after these changes then I'm totally ok with it, as the process looks really easy and smooth.

An alternative approach would be telling users to reinstall ECE hosts one by one, enabling IPv6 at host level (with the EC2 and networkmanager example) and then reconfiguring docker per the official doc (the one that we have to update anyway), but of course this would take much longer in a prod system than your proposal.

brendandw: Thanks Eduardo! We discussed this in more detail on Slack - the short version is:

 - You're right that `containers.conf` setting` ece-network` as default means new containers after upgrades will use it automatically
 - Your concern about mixed subnets (old containers on both networks, new ones only on `ece-network`) is valid - probably fine in practice since ECE uses host IPs for inter-container communication, but it does leave things inconsistent
 - We agreed to keep the procedure but add a note recommending reinstalling ECE hosts one by one via host
    maintenance docs as the safest approach for production

You mentioned you'd test an ECE upgrade on a 3-host lab after applying the procedure - that would be really valuable to confirm how containers end up. Let me know what you find!

deploy-manage/deploy/cloud-enterprise.md
```diff
@@ -60,6 +60,7 @@ The section covers the following tasks:
   - [Connect your applications to {{es}}](./cloud-enterprise/connect-elasticsearch.md)
 
 * Learn about [](./cloud-enterprise/tools-apis.md) that you can use with ECE
+* [End-to-end tutorials](/deploy-manage/deploy/cloud-enterprise/ece-tutorials.md): Follow scenario-based guides that combine infrastructure and ECE configuration from start to finish.
```
eedugon, line 63:
```suggestion
* [End-to-end tutorials](/deploy-manage/deploy/cloud-enterprise/ece-tutorials.md): Follow scenario-based guides that combine infrastructure and ECE configuration from start to finish.

```

deploy-manage/deploy/cloud-enterprise/configure-proxy-protocol.md
```diff
@@ -0,0 +1,81 @@
...
+
+:::{note}
+In ECE versions earlier than 4.2, Proxy Protocol v2 parsing on the {{es}} HTTP ports is disabled and the installation flags described on this page are not available. If you need the real client address on an earlier version, consider the following options:
+
+* Use [direct source IP preservation](./ece-load-balancers.md#ece-client-ip-preservation) instead, if your load balancer can forward connections without replacing the client source address.
+* [Upgrade to ECE 4.2 or later](/deploy-manage/upgrade/orchestrator/upgrade-cloud-enterprise.md), which is the recommended way to use Proxy Protocol v2.
+* Enable Proxy Protocol v2 on your current version by patching the proxy container set directly. Follow KB article XXXX and contact [Elastic Support](/troubleshoot/index.md#contact-us) for assistance with the procedure.
```
gigerdo, line 26: Still references 'KB article XXXX '

eedugon: yeah, sorry... work in progress :)

kunisen: @eedugon WDYT this:

```suggestion
* Enable Proxy Protocol v2 on your current version by patching the proxy container set directly. Follow [our KB article](https://ela.st/ece-configure-ppv2-via-containersets-api) and contact [Elastic Support](/troubleshoot/index.md#contact-us) for assistance with the procedure.
```

I used https://ela.st/ece-configure-ppv2-via-containersets-api to represent the KB public URL - https://support.elastic.co/knowledge/8ceb7ead. By doing so, we can avoid using a random URL and next time even if we do maintenance to the underlying KB, we could simply update the shortened URL to match it easily, and no need to worry about URL going 404.

One more thing, I said `Follow our KB article` instead of using article name directly. I think I'd want to avoid saying the KB article subject directly, because it maybe changed depending on the situation, etc, and it can cause confusion if the subject is changed later. 

Finally, another tech thing is, we don't have containerset API documented in public, so I'd also think we probably don't want to advertise the KB subject directly, because that's the way to configure PPv2 via a non-documented API. I have added a note into KB:

image

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
eedugon: verified!

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
eedugon: verified

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
eedugon: solved

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
eedugon: moved to the KB article.

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
eedugon: done!

eedugon:
@alxchalkias : Adding you as a reviewer here. This is the docs proposal for the new support of Proxy Protocol v2 and IPv6 that we are introducing in ECE 4.2.

As you can see this is a big PR that touches multiple documents, because PPv2 and IPv6 relates with multiple areas of the existing documentation. OS preparation guides (RHEL, SuSE, Ubuntu), installation, and load balancer docs are updated.

Besides that we have also created a complete end to end tutorial (the one started by Brendan in the cloud repo) to show how to configure IPv6 on AWS, and 2 generic documents about Proxy Protocol configuration and IPv6 support.

cc: @gigerdo , @erikjenks & @AlexP-Elastic

troubleshoot/deployments/cloud-enterprise/troubleshoot-ipv6-aws-integration.md
```diff
@@ -0,0 +1,170 @@
...
+| Connection reaches ECE but returns error | ECE: Proxy configuration or deployment issue | ECE proxy logs (`proxy.requests.error.log`) |
+| Client IP shows as internal IP (for example `10.89.0.1`) | {{aws}}: Proxy Protocol not enabled on target group | {{aws}} Console: Target group attributes |
+| IPv6 works but IPv4 does not, or IPv4 works but IPv6 does not | {{aws}}: NLB not dual-stack, or missing security group rules | {{aws}} Console: NLB IP address type, security group rules |
+| Container cannot reach IPv6 endpoints | ECE: Container not on dual-stack network | Check container network with `podman inspect` |
+
+## Proxy NLB issues
+
+| Issue | Solution |
+|-------|----------|
+| Health checks failing | Ensure `--proxy-protocol-lenient` was used during ECE install. Health checks must use TCP, not HTTP. |
```
alxchalkias: This contradicts the tutorial: `ece-ipv6-aws-setup.md` ("Create the ECE proxies target group") configures the health check as HTTPS with path `/_health`. Someone who follows the tutorial and then hits an issue will land here and be told their working setup is wrong.

Which one is correct when `--proxy-protocol-lenient` is set? If HTTPS with `/_health` works, I'd reword this row to focus on the lenient flag (health checks fail when the flag was omitted, because checks don't send Proxy Protocol headers). If TCP really is required, the tutorial needs the fix instead.

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,647 @@
...
+   - This creates a `/64` IPv6 subnet from your VPC's `/56` block
+
+3. **Subnets** → Select each subnet → **Actions** → **Edit subnet settings**
+   - Enable **Auto-assign IPv6 address**
+
+4. **Route Tables** → Select the route table for your subnets → **Edit routes**
+   - Add route: Destination `::/0` → Target: Your Internet Gateway
+
+5. **Security Groups** → Select or create a security group → **Edit inbound rules**
+   - Add the following rules for both IPv4 (`0.0.0.0/0`) and IPv6 (`::/0`). These rules cover basic external access for a single-node test environment. For production multi-node deployments, refer to [Networking prerequisites](./ece-networking-prereq.md) for the full list of required ports, including inter-node traffic.
```
alxchalkias: These rules open ports 9243 and 12443 to `0.0.0.0/0` and `::/0`, which bypasses the load balancers entirely, and people copy these tables into production as-is.

Can we scope those two rules to the load balancer security group or the admin's CIDR? If they're intentional for initial testing, a warning would help, something like: "the direct-access rules for 9243 and 12443 are for initial testing only; once the load balancers are verified, restrict these ports to the load balancer security group."

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,647 @@
...
+2. Select **Network Load Balancer**
+
+3. **Basic configuration**:
+    - **Name**: `ece-proxy-nlb`
+    - **Scheme**: Internet-facing
+    - **IP address type**: **Dualstack** (important - not IPv4)
+
+4. **Network mapping**:
+    - **VPC**: Select your dual-stack VPC
+    - **Availability Zones**: Select the same subnet where your ECE instance is running (the NLB must be in the same subnet as the target instance)
```
alxchalkias: Same-subnet isn't an AWS requirement: the NLB needs a subnet in the same availability zone as the target (same VPC), not the same subnet. As written this will confuse multi-node setups where proxies span AZs.

```suggestion
    - **Availability Zones**: Enable the availability zones where your ECE proxies run. The NLB needs a subnet in each AZ that contains a target instance.
```

alxchalkias (commented):
Thanks @eedugon, this is a big one but it reads well. The ingress/egress split is the right mental model and you kept it consistent across every page. The appendix for existing installations is the part I suspect support will lean on most, along with the ordering guidance (reinstall proxies before enabling PPv2 on the LB, and the warning that container-set patches get overwritten on upgrade).

I left inline comments. Three of them are accuracy issues worth fixing before merge: the health-check contradiction, the NLB subnet claim, and the security group scope. The rest is consistency and style, mostly as suggestions you can apply directly. There are also four open Vale suggestions from the style bot ("disabled", "may", "launch" twice) that are cheap to clear. Since we're blocked on the 4.2 release anyway, no rush.

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,647 @@
...
+:::::
+
+## Overview
+
+This guide covers two distinct IPv6 traffic flows, ingress and egress, which can be configured independently and have different infrastructure implications:
+
+- **IPv6 ingress (client connectivity)**: Enables IPv4 and IPv6 clients to access ECE endpoints through {{aws}} load balancers. When combined with [IP filtering rules](/deploy-manage/security/ip-filtering-ece.md), this also lets you apply network security policies to IPv6 client addresses. This includes:
+  - **Deployment traffic**: Client access to {{es}} and {{kib}} through a dual-stack Network Load Balancer (NLB) and [Proxy Protocol v2](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/edit-target-group-attributes.html#proxy-protocol) used for [client IP propagation](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/edit-target-group-attributes.html#client-ip-preservation).
+  - **Control plane traffic**: Optional access to the ECE Admin Console through a dual-stack Application Load Balancer (ALB).
```
alxchalkias: Terminology: `ece-ipv6-support.md` says "Cloud UI" throughout, while this tutorial and the troubleshooting page say "Admin Console". The docs standard is Cloud UI. Can we align, or at least introduce one as an alias on first use?

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,647 @@
...
+# Check IPv6 address is assigned
+ip -6 addr show scope global
+
+# Test IPv6 connectivity
+ping6 -c 3 ipv6.google.com
+```
+
+::::
+
+::::{step} Prepare the Host and Install ECE
```
alxchalkias: Sentence case per the style guide.

```suggestion
::::{step} Prepare the host and install ECE
```

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,647 @@
...
+1. Go to **VPC → Subnets → Create subnet**
+2. Configuration:
+   - **VPC**: Select your VPC
+   - **Subnet name**: for example, `ece-subnet-2`
+   - **Availability Zone**: Select a **different** AZ from your first subnet
+   - **IPv4 CIDR**: An unused range (for example, `10.0.32.0/24`)
+3. Click **Create subnet**
+4. **Add IPv6 to the new subnet:**
+   - Select the new subnet → **Actions → Edit IPv6 CIDRs**
+   - Add an IPv6 CIDR with a unique suffix (for example, `2a05:d018:xxx:xx20::/64`)
```
alxchalkias: The IPv6 CIDR examples switch style mid-document: the ingress step uses suffix form (`00`, `01`) while this appendix uses a realistic prefix. Worth picking one convention.

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,647 @@
...
+
+::::
+
+:::::
+
+## Troubleshooting
+
+For troubleshooting and verification commands specific to IPv6 integration on {{aws}}, refer to [Troubleshoot IPv6 integration for ECE on {{aws}}](/troubleshoot/deployments/cloud-enterprise/troubleshoot-ipv6-aws-integration.md).
+
+## Related Documentation
```
alxchalkias: Sentence case. (The Proxy Protocol page uses just "Related"; either works, but worth aligning the two pages.)

```suggestion
## Related documentation
```

deploy-manage/deploy/cloud-enterprise/configure-proxy-protocol.md
```diff
@@ -0,0 +1,104 @@
...
+
+The two most common combinations are:
+
+* **HTTP (L7) mode**: the load balancer sets the `X-Forwarded-For` header with the client source IP.
+* **TCP (L4) mode**: the load balancer sends [Proxy Protocol v2](https://www.haproxy.org/download/2.8/doc/proxy-protocol.txt) metadata, and the ECE proxies must be configured to parse it. This is also the only mechanism that can preserve an [IPv6 client address](./ece-ipv6-support.md).
+
+This page covers the TCP (L4) path: how to enable Proxy Protocol v2 parsing in the ECE proxies for deployments HTTP traffic, on ports `9200` and `9243`. This is the only Proxy Protocol configuration that ECE exposes. For the other ports, the client IP mechanism is fixed and only configured on the load balancer side, as described in [Port and mode configuration](./ece-load-balancers.md#ece-load-balancer-ports).
```
alxchalkias: Small one: "deployments HTTP traffic" here, but the "Enable" section below uses the (correct) "deployment HTTP traffic".

```suggestion
This page covers the TCP (L4) path: how to enable Proxy Protocol v2 parsing in the ECE proxies for deployment HTTP traffic, on ports `9200` and `9243`. This is the only Proxy Protocol configuration that ECE exposes. For the other ports, the client IP mechanism is fixed and only configured on the load balancer side, as described in [Port and mode configuration](./ece-load-balancers.md#ece-load-balancer-ports).
```

deploy-manage/deploy/cloud-enterprise/ece-ipv6-support.md
```diff
@@ -0,0 +1,74 @@
...
+---
+
+# IPv6 support on ECE [ece-ipv6-support]
+
+{{ece}} (ECE) can be integrated into dual-stack (IPv4 and IPv6) networks. IPv6 clients can reach your deployments and the orchestrator endpoints, and your deployments can establish outbound connections to IPv6 destinations.
+
+Ingress and egress are independent capabilities with different infrastructure requirements. You can enable either one, or both:
+
+* **[IPv6 ingress](#ece-ipv6-ingress)**: IPv4 and IPv6 clients reach your deployments and the Cloud UI. It is handled entirely in front of ECE, so the ECE hosts are not modified. This makes it available to existing IPv4 installations.
+* **[IPv6 egress](#ece-ipv6-egress)**: ECE services and deployments open outbound connections to IPv6 destinations. This traffic originates on the ECE hosts, so it requires dual-stack host and a dual-stack container network.
```
alxchalkias: Missing article: "a dual-stack host".

```suggestion
* **[IPv6 egress](#ece-ipv6-egress)**: ECE services and deployments open outbound connections to IPv6 destinations. This traffic originates on the ECE hosts, so it requires a dual-stack host and a dual-stack container network.
```

deploy-manage/deploy/cloud-enterprise/ece-networking-prereq.md
```diff
@@ -12,7 +12,11 @@ products:
 
 The first host you install ECE on initially requires the ports for all roles to be open, which includes the ports for the coordinator, allocator, director, and proxy roles. After you have brought up your initial ECE installation, only the ports for the roles that the initial host continues to hold need to remain open. Before installing a host, make sure that ports 20000, 21000, and 22000 are open for the installation script checks. Port 2375 will also be utilized on each host you install ECE on for internal Docker communication.
 
-For versions 2.4.0 and 2.4.1, IPv6 should remain enabled on any host with the Proxy role. In 2.4.2 and later, IPv6 can be disabled.
+:::::{note}
+The port requirements in this page apply independently of the IP family. In a dual-stack environment, the inbound ports listed here still receive IPv4 traffic, because client traffic reaches the ECE hosts over IPv4. Allow IPv6 in your firewalls, security controls, and routing policies for outbound traffic if you need IPv6 egress.
```
alxchalkias, line 16: Checking this is intentional: the old note ("For versions 2.4.0 and 2.4.1, IPv6 should remain enabled on any host with the Proxy role. In 2.4.2 and later, IPv6 can be disabled.") was removed entirely. Those versions are long EOL so it's probably right. Flagging in case any supported version still needs host-level IPv6 for the proxy role.

deploy-manage/deploy/cloud-enterprise/configure-host-ubuntu.md
```diff
@@ -362,3 +367,13 @@ For more information, refer to the [Docker daemon configuration overview](https:
...
 1. Repeat these steps on other hosts that you want to use with {{ece}} or follow the steps in the next section to start installing {{ece}}.
+
+## Optional: Enable dual-stack networking for IPv6 egress [ece-ubuntu-ipv6-egress]
+```{applies_to}
+deployment:
+  ece: ga 4.2
+```
+
+::::{include} /deploy-manage/deploy/_snippets/ece-docker-ipv6-daemon.md
+::::
```
gigerdo: The optional IPv6 section sits after the step that tells people to repeat the procedure or start installing ECE. RHEL does this before reboot; SUSE does it before verify. As written, a linear reader will skip dual-stack setup.

Move this section (and the snippet include) to **before** that last numbered step — ideally before the reboot so one reboot picks up `daemon.json` and `sysctl`. Ubuntu is also missing the “if you need IPv6 egress…” pointer that RHEL/SUSE have in Before you begin.

<small>*(AI-authored via Cursor 🤖)*</small>

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -0,0 +1,647 @@
...
+      --subnet 10.89.0.0/24 \
+      --subnet fd00:10:89::/64 \
+      --ipv6 \
+      ece-network
+
+    # Set as default for future containers
+    sudo tee -a /etc/containers/containers.conf > /dev/null <<'EOF'
+    [network]
+    default_network = "ece-network"
+    EOF
```
gigerdo: This `tee -a` creates a second `[network]` section. The new-host RHEL guide says to merge `default_network` into the existing `[network]` block (`network_backend="cni"` on RHEL 9 / Rocky 9) rather than duplicating the section.

Please use the same merge instructions here so existing installs don’t end up with an invalid or ambiguous `containers.conf`.

<small>*(AI-authored via Cursor 🤖)*</small>

troubleshoot/deployments/cloud-enterprise/troubleshoot-ipv6-aws-integration.md
gigerdo: This contradicts the tutorial, which configures the NLB target-group health check as HTTPS with path `/_health`. With `--proxy-protocol-lenient`, that HTTPS check is the intended setup: NLB health checks do not send Proxy Protocol headers, and lenient is what makes them succeed.

If HTTPS `/_health` is correct, reword this row around the lenient flag rather than telling people to switch to TCP. Same table also says the NLB must be in the same subnet as the instance (row below); it needs a subnet in each AZ that has a target.

<small>*(AI-authored via Cursor 🤖)*</small>

gigerdo (commented):
P1s from a product/docs pass against the 4.2 installer flags and the host-prep pages. Details are on the inline comments.

<small>*(AI-authored via Cursor 🤖)*</small>

deploy-manage/deploy/cloud-enterprise.md
```diff
@@ -60,6 +60,8 @@ The section covers the following tasks:
...
```
shainaraskas: this new section doesn't follow the rest of the IA so I'll remove it

deploy-manage/deploy/cloud-enterprise/ece-tutorials.md
shainaraskas, line 1: antipattern with only one child entry - to remove

deploy-manage/deploy/cloud-enterprise.md
shainaraskas:
```suggestion
```

deploy-manage/toc.yml
```diff
@@ -174,6 +176,9 @@ toc:
                   - file: deploy/cloud-enterprise/ece-integrations-server-api-example.md
                   - file: deploy/cloud-enterprise/switch-from-apm-to-integrations-server-payload.md
           - file: deploy/cloud-enterprise/tools-apis.md
+          - file: deploy/cloud-enterprise/ece-tutorials.md
+            children:
+              - file: deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```
shainaraskas:
```suggestion
```

deploy-manage/deploy/_snippets/ece-proxy-protocol-install-note.md
```diff
@@ -0,0 +1,5 @@
+:::{note}
+:applies_to: ece: ga 4.2
+
+If you are configuring Proxy Protocol v2 between the load balancer and the ECE proxies, add the `--proxy-protocol-version 2` and `--proxy-protocol-lenient` flags to the installation command on every host. Refer to [Load balancers](../cloud-enterprise/ece-load-balancers.md) and [Configure Proxy Protocol v2](../cloud-enterprise/configure-proxy-protocol.md) for details.
```
shainaraskas:
```suggestion
If you are configuring Proxy Protocol v2 between the load balancer and the ECE proxies, add the `--proxy-protocol-version 2` and `--proxy-protocol-lenient` flags to the installation command on every host. Refer to [](/deploy-manage/deploy/cloud-enterprise/ece-load-balancers.md) and [](/deploy-manage/deploy/cloud-enterprise/configure-proxy-protocol.md) for details.
```

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
shainaraskas: attempted a fix in https://github.com/elastic/docs-content/pull/5602/changes/9172dfb28bf1d82ad6731710d345608187599e3d

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
shainaraskas: fixed in https://github.com/elastic/docs-content/pull/5602/changes/1a5e3616046a07b16d5c967f79a2de3f16a6e3a3

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
shainaraskas: attempted fix https://github.com/elastic/docs-content/pull/5602/changes/974fd444670918647723bd40ed2e864f9e600bda

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
shainaraskas: fixed in https://github.com/elastic/docs-content/pull/5602/changes/9438ee99ac0d5724885128e5f1a4140be460d807

deploy-manage/deploy/cloud-enterprise/ece-networking-prereq.md
shainaraskas: intended

troubleshoot/deployments/cloud-enterprise/troubleshoot-ipv6-aws-integration.md
shainaraskas: fix attempted https://github.com/elastic/docs-content/pull/5602/changes/1a5e3616046a07b16d5c967f79a2de3f16a6e3a3#diff-e466e1e2f458d9ae32baff37b0e116d056e5794dbd0884a7a76917bc83278876

shainaraskas: https://github.com/elastic/docs-content/pull/5602/changes/1a5e3616046a07b16d5c967f79a2de3f16a6e3a3#diff-e466e1e2f458d9ae32baff37b0e116d056e5794dbd0884a7a76917bc83278876

deploy-manage/deploy/cloud-enterprise/configure-host-ubuntu.md
shainaraskas: fixed https://github.com/elastic/docs-content/pull/5602/changes/916fa0be2bec81713664c266c5562509c5aa221d

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
shainaraskas: fixed in https://github.com/elastic/docs-content/pull/5602/changes/0fa83b1527652dddcc59d5d3a9f643ce3ee396f6

shainaraskas (approved):
did a bunch of consistency edits on this one and addressed feedback

the tutorial for AWS has a high chance of going stale very soon - we'll need to keep an eye on it

shainaraskas (approved):
oops didn't mean to ask myself to re-review

gigerdo (approved):

PR #5753, opened 2026-04-01
Change replication target from remote to local cluster

insukcho:
Updated instructions for replicating data to a local cluster instead of a remote cluster.

## Summary

Remote cluster is leader side that the local cluster pull data. As the limitation states that it will not copy the limit items like ILM policy, we should manually create the ILM policy in local side from the remote side ILM policy.

## Generative AI disclosure

- [ ] Yes  
- [x] No


kilfoyle (approved):
LGTM! 🏎️ 
Thanks @insukcho!

shainaraskas (commented):
I worry a little that this will not be as understandable to the user. the task the user is trying to do is replicate data to a remote cluster (remote from where the data resides). where the config happens is another concern. I think it would be clearer if we said that the config needs to happen on the remote cluster directly (and maybe say why), rather than swapping `local` and `remote` terms completely

shainaraskas (approved):

PR #5876, opened 2026-04-13
Ece remote cluster filter new UI

eedugon:
## Summary

ECE now includes "environment ID" in the remote cluster filters UI, so we have to adapt the docs where we included a "workaround" due to the bug in the UI (the UI was requesting an ORG-ID, which doesn't exist in ECE, only in ECH).

cc: @alxchalkias 

Closes https://github.com/elastic/docs-content-internal/issues/515

## Generative AI disclosure

- [x] Yes  
- [ ] No  
Tool(s) and model(s) used: Cursor with Auto mode.


deploy-manage/security/remote-cluster-filtering.md
```diff
@@ -27,11 +27,7 @@ Refer to [Remote clusters and network security](/deploy-manage/remote-clusters.m
 
 ## How remote cluster filters work
 
-Remote cluster filters operate at the proxy level, allowing incoming connections based on the organization ID or {{es}} cluster ID of the local cluster that initiates the connection to the remote cluster service endpoint (default port `9443`).
-
-::::{note}
-In {{ece}}, the equivalent of the organization ID in {{ech}} is the **environment ID**, which serves the same purpose for remote cluster filtering.
```
shainaraskas, line 33: important to keep this note for people with older versions of ECE?

eedugon: not sure if it makes sense to keep it in the intro.... maybe a note in the ECE procedure saying that versions older than ECE 4.1 might show Organization ID instead of Environment ID and they are equivalent?

I thought about this, and considering showing OrgID was a bug in the ECE UI and chances of causing trouble are low I thought it was better to remove the note and not even documenting this as a "change" (with applies_to or anything like that).

But I'm opened to any suggestion.

Maybe just keeping the note in the intro is the right call to at least understand better if someone finds the bug.

shainaraskas: I like the idea of putting this in the procedure as a note 👍

shainaraskas (approved):

shainaraskas:
apart from comment (to take or leave), good to go

deploy-manage/security/remote-cluster-filtering.md
eedugon: done!

Issue #5525, opened 2026-03-16
[Website]: Instruct how to customize the retention of the EDOT/OTel data streams on traditional and serverless deployments


lucabelluccini:
### Type of issue

None

### What documentation page is affected

https://www.elastic.co/docs/reference/opentelemetry/data-streams

### What happened?

We're lacking info on:
- What is the default retention period for the EDOT/OTel data streams?
- How to customize the ILM / Data Stream Lifecycle policies (e.g. what component template should I use)?

We document a part of it at https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/tutorial-customize-built-in-policies but it's a generic topic from platform. Also it focuses on ILM while on Serverless we have DSL (https://www.elastic.co/docs/manage-data/lifecycle/data-stream).

PR #768, opened 2025-03-13
Mention more built-in ilm policies

felixbarny:

PR #5891, opened 2026-04-14
Remove non-existent OTel-specific ILM policy names

alexandra5000:
## Summary

- Removes logs-otel@lifecycle, metrics-otel@lifecycle, and traces-otel@lifecycle from the list of built-in ILM policies in the customize tutorial
- These policies do not exist in Elasticsearch. OTel data streams use the standard logs@lifecycle, metrics@lifecycle, and traces@lifecycle policies (confirmed with @felixbarny)

Fixes inaccurate content introduced in #768. Related to elastic/docs-content#5525.


felixbarny (approved):

kilfoyle (approved):
LGTM! 🦘 
Thanks @alexandra5000!

PR #5928, opened 2026-04-16
Consider multiple quantization schemes for DiskBBQ memory estimate

tteofili:
Take into account different quantization schemes when estimating memory required for DiskBBQ.


deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
```diff
@@ -91,6 +91,10 @@ estimated\ bytes &= (1,000,000 \times 4 \times 16) + (1,000,000 \times 4 \times
 
 If you're using DiskBBQ, a fraction of the clusters and centroids need to be in memory.  When doing this estimation, it makes more sense to include both the index structure and the quantized vectors together as the structures are dependent. To estimate the total bytes, first compute the number of clusters, then compute the cost of the centroids plus the cost of the quantized vectors within the clusters to get the total estimated bytes.  The default value for the number of `vectors_per_cluster` is `384`.
 
+The cost of the quantized vectors also depends on the number of bits used to quantize indexed vectors (document vectors) and the number of bits used to quantize query vectors during search. Since 9.4, you can set the following quantization schemes: 1–4 (default), 2–4, 4–4 and 7–7 bits for document and query vectors, respectively.
```
marciw, line 94: Hmm, it seems like the `9.4` additions might make this page inaccurate for previous versions, which doesn't fit our [cumulative approach to docs](https://www.elastic.co/docs/contribute-docs/how-to/cumulative-docs#reader-experience). 

Can you clarify what's specific to 9.4 and what's not, and add [applies_to badges](https://www.elastic.co/docs/contribute-docs/how-to/cumulative-docs/badge-placement) accordingly? (instead of saying `Since 9.4`.) Maybe also consider using an [admonition](https://elastic.github.io/docs-builder/syntax/admonitions/) for the 9.4-specific stuff, and make the rest of the page version-agnostic.

Let me know what you think!

tteofili: thanks @marciw, I'll fix as per your suggestions

PR #6004, opened 2026-04-20
docs: note that documents table is collapsed by default in metrics Discover (#5845)

florent-leborgne:
## Summary

This PR addresses #5845 with the following changes:

- **`solutions/observability/infra-and-hosts/discover-metrics.md`**: Added a sentence to the intro paragraph noting that the documents table is collapsed by default when the metrics profile is active, and that users can expand it at any time. This reflects the behavior introduced in elastic/kibana#260607, which sets `hideTable: true` as the default app state for the metrics data source profile.

Note: I didn't add applies_to because the page already has the right values, so no need to repeat inline

## Resolves

Closes #5845

---

> **AI-generated draft** — created with Claude Sonnet 4.6.
> Review all generated content for factual accuracy before merging.


florent-leborgne:
I assigned this task to AI, reviewed it, and made any required changes. 

✅  **Ready for formal review.**

elky says go!

benironside (approved):

PR #6032, opened 2026-04-21
Clarify how net.ipv4.tcp_retries2 setting is related with node-left

### LINKED ISSUE #2439: [Website]: Clarify how net.ipv4.tcp_retries2 setting is related with node-left on the <Troubleshoot an unstable cluster> document
### Type of issue

Missing information

### What documentation page is affected

https://www.elastic.co/docs/troubleshoot/elasticsearch/troubleshooting-unstable-cluster

### What happened?

Add clarification on how net.ipv4.tcp_retries2 setting is related with node-left on the <Troubleshoot an unstable cluster> document. Refer this discussion:
https://elastic.slack.com/archives/C045SG0FRGW/p1754056950088939

### Additional info

_No response_

yetanothertw:
Fixes #2439

## Summary

If you lower `net.ipv4.tcp_retries2` (for example to 5), Linux gives up sooner when the network is truly bad; often on the order of seconds to tens of seconds, not many minutes. Then the broken connection is more likely to show up as disconnected, which is often a clearer clue that the network/TCP path is the problem.

This is what we currently say in the docs:
image

I've just expanded on why `net.ipv4.tcp_retries2` needs to be configured to a lower value, (rather than just saying it needs to be configured properly XD)

## Generative AI disclosure

- [x] Yes  
- [ ] No  

I used Claude's Auto agent to chat to it about this setting.


troubleshoot/elasticsearch/troubleshooting-unstable-cluster.md
```diff
@@ -84,7 +95,7 @@ To determine whether the node which left the cluster with the `disconnected` rea
...
 
-Lagging is typically caused by performance issues on the removed node. However, a node may also lag due to severe network delays. To rule out network delays, ensure that `net.ipv4.tcp_retries2` is [configured properly](../../deploy-manage/deploy/self-managed/system-config-tcpretries.md). Log messages that contain `warn threshold` may provide more information about the root cause.
+Lagging is typically caused by performance issues on the removed node. However, a node may also lag due to severe network delays. Refer to [Ruling out network delays with `net.ipv4.tcp_retries2`](#troubleshooting-unstable-cluster-tcp-retries-node-left), to eliminate network delays as a possible cause for this kind of instability.. Log messages that contain `warn threshold` may provide more information about the root cause.
```
marciw:
```suggestion
Lagging is typically caused by performance issues on the removed node. However, a node may also lag due to severe network delays. Refer to [Ruling out network delays with `net.ipv4.tcp_retries2`](#troubleshooting-unstable-cluster-tcp-retries-node-left), to eliminate network delays as a possible cause for this kind of instability. Log messages that contain `warn threshold` may provide more information about the root cause.
```

troubleshoot/elasticsearch/troubleshooting-unstable-cluster.md
```diff
@@ -66,9 +66,20 @@ The [Health](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operati
...
 * `followers check retry count exceeded`: The master sent a number of consecutive health checks to the removed node. These checks were rejected or timed out. By default, each health check times out after 10 seconds and {{es}} removes the node removed after three consecutively failed health checks. Refer to [Discovery and cluster formation settings](elasticsearch://reference/elasticsearch/configuration-reference/discovery-cluster-formation-settings.md) for information about the settings which control this mechanism.
 
+## Ruling out network delays with `net.ipv4.tcp_retries2` on Linux [troubleshooting-unstable-cluster-tcp-retries-node-left]
```
marciw: Maybe instead of putting the explicit setting name in the heading, there could be a phrase like "with retransmission timeouts" or maybe even "with TCP timeouts" ? Just for readability, especially in the OTP.

*Unless* we think users might scan the page, looking for `net.ipv4.tcp_retries2` specifically, in which case ignore me 🙃

marciw (approved):
LGTM with one typo and optional edit

troubleshoot/elasticsearch/troubleshooting-unstable-cluster.md
yetanothertw: Thanks, Marci! That makes sense.

I've updated it to "with TCP timeouts", and we use the actual setting name throughout the text so anyone who does a quick ctrl+f to scan for keywords can find it.

PR #6054, opened 2026-04-22
Clarify installation restrictions for Elastic Agents

fdartayre:
Running multiple containerized agents on the same host is a supported use case. This has been confirmed by the product team multiple times.


theletterf (approved):
Merci!

PR #6063, opened 2026-04-22
[APM] Clarify that services require transactions or spans to appear in inventory

### LINKED ISSUE #5857: [APM] Service Inventory: document that transactions/spans are required for services to appear
The [Services inventory page](https://www.elastic.co/docs/solutions/observability/apm/services) doesn't currently mention that a service must be sending transactions or spans in order to appear in the inventory. Services that only send logs will not be listed.

We need to add a note or callout near the top of the Services inventory page clarifying that:

- A service only appears in the Service Inventory if it has transactions or spans ingested within the selected time window
- Sending logs alone is not sufficient for a service to be listed

This is a source of confusion, particularly for users upgrading Stack versions or those with partially instrumented services. Without this information, the missing service appears to be a bug rather than expected behavior.

alexandra5000:
## Summary

- Adds a note near the top of the Services inventory page explaining that a service only appears in the inventory if it has sent transactions or spans within the selected time window
- Clarifies that sending logs alone is not sufficient for a service to be listed

Closes #5857


alaudazzi (approved):
LGTM

PR #1906, opened 2025-06-24
Expand "Configure a lifecycle policy" docs

kilfoyle:
This PR takes a chunk out of internal issue [Requested fixes for data lifecycle docs](https://github.com/elastic/docs-content-internal/issues/4) that was initiated by a review of the ILM docs by our Support team.

It updates the "Configure a lifecycle policy" page to:

 - Add Kibana steps where we currently show only the API steps.
     - In particular, when creating an index template users aren't always sure what to specify on the "Index settings" tab, so this adds an example config.
 - Add a page overview
 - Add a section about viewing the ILM status for an index or datastream
 - Fix up smaller items, such as:
    -  Explicitly call out the Kibana "Data retention" option.
    - Emphasize that lifecycle phase changes are based on time since rollover rather than index creation time
    - Warn about updating the `logs@lifecycle` and `metrics@lifecycle` policies since they affect a LOT of indices. 
 - Provide links for things like the index lifecycle actions, mappings, etc., to help people understand these options.
 
**Please see preview pages**: 
 - [Configure a lifecycle policy](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1906/manage-data/lifecycle/index-lifecycle-management/configure-lifecycle-policy)
 - [Apply a lifecycle policy to an index](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1906/manage-data/lifecycle/index-lifecycle-management/policy-apply)
 - [View the lifecycle status of an index](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1906/manage-data/lifecycle/index-lifecycle-management/policy-view-status)
 - [Update or switch a lifecycle policy](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1906/manage-data/lifecycle/index-lifecycle-management/policy-updates)

(For reference, here's the [original version](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/configure-lifecycle-policy) of the first page).

**ES Data Management team**, if any of you can please give this a technical review I'd be very grateful! 🙏  The API instructions aren't changed, with the exception that I added [this section](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/1906/manage-data/lifecycle/index-lifecycle-management/configure-lifecycle-policy#view-lifecycle-status) about calling the _ILM explain_ API. The Kibana steps are all new.

Issue #2311, opened 2025-07-29
[Website]: Include "Add a data tier" and "Disable a Data Tier" in sidebar links


rseldner:
### Type of issue

Other

### What documentation page is affected

https://www.elastic.co/docs/manage-data/lifecycle/data-tiers

### What happened?

We used to have a dedicated `Disable an Elasticsearch` data tier page in our EC docs that was very easy to locate
|**ECH**: https://www.elastic.co/guide/en/cloud/current/ec-disable-data-tier.html (this redirects now)|<img width="311" height="307" alt="Image" src="https://github.com/user-attachments/assets/fcafb5ad-3f02-4a41-9c91-adf3711f445f" />|
|---|---|
|**ECE**: https://www.elastic.co/guide/en/cloud-enterprise/3.8/ece-disable-data-tier.html|<img width="336" height="329" alt="Image" src="https://github.com/user-attachments/assets/68917be6-ffce-4699-89b1-1eae05035b9c" />|

This page has been merged into this general data tiers doc and it's sort of buried in there in between the ECH/ECE and self-managed deployments sections.  It is not obvious where in the overall page it is located.
https://www.elastic.co/docs/manage-data/lifecycle/data-tiers

So I wonder if the headings should be tweaked so the Disable a Data Tier section appears in the sidebar.

<img width="1153" height="587" alt="Image" src="https://github.com/user-attachments/assets/be42b2f3-93a5-450f-a6e1-541d3b655c4c" />

### Additional info

Opinion: The organization of the merged docs has made things more difficult to locate.

PR #6092, opened 2026-04-23
Implement feedback on data tiers

### LINKED ISSUE #1956: Feedback on "Data tiers" docs
This is feedback that we [received here](https://github.com/elastic/docs-content/pull/1906#issuecomment-3010845871) from @stefnestor (Thanks Stef!) on the data tiers documentation.

Many of these things seem to be questions that we'd need help from either Development or Support to answer, so we may need to expand this beyond being a simple "docs" issue. 

For now, I'll park the issue here so that we don't lose this valuable feedback.

---

- [data tiers](https://www.elastic.co/docs/manage-data/lifecycle/data-tiers#add-a-data-tier)
    - could/should header#4 show on the right-side TOC?
        - image
    - Per [this](https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore/searchable-snapshots#searchable-snapshot-mount-storage-options) searchable snapshots can happen on hot+cold+frozen for quotes but only guaranteed frozen. there's various wrong logic from not understanding SS can happen on hot in the steps after these quotes
        > The hot and warm tiers store regular indices, while the frozen tier stores searchable snapshots. However, the cold tier can store either regular indices or searchable snapshots. 
        > 
       > When data reaches the cold or frozen phases, it is automatically converted to a [searchable snapshot](https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore/searchable-snapshots) by ILM.
    - "Move shards off the nodes to be removed from the cluster." 
        - they may want to push farther down not up (e.g. warm>cold not warm>hot, maybe noting frozen would be invalid option)
        - needs to ensure to call out they need to drain all shards (confirm with CAT Shards) not just update allocation (not sure how you want to handle node attrs saying "we'll do that during plan") but calling out because frequent support volume for `total_shards_per_node` and/or watermark on destination to block migrations off 
    - "If you do not intend to delete this data, you should manually restore each of the searchable snapshot indices to a regular index before disabling the data tier, by following these steps" fully mounted could just be ported across hot+cold, no? it doesn't have to be rehydrated. 
    - "Capture a comprehensive list of index and searchable snapshot names." I believe this assumes repo is `found-snapshots` but users can have custom repos so this isn't sufficiently valid 
    - "Remove the associated ILM policy (set it to null). If you want to apply a different ILM policy, follow the steps to [Switch lifecycle policies](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/configure-lifecycle-policy#switch-lifecycle-policies)." if you're modifying it you always set to `null` and then remove and then add the new ilm, you do not set `<new-policy-name>` 
    - "Optionally, specify the desired number of replica shards." > `"index.lifecycle.rollover_alias": "<alias-for-rollover>"`: we usually don't re-add a rollover alias because we're not ingesting more into the index, instead we usually set `"index.lifecycle.indexing_complete": true` to bypass the potential new ilm policy's rollover action (which noop if NA) 
    - FWIW 
        - the searchable snapshot rehydration section comes up if you need to fix data not just remove a data tier so it may potentially be worth a different section in the docs
        - see also [KB > Resolve > 4) Clean-up > Searchable Snapshots](https://support.elastic.co/knowledge/39606517) 
    - these are not equal nor even generally related sub-sections. IMO maybe "remove a data tier" should be a par-header since it applies across the board saying something like "to remove a data tier we recommend draining off shards first to avoid data loss, ECE+ECK+ECH will enforce this" 
        - image
        - if you disagree then FWIW "Elastic Cloud Hosted and Elastic Cloud Enterprise try to move all data from the nodes that are removed during plan changes." also applies to ECK software-side. 

---

## Elastic Docs AI Scoping 🤖

<details>
<summary>Docs issue scope</summary>

### Summary
The issue collects expert technical feedback (from `@stefnestor` via PR #1906) on inaccuracies and gaps in the "Data tiers" documentation page. The feedback covers multiple distinct problems: incorrect statements about which tiers support searchable snapshots, shard draining guidance that sends data in the wrong direction, a hardcoded repo name (`found-snapshots`) that will fail for users with custom repos, misleading ILM policy instructions in the restore command, a rollover alias recommendation that contradicts best practice, a missing ECK callout, and structural/navigation issues. There are no linked code changes to reconcile — the issue itself is the source of truth for what needs to change.

### Request accuracy
Accurate — the feedback is specific, technically precise, and confirmed against the current page content.

### Next action for author
Update `manage-data/lifecycle/data-tiers.md` across the sections listed below; consider looping in the ES Data Management team for validation of the searchable-snapshot and ILM policy items before merging.

### Scope boundary
Only `manage-data/lifecycle/data-tiers.md` appears to need changes; the referenced searchable-snapshots and ILM docs pages are cited as supporting evidence in the feedback but are not themselves inaccurate.

### Recommended documentation targets

| Page | URL | Action | Impact | Confidence | Why this page? |
|------|-----|--------|--------|------------|----------------|
| Data tiers | https://www.elastic.co/docs/manage-data/lifecycle/data-tiers | Update existing page | High | High | All feedback items map directly to specific sections of this single page |

### Recommendations

1. **Fix searchable-snapshot tier accuracy** (line 139 of `data-tiers.md`): The statement _"The hot and warm tiers store regular indices, while the frozen tier stores searchable snapshots"_ is incomplete — searchable snapshots can also be mounted on the hot tier. Rewrite to accurately reflect that cold and frozen phases use searchable snapshots by default via ILM, but hot is also a valid mount target; then audit the downstream steps in that section for logic that incorrectly assumes hot is never a searchable-snapshot tier.

2. **Fix shard-draining direction** (step 4 / line 293): The example instructs operators to re-allocate shards _to hot nodes_ when draining a tier. When removing a warm tier, shards should move downward (warm → cold), not upward to hot. Update the example target tier accordingly and add a note that moving to the frozen tier is not a valid destination. Also add a callout to confirm all shards have fully drained using `GET _cat/shards` (not just updating allocation settings), and warn about `total_shards_per_node` limits and watermark thresholds on the destination that can silently block migrations.

3. **Clarify fully-mounted cold indices** (step 5 of the searchable-snapshot section, line 339): The instruction _"manually restore each of the searchable snapshot indices to a regular index"_ is overly prescriptive. Fully mounted indices (cold tier) can be remounted to another supported tier without full rehydration; only partially mounted (frozen) indices require a full restore. Update the wording to distinguish these two cases.

4. **Remove hardcoded `found-snapshots` repo name** (line 401): The restore command uses `POST _snapshot/found-snapshots/...`, which is the Elastic Cloud internal default. Users with custom or self-managed snapshot repositories will fail silently. Replace with a placeholder (e.g., `<snapshot-repository-name>`) and add a note explaining how to list available repositories (`GET _snapshot`).

5. **Fix ILM policy guidance in restore command** (lines 401–408): The restore payload sets `"index.lifecycle.name": "<new-policy-name>"` directly, but the correct procedure is to restore with `null` first, confirm the index is healthy, and then assign a new policy separately. Remove the `index.lifecycle.name` field from the restore body and add a follow-up step explaining the correct order.

6. **Replace rollover-alias with `indexing_complete`** (line 408): Recommending `"index.lifecycle.rollover_alias": "<alias-for-rollover>"` on a restored index is incorrect for non-ingest use cases. Replace with a recommendation to set `"index.lifecycle.indexing_complete": true` to bypass any rollover action in the new ILM policy, and note that the rollover alias should only be set if the index is actually being used for continued ingestion.

7. **Add ECK to platform callout** (line 151): The statement _"Elastic Cloud Hosted and Elastic Cloud Enterprise try to move all data from the nodes that are removed during plan changes"_ also applies to Elastic Cloud on Kubernetes (ECK) on the software side. Add ECK to this callout.

8. **Promote "Remove a data tier" to a top-level section** (structural): The current heading hierarchy places the disable-a-tier content as a subsection of cloud-specific configuration. The feedback notes this guidance applies broadly. Consider elevating _"Remove a data tier"_ to a parallel heading with _"Add a data tier"_ and leading with a cross-platform recommendation to drain all shards first, then handling platform-specific steps in sub-sections. Also revisit whether ECH/ECE/ECK enforce this automatically (the note about ECH/ECE should be extended to ECK).

9. **Consider a standalone section or page for searchable-snapshot rehydration** (low priority): The expert feedback notes that the rehydration steps are useful beyond just the "remove a data tier" flow (e.g., when fixing data issues). Evaluate whether to move or cross-link this content from a more discoverable location, or add a cross-reference from the searchable snapshots page.

10. **TOC heading depth** (low priority): The `#### Add a data tier` heading (h4) may not render in the right-side TOC depending on TOC depth configuration. Evaluate whether to promote this heading or adjust the page's `toc_depth` frontmatter if supported.

### Notes

- Items 1–7 are accuracy or correctness issues that should be validated by the ES Data Management or Support team before merging.
- The `found-snapshots` hardcoding (item 4) is a particularly common failure point for self-managed and custom-repo users.
- Items 8–10 are structural improvements that can be addressed independently and at lower priority.

</details>

---
[Docs automation](https://github.com/elastic/docs-actions) | [From workflow: Issue Scope Analyzer](https://github.com/elastic/docs-content/actions/runs/24512526607)

React with 👍 if helpful, 👎 if not.

yetanothertw:
## Summary

Fixes #1956 Adds Stef's comments from https://github.com/elastic/docs-content/pull/1906#issuecomment-3010845871

* Adds clarification around the hot tier and searchable snapshots: an ILM `searchable_snapshot` action in the hot phase can leave a fully mounted searchable snapshot on hot (with link to the ILM action and searchable snapshots).
* Aligns tier vs. searchable snapshot copy with ILM (e.g. fully mounted on hot/warm/cold; frozen = partial only; cold can be regular or snapshot-backed).
* For remove tiers (regular indices): added a preference for moving allocation down the stack (e.g. warm to cold), not only to hot; frozen called out as wrong for regular indices; ECK mentioned with shard relocation; watermarks on destination nodes.
* new tip to reallocate fully mounted `restored-*` via `_tier_preference` when possible;

## Generative AI disclosure

- [x] Yes  
- [ ] No  

Used Claude's Auto Agent to verify claims and find relevant doc links.


yetanothertw:
Hi @kilfoyle and @stefnestor,

I've added you both as reviewers on this PR since you'd have more context around the [feedback that Stef provided](https://github.com/elastic/docs-content/pull/1906#issuecomment-3010845871) for this section (Data tiers and the child pages for adding or removing data tiers in various deployments).

Stef, please let me know if I've interpreted your suggestions correctly. For some of this I've relied on Claude for help, so would really appreciate your guidance.

Please feel free to leave any suggestions, comments, or ideas.

Many thanks in advance! 🙏

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -75,9 +75,9 @@ If in doubt, reach out to Support.
...
+When you update the deployment, {{ech}} and {{ece}} try to move all data off the nodes that are removed. On {{eck}}, node set and scaling changes similarly try to relocate shards off nodes that are removed, subject to the same allocation rules, capacity, and [disk watermarks](elasticsearch://reference/elasticsearch/configuration-reference/cluster-level-shard-allocation-routing-settings.md#disk-based-shard-allocation) on the destination nodes. For more information, refer to [{{eck}}](/deploy-manage/deploy/cloud-on-k8s.md). To disable a tier that holds only regular indices, make sure that all data on that tier can be re-allocated by reconfiguring the relevant shard allocation filters. You’ll also need to temporarily stop your {{ilm-init}} policies to prevent new indices from being moved to the data tier you want to disable.
```
kilfoyle:
```suggestion
When you update the deployment, {{ech}} and {{ece}} try to move all data off the nodes that are removed. On {{eck}}, node set and scaling changes similarly try to relocate shards off nodes that are removed, subject to the same allocation rules, capacity, and [disk watermarks](elasticsearch://reference/elasticsearch/configuration-reference/cluster-level-shard-allocation-routing-settings.md#disk-based-shard-allocation) on the destination nodes. For more information, refer to the [{{eck}}](/deploy-manage/deploy/cloud-on-k8s.md) documentation. To disable a tier that holds only regular indices, make sure that all data on that tier can be re-allocated by reconfiguring the relevant shard allocation filters. You’ll also need to temporarily stop your {{ilm-init}} policies to prevent new indices from being moved to the data tier you want to disable.
```
We might ask our friend Claude if there's a more specific section that we could link to in the ECK docs, but as is good too.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -163,21 +163,23 @@ To make sure that all data can be migrated from the data tier you want to disabl
         }
         ```
 
-        You must remove the relevant tier from the inclusion rules. For example, to disable the warm tier, the `data_warm` tier preference should be removed:
+        You must remove the relevant tier from the inclusion rules. For example, to disable the warm tier, remove the `data_warm` and set `_tier_preference` to a tier you are keeping. Prefer promoting data through the lifecycle, for example from warm to cold, not back to hot, unless the cluster has no colder tier to accept the data:
```
kilfoyle:
```suggestion
        You must remove the relevant tier from the inclusion rules. For example, to disable the warm tier, remove the `data_warm` parameter and set `_tier_preference` to a tier you are keeping. Prefer promoting data through the lifecycle, for example from warm to cold, not back to hot, unless the cluster has no colder tier to accept the data:
```

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -213,21 +215,23 @@ To make sure that all data can be migrated from the data tier you want to disabl
         }
         ```
 
-        Removing required attributes does not trigger a shard reallocation. These shards are moved when applying the plan to disable the data tier. Alternatively, you can use the [cluster re-route API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-reroute) to manually re-allocate the shards before removing the nodes, or explicitly re-allocate shards to hot nodes by using the following code:
+        Removing required attributes does not trigger a shard reallocation. These shards are moved when applying the plan to disable the data tier. Alternatively, you can use the [cluster re-route API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-reroute) to manually re-allocate the shards before removing the nodes, or set `require` so shards target a remaining lower tier. For example, for warm to cold migration, instead of warm to hot:
```
kilfoyle:
```suggestion
        Removing required attributes does not trigger a shard reallocation. These shards are moved when applying the plan to disable the data tier. Alternatively, you can use the [cluster re-route API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-reroute) to manually re-allocate the shards before removing the nodes, or set `require` so shards target a remaining lower tier. For example, to set warm to cold migration, instead of warm to hot:
```

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -245,6 +249,10 @@ To make sure that all data can be migrated from the data tier you want to disabl
         }
         ```
 
+    :::{important}
+    Confirm that no shards are left on the nodes to be removed after allocation settles: `GET /_cat/shards` (filter by `node` as needed) should show the tier empty. Updating settings starts relocation, but you must wait until [shard allocation and recovery](/deploy-manage/distributed-architecture/shard-allocation-relocation-recovery.md) finish. If shards stay on the old tier, use the [cluster allocation explain](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-allocation-explain) API. Common blockers can be [disk watermarks](/troubleshoot/elasticsearch/fix-watermark-errors.md) or [`index.routing.allocation.total_shards_per_node`](elasticsearch://reference/elasticsearch/index-settings/total-shards-per-node.md#total-shards-per-node) on the destination nodes.
```
kilfoyle:
```suggestion
    Confirm that no shards are left on the nodes to be removed after the allocation completes: `GET /_cat/shards` (filter by `node` as needed) should show that the tier is empty. Updating settings starts the relocation process, but you must wait until [shard allocation and recovery](/deploy-manage/distributed-architecture/shard-allocation-relocation-recovery.md) finish. If shards stay on the original tier, use the [cluster allocation explain](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-allocation-explain) API to determine the cause. Common reasons can be [disk watermarks](/troubleshoot/elasticsearch/fix-watermark-errors.md) or [`index.routing.allocation.total_shards_per_node`](elasticsearch://reference/elasticsearch/index-settings/total-shards-per-node.md#total-shards-per-node) on the destination nodes.
```
Maybe this improves the clarity a bit.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -259,7 +267,11 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
+:::{tip}
+Fully mounted `restored-*` [{{search-snaps}}](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#fully-mounted) on the hot, warm, or cold tier can often be moved to another remaining tier by updating [`index.routing.allocation.include._tier_preference`](#update-data-tier-allocation-rules) and related [allocation rules](/deploy-manage/distributed-architecture/shard-allocation-relocation-recovery/index-level-shard-allocation.md) alone. You do not need `POST _snapshot/.../_restore` in that case. Use a full restore to a new regular index when you are [rehydrating](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md) from a snapshot, or when you have [partially mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#partially-mounted) `partial-*` indices on frozen and need them as regular indices on another tier.
```
kilfoyle:
```suggestion
Fully mounted `restored-*` [{{search-snaps}}](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#fully-mounted) on the hot, warm, or cold tier can often be moved to another remaining tier by updating the [`index.routing.allocation.include._tier_preference`](#update-data-tier-allocation-rules) setting and related [allocation rules](/deploy-manage/distributed-architecture/shard-allocation-relocation-recovery/index-level-shard-allocation.md) alone. You do not need to run the `POST _snapshot/.../_restore` request in that case. Use a full restore to a new regular index when you are [rehydrating](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md) from a snapshot, or when you have [partially mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#partially-mounted) `partial-*` indices on a frozen tier and need them as regular indices on another tier.
```

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -259,7 +267,11 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
+:::
+
+When an [{{ilm-init}} policy’s `searchable_snapshot` action](elasticsearch://reference/elasticsearch/index-lifecycle-actions/ilm-searchable-snapshot.md) runs in `hot`, `warm`, `cold`, or `frozen`, it can convert a managed index into a [{{search-snap}}](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md) (`restored-*` in non-frozen phases, or `partial-*` in `frozen`). If the data is no longer required, the delete phase of the same policy (or a manual delete) can remove it. If you must retain the data while removing the tier, follow the following steps:
```
kilfoyle:
```suggestion
When an {{ilm-init}} policy’s [`searchable_snapshot` action](elasticsearch://reference/elasticsearch/index-lifecycle-actions/ilm-searchable-snapshot.md) runs in a `hot`, `warm`, `cold`, or `frozen` phase, it can convert a managed index into a [{{search-snap}}](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md) (`restored-*` in non-frozen phases, or `partial-*` in the `frozen` phase). If the data is no longer required, the `delete` phase of the same policy (or a manual delete) can remove it. If you must retain the data while removing the tier, follow these steps:
```

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -271,14 +283,20 @@ When data reaches the `cold` or `frozen` phases, it is automatically converted t
     GET _ilm/status
     ```
 
-2. Capture a comprehensive list of index and {{search-snap}} names.
+2. Capture a comprehensive list of index and {{search-snap}} names, and which snapshot repository each snapshot lives in.
 
-    1. The index name of the {{search-snaps}} may differ based on the data tier. If you intend to disable the cold tier, then perform the following request with the `restored-*` prefix. If the frozen tier is the one to be disabled, use the `partial-*` prefix.
+    1. The index name of the {{search-snaps}} may differ based on the data tier. If you intend to disable the cold tier, use the `restored-*` prefix. If the frozen tier is the one to be disabled, use the `partial-*` prefix. If you are removing a tier that had a [searchable_snapshot](elasticsearch://reference/elasticsearch/index-lifecycle-actions/ilm-searchable-snapshot.md) action in an earlier phase, for example such as hot, also run the same query for `restored-*` on that tier.
```
kilfoyle:
```suggestion
    1. The index name of the {{search-snaps}} may differ based on the data tier. If you intend to disable the cold tier, use the `restored-*` prefix. If the frozen tier is the one to be disabled, use the `partial-*` prefix. If you are removing a tier that had a [searchable_snapshot](elasticsearch://reference/elasticsearch/index-lifecycle-actions/ilm-searchable-snapshot.md) action in an earlier phase, for example during the `hot` phase, also run the same query for `restored-*` on that tier.
```

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -271,14 +283,20 @@ When data reaches the `cold` or `frozen` phases, it is automatically converted t
...
 
         ```sh
         GET <searchable-snapshot-index-prefix>/_settings?filter_path=**.index.store.snapshot.snapshot_name&expand_wildcards=all
         ```
 
+    2. List snapshot repository names and resolve which repository to use in the restore API. {{ech}} often defaults to a repository named `found-snapshots`, but clusters can register additional repositories (for example with the [create snapshot repository API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-snapshot-create-repository)), so do not assume a single name.
```
kilfoyle:
```suggestion
    2. List snapshot repository names to determine which repositories to use when you call the restore API. {{ech}} uses a repository named `found-snapshots` by default, but you can also register additional repositories for clusters (for example by using the [create snapshot repository API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-snapshot-create-repository)). Be careful not to assume that there is only one repository name.
```

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -315,27 +333,25 @@ When data reaches the `cold` or `frozen` phases, it is automatically converted t
...
-    4. Optionally, specify the desired number of replica shards.
+    1. Follow the steps to [specify the data tier based allocation inclusion rules](/manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md#update-data-tier-allocation-rules) for the tier you are keeping.
+    2. Use the repository and snapshot names you captured earlier in step 2. In the restore request, set `index.lifecycle.name` to `null` and set [`index.lifecycle.indexing_complete`](elasticsearch://reference/elasticsearch/configuration-reference/index-lifecycle-management-settings.md#_index_level_settings_2) to `true` for indices that are not going to keep ingesting as a rollover write index. Do not set a new `index.lifecycle.name` in the restore body. After the index is `green`, assign a different policy, if you need one, in a separate `PUT` request, following the steps described in [Switch lifecycle policies](/manage-data/lifecycle/index-lifecycle-management/policy-updates.md#switch-lifecycle-policies) which remove the old policy, then add the new one. Only set `index.lifecycle.rollover_alias` if you will keep using the index as a rollover target; otherwise leave it unset.
```
kilfoyle:
```suggestion
    2. Use the repository and snapshot names you captured earlier in step 2. In the restore request, set `index.lifecycle.name` to `null` and set [`index.lifecycle.indexing_complete`](elasticsearch://reference/elasticsearch/configuration-reference/index-lifecycle-management-settings.md#_index_level_settings_2) to `true` for indices that are not going to keep ingesting as a rollover write index. Do not set a new `index.lifecycle.name` in the restore body. After the index is `green`, assign a different policy, if you need one, in a separate `PUT` request. Follow the steps described in [Switch lifecycle policies](/manage-data/lifecycle/index-lifecycle-management/policy-updates.md#switch-lifecycle-policies) to remove the old policy and then add a new one. Only set `index.lifecycle.rollover_alias` if you will keep using the index as a rollover target; otherwise leave it unset.
```

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -315,27 +333,25 @@ When data reaches the `cold` or `frozen` phases, it is automatically converted t
...
-            "index.lifecycle.rollover_alias": "<alias-for-rollover>"
+            "index.number_of_replicas": 0,
+            "index.lifecycle.name": null,
+            "index.lifecycle.indexing_complete": true
           }
         }
         ```
 
-        The  `<searchable_snapshot_name>` refers to the above-mentioned step: "Capture a comprehensive list of index and {{search-snap}} names".
+        The `<searchable_snapshot_name>` are obtained from step 2. Adjust `index.number_of_replicas` to match your resiliency needs.
```
kilfoyle:
```suggestion
        For `<searchable_snapshot_name>` use the names that you obtained in step 2. Adjust `index.number_of_replicas` to match your resiliency needs.
```

kilfoyle (approved):
LGTM! 🦘 
Really nice work, @yetanothertw! This super complex procedure looks a lot more clear now.

I suggested a few phrasing improvements, but of course the amazing @stefnestor will be able to confirm the technical part for us. :-)

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
yetanothertw: I actually moved this into a note on the Configure data tiers for self-managed and Elastic Cloud on Kubernetes deployments page, since I've separated out the ECE/ECH and self-managed/ECK into child pages.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
yetanothertw: thanks, fixed in a separate commit

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
yetanothertw: follow the following 🤕

yetanothertw: Fixed in a new commit, thanks

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -61,7 +61,7 @@ Read [https://www.elastic.co/cloud/shared-responsibility](https://www.elastic.co
 If in doubt, reach out to Support.
 :::
 
-* Know whether you are disabling a tier that stores **regular indices** (for example hot, warm, or cold without {{search-snaps}}) or **{{search-snaps}}** (for example cold or frozen with {{ilm-init}}). The **cold** tier can hold either; use these requests to check for {{search-snap}} indices:
+* Know whether you are disabling a tier that stores regular indices or {{search-snaps}}. The frozen tier only holds [partially mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#partially-mounted) {{search-snaps}}. The cold tier can hold regular indices or [fully mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#fully-mounted) {{search-snaps}}. The hot and warm tiers usually hold regular indices, but an {{ilm-init}} policy can mount a fully mounted {{search-snap}} on the hot or warm tier (for example, when the [searchable_snapshot](elasticsearch://reference/elasticsearch/index-lifecycle-actions/ilm-searchable-snapshot.md) action runs in the `hot` or `warm` phase). Use these requests to check for {{search-snap}} indices on the tier you are removing:
```
stefnestor: As noted top of [this doc](https://www.elastic.co/docs/reference/elasticsearch/index-lifecycle-actions/ilm-searchable-snapshot) ILM searchable snapshots only apply mounted to hot+cold and partial to frozen. Warm isn't in the mix & supposedly never hosts searchable snapshots. Info confirmation [link](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/index-lifecycle#phases-availability).

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -75,9 +75,9 @@ If in doubt, reach out to Support.
 
 ### Remove a tier with regular indices [non-searchable-snapshot-data-tier]
 
-The hot and warm tiers store regular indices, while the frozen tier stores {{search-snaps}}. The cold tier can store either; use the checks in [Before you remove a data tier](#before-you-remove-a-data-tier) if you are unsure.
+The frozen tier only stores [partially mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#partially-mounted) {{search-snaps}}. [Fully mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#fully-mounted) {{search-snaps}} can be allocated to the hot, warm, or cold tier depending on the {{ilm-init}} phase, while the cold tier can also hold regular indices. Use the checks in [Before you remove a data tier](#before-you-remove-a-data-tier) if you are unsure what is on the tier.
```
stefnestor: ditto ☝️

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -163,21 +163,23 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
+        You must remove the relevant tier from the inclusion rules. For example, to disable the warm tier, remove the `data_warm` parameter and set `_tier_preference` to a tier you are keeping. Prefer promoting data through the lifecycle, for example from warm to cold, not back to hot, unless the cluster has no colder tier to accept the data:
 
         ```sh
         PUT /my-index/_settings
         {
             "routing": {
               "allocation": {
                 "include": {
-                    "_tier_preference": "data_hot"
+                    "_tier_preference": "data_cold,data_hot" <1>
```
stefnestor: Per [ILM Migrate](https://www.elastic.co/docs/reference/elasticsearch/index-lifecycle-actions/ilm-migrate) the `_tier_preference` sets `data_cold,data_warm,data_hot` and not just `data_cold,data_hot`.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -213,21 +215,23 @@ To make sure that all data can be migrated from the data tier you want to disabl
         }
         ```
 
-        Removing required attributes does not trigger a shard reallocation. These shards are moved when applying the plan to disable the data tier. Alternatively, you can use the [cluster re-route API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-reroute) to manually re-allocate the shards before removing the nodes, or explicitly re-allocate shards to hot nodes by using the following code:
+        Removing required attributes does not trigger a shard reallocation. These shards are moved when applying the plan to disable the data tier. Alternatively, you can use the [cluster re-route API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-reroute) to manually re-allocate the shards before removing the nodes, or set `require` so shards target a remaining lower tier. For example, to set warm to cold migration, instead of warm to hot:
```
stefnestor:
```suggestion
        Removing required attributes does not trigger a shard reallocation. These shards are moved when applying the plan to disable the data tier. Alternatively, you can use the [cluster re-route API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-reroute) to manually re-allocate the shards before removing the nodes, or set `require` to migrate shards to a desired tier. For example to force an index to nodes with `data` attribute of `cold`:
```

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -213,21 +215,23 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
                 "require": {
-                    "data": "hot"
+                    "data": "cold"
                 }
               }
             }
         }
         ```
 
+        Adjust the `data` value to match the [custom node attributes](elasticsearch://reference/elasticsearch/configuration-reference/node-settings.md#custom-node-attributes) and [index-level shard allocation filters](elasticsearch://reference/elasticsearch/index-settings/shard-allocation.md) your indices already use. Prefer cold or the next lower tier over hot when you are only removing a middle tier. Do not send regular indices to the frozen tier.
```
stefnestor:
```suggestion
        Adjust the `data` value to match the [custom node attributes](elasticsearch://reference/elasticsearch/configuration-reference/node-settings.md#custom-node-attributes) and [index-level shard allocation filters](elasticsearch://reference/elasticsearch/index-settings/shard-allocation.md) your indices already use.  You cannot send regular indices to the frozen tier.
```

> Prefer cold or the next lower tier over hot when you are only removing a middle tier.

FWIW hot>warm>cold>frozen is industry-ish but our jargon. So there is no "next lower" other than warm between hot+cold.

It's probably not worth spending time to unwind the historical dredge here. 

But for context, historically, users used to self-manage data temperatures via the `allocation.require.data` setting. We then moved them as part of ILM over to `_tier_preference` with the understanding that they could still use the first, but it should be for quote unquote "other things" now instead of data temperatures (otherwise you can sometimes end up with blocking setting conflicts). For example, if hardware profiles are uneven and they want to self-manage that. 

But calling out in context, while definitely allowed, it's also something Support flags to users as "probably indicates your migration onto ILM is incomplete" _every time_ we see this type of `data: cold`, it's very 👻 "things are very likely going to break" flag to us.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -245,6 +249,10 @@ To make sure that all data can be migrated from the data tier you want to disabl
         }
         ```
 
+    :::{important}
+    Confirm that no shards are left on the nodes to be removed after the allocation completes: `GET /_cat/shards` (filter by `node` as needed) should that show the tier is empty. Updating settings starts the relocation process, but you must wait until [shard allocation and recovery](/deploy-manage/distributed-architecture/shard-allocation-relocation-recovery.md) finish. If shards stay on the original tier, use the [cluster allocation explain](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-allocation-explain) API to determine the cause. Common reasons can be [disk watermarks](/troubleshoot/elasticsearch/fix-watermark-errors.md) or [`index.routing.allocation.total_shards_per_node`](elasticsearch://reference/elasticsearch/index-settings/total-shards-per-node.md#total-shards-per-node) on the destination nodes.
```
stefnestor: I like it. Noting CAT Allocation to see IF any shards remain is way easier for users to check when node gets to 0 (at least until https://github.com/elastic/elasticsearch/issues/141786). And then CAT Shards comes up to see which shards if they need to investigate.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -259,7 +267,11 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
+:::{tip}
+Fully mounted `restored-*` [{{search-snaps}}](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#fully-mounted) on the hot, warm, or cold tier can often be moved to another remaining tier by updating the [`index.routing.allocation.include._tier_preference`](#update-data-tier-allocation-rules) setting and related [allocation rules](/deploy-manage/distributed-architecture/shard-allocation-relocation-recovery/index-level-shard-allocation.md) alone. You do not need to run `POST _snapshot/.../_restore` request in that case. Use a full restore to a new regular index when you are [rehydrating](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md) from a snapshot, or when you have [partially mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#partially-mounted) `partial-*` indices on a frozen tier and need them as regular indices on another tier.
```
stefnestor: Again not warm.

As right but carefulness on phrasing: users can technically self-manage searchable snapshots in which case they can call them whatever they want and them still be fully mounted. When ILM manages the index, we as Elastic chose `restored-*` as our index pattern.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -259,7 +267,11 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
+:::
+
+When an [{{ilm-init}} policy’s `searchable_snapshot` action](elasticsearch://reference/elasticsearch/index-lifecycle-actions/ilm-searchable-snapshot.md) runs in a `hot`, `warm`, `cold`, or `frozen` phase, it can convert a managed index into a [{{search-snap}}](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md) (`restored-*` in non-frozen phases, or `partial-*` in the `frozen` phase). If the data is no longer required, the `delete` phase of the same policy (or a manual delete) can remove it. If you must retain the data while removing the tier, follow these steps:
```
stefnestor: ditto not warm

IF you just do a "manual delete" you will leave a stray searchable snapshot in the repository. If you do that in mass or over time, you can cause repository-side disk issues.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -271,14 +283,20 @@ When data reaches the `cold` or `frozen` phases, it is automatically converted t
...
+    1. The index name of the {{search-snaps}} may differ based on the data tier. If you intend to disable the cold tier, use the `restored-*` prefix. If the frozen tier is the one to be disabled, use the `partial-*` prefix. If you are removing a tier that had a [searchable_snapshot](elasticsearch://reference/elasticsearch/index-lifecycle-actions/ilm-searchable-snapshot.md) action in an earlier phase, for example during the `hot` phase, also run the same query for `restored-*` on that tier.
 
         ```sh
         GET <searchable-snapshot-index-prefix>/_settings?filter_path=**.index.store.snapshot.snapshot_name&expand_wildcards=all
         ```
 
+    2. List snapshot repository names to determine which repositories to use when you call the restore API. {{ech}} uses a repository named `found-snapshots` by default, but you can also register additional repositories for clusters (for example by using the [create snapshot repository API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-snapshot-create-repository)). Be careful not to assume that there is only one repository name.
```
stefnestor: The command just prior reports _per index_ both its repository name and snapshot name & since users can Searchable Snapshots point to different repositories per ILM policy (and/or its changes across versions), you don't want to run this command as a global-assume for all indices
```
filter_path=**.index.store.snapshot
```

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -315,27 +333,25 @@ When data reaches the `cold` or `frozen` phases, it is automatically converted t
...
-    4. Optionally, specify the desired number of replica shards.
+    1. Follow the steps to [specify the data tier based allocation inclusion rules](/manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md#update-data-tier-allocation-rules) for the tier you are keeping.
+    2. Use the repository and snapshot names you captured earlier in step 2. In the restore request, set `index.lifecycle.name` to `null` and set [`index.lifecycle.indexing_complete`](elasticsearch://reference/elasticsearch/configuration-reference/index-lifecycle-management-settings.md#_index_level_settings_2) to `true` for indices that are not going to keep ingesting as a rollover write index. Do not set a new `index.lifecycle.name` in the restore body. After the index is `green`, assign a different policy, if you need one, in a separate `PUT` request. Follow the steps described in [Switch lifecycle policies](/manage-data/lifecycle/index-lifecycle-management/policy-updates.md#switch-lifecycle-policies) to remove the old policy and then add a new one. Only set `index.lifecycle.rollover_alias` if you will keep using the index as a rollover target; otherwise leave it unset.
```
stefnestor: > if you need one, in a separate `PUT` request.

Indices restore with their earlier ILM cache so you still have to follow https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/policy-updates#switch-lifecycle-policies to pivot to new policy. It's not just a "separate `PUT`" you actually have to run https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ilm-remove-policy

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -315,27 +333,25 @@ When data reaches the `cold` or `frozen` phases, it is automatically converted t
...
 
         ```sh
-        POST _snapshot/found-snapshots/<searchable_snapshot_name>/_restore
+        POST _snapshot/<snapshot-repository-name>/<searchable_snapshot_name>/_restore
```
stefnestor:
```suggestion
        POST _snapshot/<snapshot_repository_name>/<searchable_snapshot_name>/_restore
```
nitpick: formatting par to previous

stefnestor:
I like it, cheers @yetanothertw !! 🙇‍♀️

FWIW noting for you & @kilfoyle that high-level the content you're editing about removing a data tier is on the page for ECH+ECE but applies to _all_ of stack (ECK & self-managed too). Only the screenshots part is ECH+ECE specific the but ES-level is all platforms & we don't have that same content elsewhere. Could we perhaps move the section under the "Data Tiers" page instead?

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
yetanothertw: Thanks, removing mentions of warm in a new commit

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
yetanothertw: Thanks, fixed in a separate commit

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
yetanothertw: ok, removing that in a separate commit

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
yetanothertw: Fixing in a new commit, thanks

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
yetanothertw: Ok, changing back to the original doc as I think I've made this more confusing than it was.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
yetanothertw: I've tried to reword that to make that point.

yetanothertw:
Hi @stefnestor, thank you for reviewing this! 🌞 

> high-level the content you're editing about removing a data tier is on the page for ECH+ECE but applies to all of stack (ECK & self-managed too). Only the screenshots part is ECH+ECE specific the but ES-level is all platforms & we don't have that same content elsewhere. Could we perhaps move the section under the "Data Tiers" page instead?

Interesting that you say that. For https://github.com/elastic/docs-content/issues/2311 I've broken up the [Data tiers](https://www.elastic.co/docs/manage-data/lifecycle/data-tiers) page into these two child pages:
* [Manage data tiers in ECH or ECE](https://www.elastic.co/docs/manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece)
* [Manage data tiers in self-managed and ECK](https://www.elastic.co/docs/manage-data/lifecycle/data-tiers/manage-data-tiers-self-managed-eck)

I think rather than moving everything back into one giant page, we can add the missing info about removing data tiers in self-managed and ECK on this child page: [Manage data tiers in self-managed and ECK](https://www.elastic.co/docs/manage-data/lifecycle/data-tiers/manage-data-tiers-self-managed-eck)

Is that something you could help me with? I can create an issue and maybe you could help me scope/refine it? Or you can create a draft PR and I can edit that if that's easier, wdyt?

I've created a new issue here: https://github.com/elastic/docs-content/issues/6186

stefnestor:
👋 @yetanothertw, cheers! FWIW the "draft" is the exact same section exactly as we've modified it just without the screenshots (potentially replacing with text form of `console` in+out puts). ES treats this platform agnostic, so it's literally the same outside how platform-level you spin down afterwards.

PR #6077, opened 2026-04-22
[Observability] 9.4.0 release notes

mdbirnstiehl:
## Summary

This PR closes #6075 and adds the 9.4.0 Observability release notes.

## Generative AI disclosure

1. Did you use a generative AI (GenAI) tool to assist in creating this contribution?
- [x] Yes  
- [ ] No  

2. If you answered "Yes" to the previous question, please specify the tool(s) and model(s) used (e.g., Google Gemini, OpenAI ChatGPT-4, etc.).

Tool(s) and model(s) used: Claude Code

PR #6093, opened 2026-04-23
Update index.md

miguelsantosalvarez:
It is planned for the stack but only available for Serverless

## Summary
As described on [the docs](https://www.elastic.co/docs/solutions/observability/streams/management/extract/concat) it is available for Serverless but not for the rest of the stack for the moment. The note without context made one of our customer think they have it available already.

## Generative AI disclosure

- [ ] Yes  
- [ X] No

### DIFF
--- a/release-notes/elastic-observability/index.md
+++ b/release-notes/elastic-observability/index.md
@@ -53,7 +53,6 @@ To check for security updates, go to [Security announcements for the Elastic sta
 
 
 ### Features and enhancements [elastic-observability-9.3.1-features-enhancements]
-* Adds a concat processor to Streams on Serverless [#247940]({{kib-pull}}247940).
 * Adds error handling to {{observability}} landing page redirect logic [#254171]({{kib-pull}}254171).
 * Adds support for ECS formatted errors in service details [#254138]({{kib-pull}}254138).
 
--- a/release-notes/elastic-observability/index.md
+++ b/release-notes/elastic-observability/index.md
@@ -53,7 +53,7 @@ To check for security updates, go to [Security announcements for the Elastic sta
 
 
 ### Features and enhancements [elastic-observability-9.3.1-features-enhancements]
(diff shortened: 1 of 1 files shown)


theletterf:
@florent-leborgne

release-notes/elastic-observability/index.md
```diff
@@ -53,7 +53,7 @@ To check for security updates, go to [Security announcements for the Elastic sta
 
 
 ### Features and enhancements [elastic-observability-9.3.1-features-enhancements]
-* Adds a concat processor to Streams [#247940]({{kib-pull}}247940).
+* Adds a concat processor to Streams on Serverless [#247940]({{kib-pull}}247940).
```
florent-leborgne:
```suggestion
```
It should be removed altogether as if that's serverless only then it would be announced on the serverless release notes page (it is actually on that page here https://www.elastic.co/docs/release-notes/cloud-serverless#serverless-changelog-01262026-features-enhancements)

cc @mdbirnstiehl to double check that this will indeed be caught by Obs 9.4 release notes

florent-leborgne (changes requested):
Thanks @miguelsantosalvarez, I actually think that this should be removed from this release note as it's only going to be from 9.4 - serverless release notes are separate (see my inline comment), this page here is just for traditional versioned releases

release-notes/elastic-observability/index.md
miguelsantosalvarez: It makes totally sense, thanks a lot @florent-leborgne

florent-leborgne (approved):

release-notes/elastic-observability/index.md
mdbirnstiehl: Got it in the [9.4 release notes](https://github.com/elastic/docs-content/pull/6077/changes#diff-80bbf1d270ca944bb6746185ccedefd7a7a7a37b2e25ce9140eecb5fc913cc77R67)

PR #6097, opened 2026-04-23
Fix inaccurate Elastic Agent standalone logging config docs

### LINKED ISSUE #5780: [Internal]: Elastic Agent Logging Configuration documentation is inaccurate and misleading
### Description

Elastic Agent logging has a set of configurable outputs. These are controlled by `agent.logging.to_xxx` settings. The possible output options are: `to_stderr`, `to_files`, `to_syslog` and `to_eventlog`, and they each can have their own settings. This is documented on the [following page](https://www.elastic.co/docs/reference/fleet/elastic-agent-standalone-logging-config).

This documentation is inaccurate in two major ways.

#### The internal logging output is not configurable

Elastic Agent will always log to a set of files, as if `to_files` was enabled. In fact, the documentation claims this is so. However, this is an internal implementation detail of Elastic Agent, as is not configurable at all. It's a completely separate logging output which we don't let users configure, because it risks breaking important features. The current documentation, however, makes it sound like opposite is true:

```
`agent.logging.files.path`
The directory that log files is written to.
Logs file names end with a date and optional number: log-date.ndjson, log-date-1.ndjson, and so on as new files are created during rotation.

macOS: /Library/Elastic/Agent/data/elastic-agent-*/logs/elastic-agent.ndjson
Linux: /opt/Elastic/Agent/data/elastic-agent-*/logs/elastic-agent.ndjson
Windows: C:\Program Files\Elastic\Agent\data\elastic-agent-*\logs\elastic-agent.ndjson
DEB: /var/lib/elastic-agent/data/elastic-agent-*/logs/elastic-agent.ndjson
RPM: /var/lib/elastic-agent/data/elastic-agent-*/logs/elastic-agent.ndjson
```

These are in fact the paths agent will write logs to as part of the internal output, but modifying this option will not affect them at all.

#### Logging outputs are mutually exclusive and silently follow a priority list

By default, we have `to_stderr: true`. If a user sets `to_files: true`, it unexpectedly has no effect. The reason is that only one output can be active at a time. However, if multiple outputs are enabled, agent will silently pick one according to a priority list without indicating it in any way. The priority list goes like so `to_stderr > to_syslog > to_event_log > to_files`.

For example, if the user actually wants agent to log to rotating files of their own choice, they need to set `to_stderr: false` in addition to the `to_files` configuration.

#### Summary

These two inaccuracies make the experience very confusing for users who want to configure agent's existing log files. The documentation indicates that this is possible, but it's not. Furthermore, enabling the option does nothing, as `to_stderr` is enabled by default and has higher priority.

### Resources

No new features. As far as I can tell, this is the case since at least 2020. The agent team would like to improve this situation, but doing so would be a breaking change and needs to wait until 10.0.

### Feature differences

This feature is identical in all deployment methods.

N/A

### Serverless release

Available since the start.

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @swiatekm 

**Stakeholders:** @ebeahan @cmacknz

alexandra5000:
## Summary

- Clarifies that Elastic Agent always writes to an internal set of rotating log files that cannot be disabled or reconfigured, and documents the per-platform paths for those internal logs.
- Documents that external logging outputs (`to_stderr`, `to_syslog`, `to_eventlog`, `to_files`) are mutually exclusive and silently follow a priority order: `to_stderr` > `to_syslog` > `to_eventlog` > `to_files`.
- Updates the example configuration to include `to_stderr: false`, showing the correct way to enable file logging.
- Corrects the default for `agent.logging.to_files` from `true` to `false`.

Closes https://github.com/elastic/docs-content/issues/5780


swiatekm (approved):

theletterf (approved):
LGTM

PR #6111, opened 2026-04-24
self-signed statements reviewed in ECK http and transport TLS docs

eedugon:
Both the transport and http docs of ECK include some statements about ECK creating self-signed certificates for the instances, where that's not really the case.

ECK now generates a CAs (self-signed of course) for transport and http, and then uses the CAs to issue the required certs for the instances, hence ECK is not using self-signed certs.

Besides that I've added the following small improvements:
- Provide the command to grab the HTTP CA (which is actually more useful than providing the command to grab the cert that we are currently giving and it's not really useful (except if the cert was really self-signed :) )
- Change `isCA: true` to false in one of the manifests for cert-manager, as the objective of the manifest is to create a server cert and not a CA.
- Updated `hulk` examples to `quickstart`.

PS - I strongly believe in the past ECK was probably creating self-signed certs for HTTP by default, because multiple items point to that:
- The `spec` feature is called `tls.selfSignedCertificate`, even if it's creating a CA + cert.
- Originally we told the user to grab `tls.crt` from the public HTTP secret, which is not useful anymore if that cert is signed by a private ECK-managed CA (a user needs the CA and not the HTTP cert for trusting purposes).


deploy-manage/security/k8s-https-settings.md
```diff
@@ -87,7 +97,9 @@ spec:
       - quickstart
 ```
 
-Here is how to issue multiple {{es}} certificates from a single self-signed CA. This is useful for example for [Remote clusters](/deploy-manage/remote-clusters/eck-remote-clusters.md) which need to trust each other’s CA, in order to avoid mounting N CAs when a cluster is connected to N other clusters.
+#### Custom private CA and certificate using cert-manager [k8s_custom_http_ca_using_cert_manager]
```
eedugon, line 100: I've added a new header because the example below does not create a self-signed cert (previous header and example), but a CA + cert.

deploy-manage/security/k8s-https-settings.md
```diff
@@ -87,7 +97,9 @@ spec:
...
+
+Here is how to set up a private CA and use it to issue {{es}} certificates. Additional certificates can be issued from the same CA, allowing multiple clusters or services to share a common trust root. This is useful, for example, for [Remote clusters](/deploy-manage/remote-clusters/eck-remote-clusters.md), which need to trust each other’s CA. Using a shared CA avoids having to configure and mount N different CAs when a cluster is connected to N other clusters.
```
marciw:
```suggestion
To set up a private CA and issue {{es}} certificates, you can use `cert-manager`. Additional certificates can be issued from the same CA, allowing multiple clusters or services to share a common trust root. This is useful, for example, for [Remote clusters](/deploy-manage/remote-clusters/eck-remote-clusters.md), which need to trust each other’s CAs. Using a shared CA avoids having to configure and mount N different CAs when a cluster is connected to N other clusters.
```

marciw (approved):
LGTM with one small suggestion

deploy-manage/security/k8s-https-settings.md
marciw: (mostly trying to write around "Here is how to," which isn't quite idiomatic. so feel to rewrite the rest!)

PR #6368, opened 2026-05-08
Updates Jina note to clarify that external inference is also supported

kosabogi:
Updates the recently added Jina note to clarify that Jina models are available not only through EIS, but also through external inference providers.


seanhandley (approved):

lcawl (approved):
LGTM, thanks!

alaudazzi (approved):
LGTM

PR #6375, opened 2026-05-08
Typo error as instead of `Percentage`, the term `Parentage` is written on Session View page.

divyaaghi-qasource:
**Documentation links**
Doc Link: https://www.elastic.co/docs/solutions/security/investigate/session-view

**Description**
Typo error as instead of `Percentage`, the term `Parentage` is written on Session View page.

**Screenshot:**

rt18

**Which documentation set(s) does this bug apply to?**
ESS and serverless

**Release version**
9.4

**Testing environment**
Build deployed from the production cloud environment: https://cloud.elastic.co/

### DIFF
--- a/solutions/security/investigate/session-view.md
+++ b/solutions/security/investigate/session-view.md
@@ -13,7 +13,7 @@ products:
 
 # Session View [security-session-view]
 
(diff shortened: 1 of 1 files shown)


natasha-moore-elastic:
Hi @divyaaghi-qasource, in this case "parentage" is correct – it refers to the fact that processes are shown as a "parent-child" relationship.

PR #6383, opened 2026-05-08
Move the bootstrap checks from system config to settings config

### LINKED ISSUE #3140: Rehome Elasticsearch bootstrap checks section
As identified when analyzing [this issue](https://github.com/elastic/docs-content/issues/3007#issuecomment-3311730535), the bootstrap checks don't belong to the OS System configuration that needs to be performed before installing Elasticsearch.

My suggestion is:

> As those are descriptive and also not related with OS configuration, I'd suggest to place them directly under `Deploy an Elasticsearch cluster`, at the same level as all the installation docs and Important system configuration. Probably after Important system configuration and before the install docs, as it might be a good reading for a user before the actual installation.

yetanothertw:
## Summary

Fixes #3140

* Added a Bootstrap checks item at the top of the [Important settings configuration](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/important-settings-configuration) page. I've included a short explanation that checks cover ES/JVM and OS limits, with links to [Bootstrap checks](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/bootstrap-checks) and [Important system configuration](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/important-system-configuration) (host limits).

* Updated the [Important system configuration](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/important-system-configuration) to remove the bootstraps section and replaced it with an admonition right after the checklist because I wanted to preserve some of the context that explains that OS limits can often fail checks in prod mode.

## Note to the reviewer

The build is failing because of a link that needs to updated in https://github.com/elastic/elasticsearch/pull/148641 which can only be merged once this PR has been merged and created that new link. It's kind of a chicken and egg situation, so you might need to build locally to review. 

## Generative AI disclosure

- [x] Yes  
- [ ] No  

Yes, I used Claude in Auto mode to find relevant links that needed updating.


deploy-manage/deploy/self-managed/important-settings-configuration.md
```diff
@@ -26,6 +27,18 @@ products:
...
 
+## Bootstrap checks [bootstrap-checks]
+
+When each node starts, {{es}} runs [bootstrap checks](/deploy-manage/deploy/self-managed/bootstrap-checks.md) against {{es}} settings, JVM configuration, and the host operating system to confirm they meet minimum safety requirements. If a check fails, you can resolve it by updating the cluster or JVM settings described on this page (for example networking or discovery settings), by adjusting the server as described in [Important system configuration](/deploy-manage/deploy/self-managed/important-system-configuration.md) (for example file descriptors or `vm.max_map_count`), or both.
+
+### Development mode vs. production mode
+
+By default, {{es}} assumes that you are working in development mode. If a bootstrap check fails in development mode, {{es}} writes a warning to the log file but the node still starts.
+
+When you configure settings such as [`network.host`](#network.host) so that {{es}} enters production mode, failed bootstrap checks become startup exceptions and prevent the node from starting. That reduces the risk of data loss from a misconfigured cluster or host.
```
eedugon: > When you configure settings such as [`network.host`](#network.host) so that {{es}} enters production mode

The previous sentence, with `so`, makes the reader wonder if they should already know about it. Also the `such as` feels a bit vague.

I'd suggest something more direct, even touching the first paragraph.

You can say smething like:

When `network.host` or `transport.host` [network settings](link-to-networking-settings) are set to a localhost address, {{es}} consider the node to be in development mode. In this mode, when a boostrap check fails, {{es}} writes a warning to the log file, but the node still starts.

When any of those settings are configured to a non-loopback address, {{es}} considers the node to be in production mode, and failed bootstrap checks become startup exceptions and prevent the node from starting.

Source: our great Colin back in 2017 (https://discuss.elastic.co/t/production-mode-vs-development-mode-in-elastic-kibana/91074)

deploy-manage/deploy/self-managed/important-settings-configuration.md
```diff
@@ -110,7 +123,7 @@ network.host: 192.168.1.10
 ```
 
 ::::{important}
-When you provide a value for `network.host`, {{es}} assumes that you are moving from development mode to production mode, and upgrades a number of system startup checks from warnings to exceptions. See the differences between [development and production modes](important-system-configuration.md#dev-vs-prod).
+When you provide a value for `network.host`, {{es}} assumes that you are moving from development mode to production mode, and upgrades a number of system startup checks from warnings to exceptions. Refer to [bootstrap checks](#bootstrap-checks) for more information.
```
eedugon, line 126: This aligns with with my previous comment.... but what is the purpose of this important note if you have added the same information just before in the same doc?

The note talks about network.host, and the section is about path-settings. I'm lost

deploy-manage/deploy/self-managed/important-system-configuration.md
```diff
@@ -23,6 +23,12 @@ The following settings **must** be considered before going to production:
 * [](/deploy-manage/deploy/self-managed/executable-jna-tmpdir.md): Ensure JNA and native libraries can execute from a temp path that is not mounted `noexec` (Linux only).
 * [](/deploy-manage/deploy/self-managed/system-config-tcpretries.md): Lower `net.ipv4.tcp_retries2` so node and network failures are detected sooner than the Linux default (Linux only).
 
+::::{admonition} How these limits are enforced
+This page lists operating system limits you must set before {{es}} serves production traffic. {{es}} verifies many of these expectations through [bootstrap checks](/deploy-manage/deploy/self-managed/bootstrap-checks.md) at node startup. In production mode, a failed check stops the node from starting rather than only logging a warning.
+
+For what bootstrap checks are and how development mode differs from production mode, refer to [Important settings configuration](/deploy-manage/deploy/self-managed/important-settings-configuration.md#bootstrap-checks).
+::::
+
```
eedugon, line 29: Having 2 links to undertand bootstrap checks doesn't feel natural.

LINK 1:

> {{es}} verifies many of these expectations through [bootstrap checks](/deploy-manage/deploy/self-managed/bootstrap-checks.md) at node startup

LINK 2 in the next paragraph, to a different page but to the boostrap-checks section

> For what bootstrap checks are and how development mode differs from production mode, refer to [Important settings configuration](/deploy-manage/deploy/self-managed/important-settings-configuration.md#bootstrap-checks).

IMO, the first link should also explain `what bootstrap checks are and how development mode differs from production mode`, so we wouldn't need a second link.

We need to review a bit the relation, scope and intention regarding bootstraps checks in the 3 docs:
- bootstrap checks doc
- important system config doc
- important settings config doc

eedugon (commented):
Note that the preview is not working yet.

I'm totally ok with including bootstrap checks information in the important settings configuration doc, but I shared some comments about them.

deploy-manage/deploy/self-managed/important-settings-configuration.md
yetanothertw: Not sure I'm following you here. 
This admonition already existed on the page, so I've updated the link that was pointing to the system settings page so that it now points to the new section I created in the settings configuration (not the existing **Path settings** section):
image

This may be confusing because the preview won't build because a link needs to be updated in the es repo (https://github.com/elastic/elasticsearch/pull/148641) but that can only be merged once this PR is merged. Kind of a chicken and egg situation. 
Building locally would probably help to alleviate this confusion.

yetanothertw: this is what the updated admonition looks like in context: 
image


The reason it's linking to the section that's included at the top of the same page is for situations in which people just skip directly to the **Network host settings** section and miss the **Bootstrap checks** section. Users don't tend to read a doc cover to cover or sequentially, so I think that admonition is an extra insurance layer that the info isn't missed

deploy-manage/deploy/self-managed/important-system-configuration.md
yetanothertw: Okay, I've removed the second link and we now point only to the [Bootstrap checks ](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/bootstrap-checks) page. 

My intention with this note was to conceptually link the system configuration limits that we set in the System config section/pages to how they're used in the bootstrap checks at node startup. 

Is there anything else you think we need to add/mention here? @eedugon

deploy-manage/deploy/self-managed/important-settings-configuration.md
yetanothertw: I've updated this in a new commit, thank you

deploy-manage/deploy/self-managed/important-settings-configuration.md
eedugon: > The note talks about network.host, and the section is about path-settings. I'm lost

sorry, that was completely wrong, I didn't see the big jump in lines when reviewing the content 🤦 

Yeah, the note makes sense, sorry for the noise.

deploy-manage/deploy/self-managed/important-system-configuration.md
eedugon: All good! I still see both links, but no worries.
It's just that needing two links to learn about bootstrap checks don't feel natural to me.

eedugon (approved):
LGTM!

deploy-manage/deploy/self-managed/important-system-configuration.md
yetanothertw: That's odd, I think you may have gremlins in your system messing with your github today 😄 

This is what the newly added admonition looks like (one link only):
Image

eedugon: I was probably looking at a specific commit.... ( 🤦 )^2 
:-D

PR #6402, opened 2026-05-11
Mark Prometheus remote write docs as preview for serverless

felixbarny:
The Prometheus remote write endpoint page already carried a preview designation for the stack, but the front matter did not include serverless. That meant readers on Elastic Serverless would not see the same preview labeling as on self-managed or Elastic Cloud Hosted. This change adds `serverless: preview` alongside `stack: preview` in `tsds-ingest-prometheus-remote-write.md` so the feature tier is shown consistently across deployment types.


yetanothertw (approved):
That's great, thank you! 🚢

PR #6455, opened 2026-05-13
Document cases.pushCases workflow step

### LINKED ISSUE #6246: [Internal]: New cases worfklow step `cases.pushCases`
### Description

Adds a new cases workflow step: `cases.pushCases`. It accepts a list of case ids that will then be pushed to their external connector.

### Resources

https://github.com/elastic/kibana/pull/267539

### Which deployment methods does this change impact?

### Feature differences

_No response_

### What Elastic Stack release is this request related to?

9.5

### Serverless release

The week of May 4th

### Collaboration model

The documentation team will create the first draft

### Point of contact.

**Main contact:** @janmonschke 

**Stakeholders:** @melissaburpo

benironside:
Closes #6246.

Adds reference docs for the new `cases.pushCases` workflow step, which pushes a list of cases to their configured external connector. Available in 9.4. Source: [Kibana #267539](https://github.com/elastic/kibana/pull/267539).

## AI Disclosure

Made with Claude Opus 4.7 and Cursor.


janmonschke:
> Does the optional top-level push-case: true config do anything on this step? (Likely a no-op, since the step is the push.)

That config is not available for this step since it does not make sense in this context.

> Is fail-fast across the parallel push intentional? (One bad case in case_ids errors the whole step, rather than partial-success like cases.getCases.)

Yes, in this case the action failed and would need to be retried.

explore-analyze/workflows/steps/cases.md
```diff
@@ -596,6 +599,47 @@ Remove an observable from a case.
...
+```
+
+% Ben Ironside Goldstein, 2026-05-12: Two items for SME confirmation before merge.
+% 1. The optional top-level `push-case: true` config is technically inherited via the shared
+%    cases step config schema. Its behavior on this step is unconfirmed (likely a no-op since
+%    the step is the push). If confirmed no-op, consider noting it inline.
+% 2. The handler pushes cases in parallel and returns `{ error }` on the first failure (fail-fast)
+%    rather than the `{ errors: [...] }` partial-success pattern used by `cases.getCases`.
+%    Confirm this is intentional behavior to document.
+
```
benironside:
```suggestion
```

janmonschke (approved):

nastasha-solomon (approved):
Nice!

PR #6539, opened 2026-05-18
Clarify remote-storage concerns

DaveCTurner:
Some users believe their remote storage has adequate performance because
they can achieve very high raw throughput. This commit clarifies that
the throughput is rarely the issue, and instead they must focus on
I/O latency when choosing a storage architecture.

### DIFF
--- a/deploy-manage/production-guidance/optimize-performance/indexing-speed.md
+++ b/deploy-manage/production-guidance/optimize-performance/indexing-speed.md
@@ -147,9 +147,9 @@ deployment:
   ece: all
 ```
 
-Directly-attached (local) storage generally performs better than remote storage because it is simpler to configure well and avoids communications overheads.
+{{es}} clusters using directly-attached (local) storage generally perform better than those using remote storage. This is typically because direct storage achieves lower latency on I/O operations. Remote storage can often achieve very high throughput, but this is less important to most {{es}} workloads than the latency on individual operations.
 
-Some remote storage performs very poorly, especially under the kind of load that {{es}} imposes. However, with careful tuning, it is sometimes possible to achieve acceptable performance using remote storage too. Before committing to a particular storage architecture, benchmark your system with a realistic workload to determine the effects of any tuning parameters. If you cannot achieve the performance you expect, work with the vendor of your storage system to identify the problem.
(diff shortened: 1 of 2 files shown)


deploy-manage/production-guidance/optimize-performance/indexing-speed.md
```diff
@@ -147,9 +147,9 @@ deployment:
   ece: all
 ```
 
-Directly-attached (local) storage generally performs better than remote storage because it is simpler to configure well and avoids communications overheads.
+{{es}} clusters using directly-attached (local) storage generally perform better than those using remote storage. This is typically because direct storage achieves lower latency on I/O operations. Remote storage can often achieve very high throughput, but this is less important to most {{es}} workloads than the latency on individual operations.
```
yetanothertw: Suggesting a rewrite for clarity and conciseness, flipping the focus on the direct storage benefit:

```suggestion
{{es}} clusters using directly-attached (local) storage generally perform better than those using remote storage. Direct storage typically provides lower latency for I/O operations, which is more critical for most {{es}} workloads than the high throughput that remote storage can often achieve.
```

deploy-manage/production-guidance/optimize-performance/search-speed.md
```diff
@@ -77,9 +77,9 @@ deployment:
   ece: all
 ```
 
-Directly-attached (local) storage generally performs better than remote storage because it is simpler to configure well and avoids communications overheads.
+{{es}} clusters using directly-attached (local) storage generally perform better than those using remote storage. This is typically because direct storage achieves lower latency on I/O operations. Remote storage can often achieve very high throughput, but this is less important to most {{es}} workloads than the latency on individual operations.
```
yetanothertw: Similar to the suggestion above:
```suggestion
{{es}} clusters using directly-attached (local) storage generally perform better than those using remote storage. Direct storage typically provides lower latency for I/O operations, which is more critical for most {{es}} workloads than the high throughput that remote storage can often achieve.
```

yetanothertw (approved):
Thanks for clarifying this distinction! 🪴 

I've left a small rewrite suggestion for your consideration.

DaveCTurner:
Thanks, sounds good

Issue #1468, opened 2025-05-23
Missing information on Add Docker metadata


quarky42:
Can you please point me to where the actual metadata values are documented or please add the metadata values that are accessible when using "Add docker metadata" feature?

Trial and error is incredibly time consuming and clear documentation is extremely valuable for feature adoption.

I specifically am trying to figure out how to add a dynamic field whose name is determined by the docker compose service name or the docker swarm service name as our services run on both compose and swarm.

quarky42:
No hard feelings.  I'll take a late response over no response any day and my sincere thanks for taking a moment to see if there is still any value here.  It might be OBE, but I appreciate that you're still checking. 

I also thank you for being understanding / gentle that my memory of some of the circumstances surrounding this request has faded.

eedugon:
my 2 cents to this issue with the focus on docs improvements:

- To avoid confusion maybe we need to clarify that `add_docker_metadata` is used to retrieve information from the docker environment and it cannot be used to customize or add information.

- We need to link better between the [provider](https://www.elastic.co/docs/reference/fleet/docker-provider) doc and the processor doc, because the provider uses the processor by default.

- On top of this we also have the [autodiscover doc](https://www.elastic.co/docs/reference/beats/filebeat/configuration-autodiscover). Autodiscover relies on the provider, which relies on the processor. In general the information is spread at those 3 levels. Autodiscover allows creating different inputs configuration depending on the information provided by the provider+processor.

> Can you please point me to where the actual metadata values are documented or please add the metadata values that are accessible when using "Add docker metadata"

The metadata fields are the ones shared by @vishaangelova , maybe we have to clarify how we are presenting the info:
- The [provider doc](https://www.elastic.co/docs/reference/fleet/docker-provider) shows the fields retrieved by the processor.
- Also the filebeat doc shared by Visha, but the provider doc has probably higher importance.
- For the "values" of those fields we have the description column, but it's true that the `labels` might deserve a better description of how exactly the labels are stored (they are `de-dotted` by default if I'm not mistaken).

The [kubernetes provider doc](https://www.elastic.co/docs/reference/fleet/kubernetes-provider) is much richer and longer. Not sure if the docker one just doesn't allow any customization or it's just missing. For kubernetes labels there are features (dedot vs not dedot, etc) that can be used and affect labels metadata fields. Not sure if that's just missing in docker or we don't support it for docker.

We would need some help from product team to clarify all that in order to improve our docs.

[...]

PR #6573, opened 2026-05-19
document add_docker_metadata fields and fix labels.dedot default

alexandra5000:
## Summary

Addresses [issue #1468](https://github.com/elastic/docs-content/issues/1468) — users couldn't find what metadata fields `add_docker_metadata` actually adds, and were getting confused by the undocumented label key transformation.

- **`add_docker_metadata-processor.md`**: replaces the "Container ID / Name / Image / Labels" bullet list with a proper field table showing actual field names (`container.id`, `container.name`, `container.image.name`, `container.labels.<key>`). Fixes the `labels.dedot` default from `false` → `true` (source code has `DeDot=true`). Adds a "Working with labels" section with a Compose/Swarm example and a note that dynamic field names are not supported.
- **`docker-provider.md`**: adds a Provider configuration section (`host`, `ssl`, `cleanup_timeout`), a variables table, and a "Working with labels" subsection explaining the two label representations. Adds cross-links to the processor doc.


vishaangelova:
@alexandra5000 could we have a technical review from an SME on the subject?

alexandra5000:
@vishaangelova yup, I reached out to @belimawr on Slack, I'm waiting for his review

reference/fleet/add_docker_metadata-processor.md
```diff
@@ -22,12 +22,31 @@ The `add_docker_metadata` processor annotates each event with relevant metadata
...
-* Container ID
-* Name
-* Image
-* Labels
+| Field | Description |
+| --- | --- |
+| `container.id` | Container ID |
+| `container.name` | Container name |
+| `container.image.name` | Container image name |
+| `container.labels.<key>` | One field per Docker label. By default, dots (`.`) in label keys are replaced with underscores (`_`). Refer to [`labels.dedot`](#labels-dedot-behavior) for more information. |
```
belimawr: I'm not sure `labels.dedot` is the correct text for this link. I followed the link expecting to see the documentation for `lablels.dedot`, but landed on a section about "Working with labels".

I'd change it to
```suggestion
| `container.labels.<key>` | One field per Docker label. By default, dots (`.`) in label keys are replaced with underscores (`_`). Refer to [working with labels](#labels-dedot-behavior) section for more information. |
```

reference/fleet/add_docker_metadata-processor.md
```diff
@@ -22,12 +22,31 @@ The `add_docker_metadata` processor annotates each event with relevant metadata
...
+| `com.docker.compose.service` | `container.labels.com_docker_compose_service` |
+| `com.docker.swarm.service.name` | `container.labels.com_docker_swarm_service_name` |
+
+To preserve the original key names with dots, set `labels.dedot` to `false`. In that case, the labels are stored as nested objects (for example, `container.labels.com.docker.compose.service`).
+
+::::{note}
+Creating a field with a name that is dynamically determined at runtime (for example, using the value of a label as a field name) is not supported by this processor or by `add_fields`. To react to label values, use [conditions-based autodiscover templates](/reference/fleet/conditions-based-autodiscover.md) instead.
+::::
+
+The Docker provider exposes the same labels as variables for configuration templating (`${docker.container.labels.com.docker.compose.service}`). The provider preserves dots in the key for variable references, regardless of the `labels.dedot` setting. For more information, refer to [Docker provider](/reference/fleet/docker-provider.md).
```
belimawr: Maybe this is my engineer brain who knows the internals of Filebeat talking, take this with a grain of salt or just ignore it ;)

I don't like the 'preserves' verb here.... Variable resolution is not related to the fields defined in the events, but this paragraph makes it sound like otherwise.
```suggestion
The Docker provider exposes the same labels as variables for configuration templating (`${docker.container.labels.com.docker.compose.service}`). The provider always references keys in the dotted format. For more information, refer to [Docker provider](/reference/fleet/docker-provider.md).
```

The Docker provider runs way before any processor, it uses its variables to render an input configuration, once the input is running, any defined processor will enrich/modify the ingested events, that's when the `add_docker_meteadata` processor actually runs.

reference/fleet/docker-provider.md
```diff
@@ -11,7 +11,34 @@ products:
 
 # Docker provider [docker-provider]
 
-Provides inventory information from Docker. The available dynamic variables are:
+Provides inventory information from Docker. {{agent}} uses the Docker provider to automatically discover containers and build input configurations. It is also responsible for enriching events with container metadata through the [`add_docker_metadata` processor](/reference/fleet/add_docker_metadata-processor.md), which it applies by default to all inputs it generates.
```
belimawr: The Docker provider and the `add_docker_metadata` processor are totally independent pieces of software. They even run on different process.

Expanding on what I mentioned [before](https://github.com/elastic/docs-content/pull/6573/changes#r3288923488): The Docker provider runs inside Elastic Agent, it is responsible for discovering containers and "rendering" (technically the provider variables are used to render the input configuration, but let's keep it simple) the input configuration. Then the Elastic Agent send this configuration to a Beat or Otel collector, both are different process. Once the Beat/Otel collector starts running, it will instantiate its own (if configured to do so), `add_docker_metadata` processor to enrich the events it is ingesting.

What also happens, is that the input configuration uses the `add_fields` processor with values resolved from the Docker provider to enrich the final event, here  is an example of an input generated by the Docker integration, note that there is no `add_docker_metadata` processor but the labels are added via `add_fields` processor.
```
    - data_stream:
        namespace: default
      id: filestream-docker-e20c8b8c-5558-4d91-9d07-9029aa798f01-docker-30f4b47c10cdb2a794c3f2cf996e2f837cdb60cf071c4b0e21cf74bc1fe80877
      meta:
        package:
            name: docker
            policy_template: docker
            release: ga
            version: 2.15.1
      name: docker-1
      original_id: filestream-docker-e20c8b8c-5558-4d91-9d07-9029aa798f01
      package_policy_id: e20c8b8c-5558-4d91-9d07-9029aa798f01
      policy:
        revision: 2
      processors:
        - add_fields:
            fields:
                id: 30f4b47c10cdb2a794c3f2cf996e2f837cdb60cf071c4b0e21cf74bc1fe80877
                image.name: mingrammer/flog
                labels:
                    a_doted_label: it works
                    foo: bar
                name: flog
            target: container
      revision: 1
      streams:
        - data_stream:
            dataset: docker.container_logs
            elasticsearch:
                dynamic_dataset: true
                dynamic_namespace: true
            type: logs
          id: docker-container-logs-flog-30f4b47c10cdb2a794c3f2cf996e2f837cdb60cf071c4b0e21cf74bc1fe80877
          parsers:
            - container:
                format: docker
                stream: all
          paths:
            - /var/lib/docker/containers/30f4b47c10cdb2a794c3f2cf996e2f837cdb60cf071c4b0e21cf74bc1fe80877/*-json.log
      type: filestream
```

Last, but not least, Elastic Agent always adds (well, it's hardcoded inside Beats) the `add_docker_metadata` processor to all Beats, but the user has no control over its configurations. I believe this will change in 9.5.0.

Given all that, here is how I'd change this paragraph:
```suggestion
Provides inventory information from Docker. {{agent}} uses the Docker provider to automatically discover containers and build input configurations.
```

reference/fleet/docker-provider.md
```diff
@@ -20,6 +47,19 @@ Provides inventory information from Docker. The available dynamic variables are:
 | `docker.container.image.name` | `string` | Image of the container |
 | `docker.container.labels` | `object` | Labels of the container |
 
+### Working with labels [_working_with_labels]
+
+Label keys are available as variables using their original names, including dots. For example, for a container with the label `com.docker.compose.service=redis`, you can reference it as `${docker.container.labels.com.docker.compose.service}`.
+
+However, when the Docker provider enriches events using the `add_docker_metadata` processor, dots in label keys are replaced with underscores by default (`labels.dedot: true`). This means that in the resulting {{es}} document, the same label is stored as `container.labels.com_docker_compose_service`.
```
belimawr: I'm not sure this behaviour can be configured in the Docker provider. I'd also remove all references to the `add_docker_metadata` processor.

```suggestion
However, when the Docker provider enriches events, dots in label keys are replaced with underscores. This means that in the resulting {{es}} document, the same label is stored as `container.labels.com_docker_compose_service`.
```

reference/fleet/docker-provider.md
```diff
@@ -20,6 +47,19 @@ Provides inventory information from Docker. The available dynamic variables are:
...
+
+| Context | Label key format |
+| --- | --- |
+| Variable reference in the configuration template (`${...}`) | Dots preserved — `docker.container.labels.com.docker.compose.service` |
+| Field in the event/{{es}} document | Dots replaced by `_` — `container.labels.com_docker_compose_service` |
+
+For more details on the label behavior, refer to [Working with labels](/reference/fleet/add_docker_metadata-processor.md#labels-dedot-behavior) in the `add_docker_metadata` processor documentation.
```
belimawr: I'm not sure how much of this can be influenced. I can do some more testing/investigation later.

```suggestion
```

reference/fleet/add_docker_metadata-processor.md
alexandra5000: You're right, I think I rewrote it but didn't change the link. Thanks for catching that!

reference/fleet/add_docker_metadata-processor.md
alexandra5000: It's better to be precise! That's one of the reasons why we need the review - for the engineering 🧠

reference/fleet/docker-provider.md
alexandra5000: I now see that I conflated some processes and also overestimated how much users can configure. This is a very nice explanation, thank you for providing the context I needed!

reference/fleet/add_docker_metadata-processor.md
```diff
@@ -22,12 +22,31 @@ The `add_docker_metadata` processor annotates each event with relevant metadata
...
+| Field | Description |
+| --- | --- |
+| `container.id` | Container ID |
+| `container.name` | Container name |
+| `container.image.name` | Container image name |
+| `container.labels.<key>` | One field per Docker label. By default, dots (`.`) in label keys are replaced with underscores (`_`). Refer to the [working with labels](#labels-dedot-behavior) section for more information. |
+
+### Working with labels [labels-dedot-behavior]
+
+Docker label keys often contain dots, for example `com.docker.compose.service`. By default (`labels.dedot: true`), the processor replaces dots with underscores when writing labels to the event:
```
theletterf:
```suggestion
Docker label keys often contain dots, for example `com.docker.compose.service`. By default, the processor replaces dots with underscores when writing labels to the event:
```

theletterf (approved):
LGTM, just a nit

reference/fleet/add_docker_metadata-processor.md
eedugon: ++ to that

reference/fleet/add_docker_metadata-processor.md
```diff
@@ -22,12 +22,31 @@ The `add_docker_metadata` processor annotates each event with relevant metadata
...
+| Original Docker label | Field name in event |
+| --- | --- |
+| `com.docker.compose.service` | `container.labels.com_docker_compose_service` |
+| `com.docker.swarm.service.name` | `container.labels.com_docker_swarm_service_name` |
+
+To preserve the original key names with dots, set `labels.dedot` to `false`. In that case, the labels are stored as nested objects (for example, `container.labels.com.docker.compose.service`).
+
+::::{note}
+Creating a field with a name that is dynamically determined at runtime (for example, using the value of a label as a field name) is not supported by this processor or by `add_fields`. To react to label values, use [conditions-based autodiscover templates](/reference/fleet/conditions-based-autodiscover.md) instead.
+::::
```
eedugon, line 47: Very good note. I think we should include it in the `add_kubernetes_metadata` processor too.

One caveat of the note is that we are linking to the doc about kubernetes autodiscovery, not docker autodiscovery (but docker autodiscovery doc doesn't exist at `fleet` level, only at individual beats documentation)

reference/fleet/docker-provider.md
```diff
@@ -11,7 +11,34 @@ products:
 
 # Docker provider [docker-provider]
 
-Provides inventory information from Docker. The available dynamic variables are:
+Provides inventory information from Docker. {{agent}} uses the Docker provider to automatically discover containers and build input configurations.
```
eedugon: Should we mention also in the docker provider introduction that it relies on `add_docker_metadata` with a link to the other doc?

eedugon (approved):
Looks very good, all the changes really improve the content.

I'll share some concerns for your consideration, as maybe it makes sense to address them in a separate issue, or we could ignore them.

- The relation between the provider, autodiscover functionality, and the processor `add_docker_metadata` has improved but it's still not ideal. I've suggested to add a link from the provider to add_docker_metadata, which will improve this further, but the relation with autodiscover functionality is still weak.

- We don't have any doc showing docker based autodiscover in fleet docs, the autodiscovery docs we have are about kubernetes autodiscovery with the kubernetes provider. The concepts are very similar, but there's no such document (maybe something to address in a separate issue).

- Maybe we could add something about docker autodiscover in the docker provider doc (the original beats docs cover the provider + autodiscover in the same doc, for example https://www.elastic.co/docs/reference/beats/filebeat/configuration-autodiscover).

reference/fleet/docker-provider.md
alexandra5000: I actually had this in an earlier version of the PR, but @belimawr pointed out that the Docker provider and the `add_docker_metadata` processor are independent pieces of software that run in separate processes (the provider runs inside Elastic Agent to discover containers and render input configurations, while `add_docker_metadata` runs inside Beats/OTel collector after the input is started). Saying the provider "relies on" or "applies" the processor would therefore be technically inaccurate.

That said, a neutral cross-reference might still be useful for readers who land on one page and want to find the other. Would you be ok with something like:

> For information on how container metadata is added to events, refer to [`add_docker_metadata` processor](/reference/fleet/add_docker_metadata-processor.md).

reference/fleet/add_docker_metadata-processor.md
alexandra5000: I added the note to the kubernetes processor page like you suggested. 

Regarding the link, I changed it to dynamic-input-configuration.md, although you're right and what would really work here would be a docker autodiscovery doc on fleet level. Since we don't have that (yet) I tried to find the next best thing that could be useful to the users in this context.

alexandra5000:
Thanks a lot for your input @eedugon! I applied what I could and opened a new issue for the bigger changes that are out of scope for this PR: https://github.com/elastic/docs-content/issues/6797 I'll try to pick it up when I have the time (or maybe someone else from the team will).

belimawr:
@alexandra5000 let me know once the PR is ready for a final review.

alexandra5000:
> @alexandra5000 let me know once the PR is ready for a final review.

@belimawr Go ahead - I didn't change much, you can just do a quick check on the two latest commits. Thanks!

belimawr (approved):
LGTM

reference/fleet/docker-provider.md
eedugon: thanks for the explanation @alexandra5000 and @belimawr , it helps a lot!

Anyway, for reference:

> the provider runs inside Elastic Agent to discover containers and render input configurations

If I'm not mistaken, when the provider, and through autodiscover, renders the `beats inputs`, these inputs include `add_docker_metadata` by default. That's the relation I tried to mean with `rely on`.

Anyway it's not a big deal, thanks a lot!

belimawr: @eedugon, no the input themselves do not include the `add_docker_metadata` processor, however, Elastic Agent always enables a global `add_docker_metadata` processor in all Beats.

So the processor will run, but not as part of the input configuration. The key difference here is that no integration can configure/influence this processor.

PR #6580, opened 2026-05-20
[Connectors] Document defaultModel for Azure OpenAI

alaudazzi:
This PR adds a new optional step 7 to document the default model field with a note explaining that:
- Required only for APIM and other proxy endpoints that cannot infer the model from the deployment URL
- Standard Azure OpenAI deployments do not need it
- Behaves as a fallback (only injected when the request body doesn't already specify a model)

Closes https://github.com/elastic/docs-content-internal/issues/985

## Notes for the reviewers

1. This fix was backported to `8.19.12, 8.19.13, 9.2.5, 9.2.7, 9.3.1,` and `9.4.0.` I guess we don't need to use any applies_to in this file, and once merged, we backport this PR to `8.19`.  To be confirmed.
2. Frontmatter products: current frontmatter covers only security, but the Observability AI Assistant is also affected. Do we need to add observability to the frontmatter?

## Generative AI disclosure

- [x] Yes  
- [ ] No


kosabogi (approved):
LGTM!

benironside (approved):

PR #6598, opened 2026-05-20
Refine ECE SUSE host configuration page

marciw:
## Summary

Created by Claude

Fixes https://github.com/elastic/docs-content-internal/issues/24

Preview of main changes to review: [Configure a SuSE host](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/6598/deploy-manage/deploy/cloud-enterprise/configure-host-suse)

- Rewrite the SUSE host configuration page: drop the unsupported Docker 24.0.7 examples, flag SLES 12 SP5 EOL (October 2024), drop SLES 12-specific install commands, and label the remaining Docker command as an example to cross-check against the [Support matrix](https://www.elastic.co/support/matrix#elastic-cloud-enterprise).
- Reword the intros on all three sibling pages (RHEL, SUSE, Ubuntu) so they diverge enough to be indexed independently by search engines (per eedugon's "extra improvement" bullet in the issue).
- Drop the "Think like a minimalist / Cover your bases" intro from the SUSE page per kunisen's [comment on the issue](https://github.com/elastic/docs-content-internal/issues/24#issuecomment-2965321834).

## :rotating_light: SME validation needed before merge

All commands have been tested in SLES 15 SP4 by @eedugon 

- [X] Yes  
- [ ] No  
Tools and models used: Claude Code CLI default


deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
```diff
@@ -12,7 +12,7 @@ products:
 
 # Configure a RHEL host [ece-configure-hosts-rhel-centos]
 
-The following instructions show you how to prepare your hosts on Red Hat Enterprise Linux 8 (RHEL 8), 9 (RHEL 9), and Rocky Linux 8 and 9.
+Red Hat Enterprise Linux 8 and 9, along with Rocky Linux 8 and 9, run {{ece}} on Podman rather than Docker. Use these steps to install Podman with `dnf`; configure the Podman socket so the standard `docker` CLI works against it; and apply the SELinux, XFS, and kernel tuning that ECE expects on Red Hat-family hosts.
```
eedugon:
```suggestion
Red Hat Enterprise Linux 8 and 9, along with Rocky Linux 8 and 9, run {{ece}} (ECE) on Podman rather than Docker. Use these steps to install Podman and configure Podman, and apply the SELinux, XFS, and kernel tuning that ECE expects on Red Hat-family hosts.
```

I'd omit the `dnf` and socket part. A bit too low level for the intro.

deploy-manage/deploy/cloud-enterprise/configure-host-ubuntu.md
```diff
@@ -12,7 +12,7 @@ products:
 
 # Configure an Ubuntu host [ece-configure-hosts-ubuntu]
 
-The following instructions show you how to prepare your hosts on Ubuntu.
+Use the steps on this page to prepare an Ubuntu server for {{ece}} (ECE): install Docker from the upstream Docker `apt` repository, set up an XFS volume for cluster data, tune kernel and systemd parameters for production workloads, and pin the Docker version so unattended upgrades don't break ECE.
```
eedugon:
```suggestion
Use the steps on this page to prepare an Ubuntu server for {{ece}} (ECE): install and configure Docker, set up an XFS volume for ECE data, tune kernel and systemd parameters for production workloads, and pin the Docker version to avoid unexpected upgrades.
```

eedugon (commented):
I've reviewed RHEL and Ubuntu small changes.
I'll test all SuSE commands and the new content in a lab environment and get back to you next week!

Thanks a lot for working on this issue!

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,90 @@ products:
...
-1. Remove Docker and previously installed podman packages (if previously installed).
+
+1. Remove Docker and any previously installed podman packages.
 
     ```sh
     sudo zypper remove -y docker docker-ce podman podman-remote
     ```
 
-2. Update packages to the latest available versions
+2. Update packages to the latest available versions.
```
eedugon:
```suggestion
1. Update packages to the latest available versions.
```

Let's start all numbered list items with 1., for easier maintenance.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,90 @@ products:
...
 
     ```sh
     sudo zypper refresh
     sudo zypper update -y
     ```
 
-3. Install Docker and other required packages:
+3. Install Docker and other required packages on SLES 15. The following command is an example of installing Docker {{ece-docker-version}}. To install a different Docker version, replace {{ece-docker-version}} with your preferred version from the [Support matrix](https://www.elastic.co/support/matrix#elastic-cloud-enterprise).
```
eedugon:
````suggestion
1. List the available versions of docker package in the repositories:

    ```sh
    suso zypper search -s -t package --match-exact docker
    ```

    From the output list make note of the version you want to install. Check the [Support matrix](https://www.elastic.co/support/matrix#elastic-cloud-enterprise) for the list of supported versions for your OS. If the latest available version is compatible, you can ignore the explicit version set in the following command.

1. Install Docker and other required packages on SLES 15. The following command is an example of installing Docker 25:
````

File: deploy-manage/deploy/cloud-enterprise/configure-host-suse.md [hunk 5]
```diff
@@ -12,95 +12,90 @@ products:
...
 
-    * For SLES 12:
+    ```sh subs=true
+    sudo zypper install -y curl device-mapper lvm2 net-tools docker={{ece-docker-version}}.*
```
eedugon:
```suggestion
    sudo zypper install -y curl device-mapper lvm2 net-tools docker=25.0.6_ce-150000.207.1 <1>
```

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,90 @@ products:
...
+    ```
```
eedugon, line 116:
````suggestion
    ```
    1. Replace `27.5.1_ce-150000.218.1` with the exact version to install, or just leave `docker` to install the latest available version of the docker package.
````

File: deploy-manage/deploy/cloud-enterprise/configure-host-suse.md [hunk 7]
```diff
@@ -12,95 +12,90 @@ products:
...
+
+    ::::{tip}
+    If `zypper` reports that the requested Docker version isn't available, make sure the SUSE **Containers Module** is enabled, or refer to [SUSE's documentation](https://documentation.suse.com/sles/15-SP6/html/SLES-all/cha-docker-installation.html) for adding the upstream Docker repository.
```
eedugon: That link looks broken

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,90 @@ products:
...
+    ::::
+
+    ::::{note}
+    Installation on SLES 12 SP5 is no longer covered here because SLES 12 SP5 is past general support end of life. If you're maintaining an existing SLES 12 SP5 deployment, install the last Docker version that SUSE shipped for SLES 12 SP5 and plan a migration to SLES 15.
+    ::::
+
+4. Set up the OS groups and add your user.
```
eedugon: Here we have a signifficant deviation in the Ubuntu, RHEL and SuSE guides (cc: @kunisen )

Originally we suggested to create a user called `elastic` to install ECE. It looks at some point we allowed to install ECE with a different user. Some guides (RHEL) still show the `elastic` user, while the ubuntu guide uses the current user but then it shows content (/etc/limits.conf) with the `elastic` user hardcoded).

I think we should try to align the 3 installation guides, and if we support both ways we should state it clearly, and probably suggest the creation of the `elastic` user with an optional step.

Update: I can confirm that we support a normal (not `elastic`) user, but anyway we should align the 3 guides and probably include a section to do specific user and group setup.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: Note that `docker=27.0*` (or another type of wildcard) doesn't work with zypper. The version needs to match exactly the package version to install.

I've included `docker=25.0.6_ce-150000.207.1`, as Docker 25 is the officially supported package for SLES15 based on the support matrix `today`.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: For example @marciw , this TIP is great on the system limits:

image

And for that TIP to make better sense, when we talk about the user and groups we should also give the option of creating the `elastic` user, and now that's missing.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -246,8 +237,6 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
     sudo install -o $USER -g elastic -d -m 700 /mnt/data/docker
     ```
 
-
-
 ## Configure the Docker daemon [ece-configure-docker-daemon-sles12]
 
 1. Edit `/etc/docker/daemon.json`, and make sure that the following configuration values are present:<br>
```
eedugon: This `/etc/docker/daemon.json` file is missing critical info we added recently on the ubuntu file:

image

In this case we need to simply combine all the content. The current content is valid and correct also, but we need to add those limits to the docker daemon.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -266,7 +255,7 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
     }
     ```
 
-2. The user installing ECE must have a User ID (UID) and Group ID (GID) of 1000 or higher. Make sure that the GID matches the ID of the `elastic`` group created earlier (likely to be 1000). You can set this using the following command:
+2. The user installing ECE must have a User ID (UID) and Group ID (GID) of 1000 or higher. Make sure that the GID matches the ID of the `elastic` group created earlier (likely to be 1000). You can set this using the following command:
```
eedugon: That step at this stage causes confusion, we should move it upwards, together with the creation of the user and group.

My suggestion is to have a single location for user, group and permissions setup.

The `usermod` mentioned here is probably needed because we need the user to have as `primary group`  the `elastic` group, but let's move it to the right place.

Another thing to add is a hint for the reader to get the GID of the `elastic` group. They could do that with `sudo cat /etc/group | grep elastic` for example (maybe there's a smoother way).

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -246,8 +237,6 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
     sudo install -o $USER -g elastic -d -m 700 /mnt/data/docker
     ```
 
-
```
eedugon, line 253: The step 7
> 7. If you [set up a new device with XFS](#ece-xfs-setup-sles12) earlier:

Is a bit contradictory, becuase if that was really done, both the mount and chown shouldn't be needed.

It feels like a mess, much better to just tell the user to ensure /mnt/data exist and with proper owner, group owner and permissions, and in case it's NOT correct, refer to the proper step, but write the config steps in a single location, and not twice.

Note that step 8 (/mnt/data/docker) is really needed.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -311,15 +300,15 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
+
     For non-{{es}} connections such as the proxy layer, consider reducing the following TCP keepalive parameters to detect stale network sessions and prevent firewalls from dropping silent connections:
     * `net.ipv4.tcp_keepalive_time`
     * `net.ipv4.tcp_keepalive_intvl`
     * `net.ipv4.tcp_keepalive_probes`
     :::
 
 
-    1. Ensure settings in /etc/sysctl.d/*.conf are applied on boot
+    1. Ensure settings in /etc/sysctl.d/*.conf are applied on boot:
```
eedugon: The following script is broken, as `/var/lib/cloud/scripts/per-boot/00-load-sysctl-settings` cannot be created if the base directory doesn't exist.

We should double check with devs if that's still needed (check the other OS guides to see if we have that). It's a directory related with `cloud-init`, but not all OS images have cloud-init isntalled, and we don't mention it on our docs :(

For the command to work we could just do a `sudo mkdir -p  /var/lib/cloud/scripts/per-boot/`, but that wouldn't mean that the script would run on every boot.

Hopefully we should be able to remove all that part, find an alternative, or tell users that `cloud-init` is a pre-requisite for this installation.

eedugon (changes requested):
Installation works fine with SLES15-SP4 and Docker 25, but not with newer SPs where Docker 25 is not available, and for some reason ECE install doesn't work in SuSE with Docker 27.

Also SP4 is close to EOL and it's difficult to find images, and newer SPs (SP7) don't include Docker 25 anymore in the package repositories. This has been raised to cloud team for their consideration. Ideally we should support newer SPs and Docker versions soon.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: In a similar approach as with user & group, probably creating the mount point & directories and setting permissions should be all together in its own separate section.

deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
marciw: agree! Fixed in 248d9f19f981c61708aed199cc97acab3a27d407

deploy-manage/deploy/cloud-enterprise/configure-host-ubuntu.md
marciw: 👍 fixed in 2550b484af6f48f957b5c221450cee63812f5cc2

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
marciw: leaving this one open so we can get input from SMEs

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
marciw: Fixed in 32578c3082ab381e0b7da516cac6d63ae93d61d1

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
marciw: fixed in 32578c3082ab381e0b7da516cac6d63ae93d61d1

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
marciw: fixed in 32578c3082ab381e0b7da516cac6d63ae93d61d1

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
marciw: fixed in 32578c3082ab381e0b7da516cac6d63ae93d61d1

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
marciw: fixed in 32578c3082ab381e0b7da516cac6d63ae93d61d1

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
marciw: fixed in 32578c3082ab381e0b7da516cac6d63ae93d61d1

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
marciw: fixed in 32578c3082ab381e0b7da516cac6d63ae93d61d1

marciw:
thanks for all the detailed comments, @eedugon! I think I've addressed everything except for the one I left open. Take a look when you get a chance.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,108 @@ products:
 
 # Configure a SUSE host [ece-configure-hosts-sles12]
 
-The following instructions explain how to prepare your hosts on SUSE Linux Enterprise Server 12 SP5 (SLES 12) or 15 (SLES 15).
+SUSE Linux Enterprise Server (SLES) hosts use `zypper` to install Docker. They also require manual setup of XFS quotas, since SLES doesn't ship XFS as the default file system. The steps on this page target SLES 15.
```
eedugon:
```suggestion
SUSE Linux Enterprise Server (SLES) hosts use `zypper` to install Docker. They also require manual setup of XFS quotas, since SLES doesn't ship XFS as the default file system. The steps on this page target SLES 15 SP4. 
```

Not ideal but we have to clarify our steps are for SP4, as they don't work with SP7 today (SP7 doesn't come with a supported Docker version).

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -313,9 +312,7 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
         sudo systemctl enable docker
         ```
 
-4. Recommended: Tune your network settings.
-
-    Create a `70-cloudenterprise.conf` file in the `/etc/sysctl.d/` file path that includes these network settings:
+1. For best results, tune your network settings. Create a `70-cloudenterprise.conf` file in `/etc/sysctl.d/` and include these settings:
```
eedugon: Tune network settings should belong to the previous section (Update the configuration settings or Update system configuration), as it's not related with Docker configuration.

Please address the other comments on these items before moving the content.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,108 @@ products:
...
+    ```
 
-    * For SLES 15:
+    1. Replace `25.0.6_ce-150000.207.1` with the exact package version from the repository listing. Note that zypper does not support wildcards (for example, `docker=25.*`). To install the latest available version, specify `docker` without a version number.
+
+    ::::{tip}
+    If `zypper` reports that the requested Docker version isn't available, make sure the SUSE **Containers Module** is enabled. Refer to the [SUSE documentation](https://documentation.suse.com/) for instructions on adding the upstream Docker repository.
+    ::::
+
+1. Set up the OS groups and add your user.
```
eedugon, line 40:
```suggestion
## Prepare the user account for ECE

The following commands assume that you are logged in as the non-root user account that will run ECE. We recommend using a dedicated `elastic` user, but you can also use an existing non-root user account with a UID greater than 1000.

1. Set up the OS groups and add your user.
```

Objectives of the suggestion:
- Separate user & group setup to a dedicated section.
- Tell (and recommend) readers that they can create a dedicated user account (i.e `elastic`) for ECE installation.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,108 @@ products:
...
-    ```sh
-     sudo groupadd elastic
-     sudo groupadd docker
-    ```
+        ```sh
+        sudo usermod -g <elastic_group_gid> $USER
+        ```
 
-2. Add the user to these groups:
+1. Stop the `nscd` service and prevent it from starting automatically (it can interfere with Elastic services):
```
eedugon: Let's move this item (nscd) somewhere else:

I'd suggest to move it to the "Update the configurations settings" / "System configuration" section.
Or place it at the end of "docker install" section, but not together with user & group configuration.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -224,40 +222,45 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
 
     ```json
     {
+      "default-ulimits": {
+        "nofile": {
+          "Name": "nofile",
+          "Soft": 1048576,
+          "Hard": 1048576
+        }
+      },
```
eedugon:
```suggestion
```

default-ulimits -> nofile are already present in line 273.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -329,20 +326,21 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
+
     For non-{{es}} connections such as the proxy layer, consider reducing the following TCP keepalive parameters to detect stale network sessions and prevent firewalls from dropping silent connections:
     * `net.ipv4.tcp_keepalive_time`
     * `net.ipv4.tcp_keepalive_intvl`
     * `net.ipv4.tcp_keepalive_probes`
     :::
 
 
-    1. Ensure settings in /etc/sysctl.d/*.conf are applied on boot
+    1. Ensure the settings in `/etc/sysctl.d/*.conf` are applied at startup:
```
eedugon:
```suggestion
    1. (Optional) Ensure `/etc/sysctl.d/*.conf` settings are applied at startup:
    
        This workaround requires `cloud-init` and is only needed if the settings are not automatically applied after a reboot.
```

eedugon (commented):
Things are looking much better, thanks a lot!
Sharing some extra comments for small movements and refinement. Afterwards I'll do a new set of testing and I might suggest small changes on the order of the directories and mount point configuration steps.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,108 @@ products:
 
 # Configure a SUSE host [ece-configure-hosts-sles12]
 
-The following instructions explain how to prepare your hosts on SUSE Linux Enterprise Server 12 SP5 (SLES 12) or 15 (SLES 15).
+SUSE Linux Enterprise Server (SLES) hosts use `zypper` to install Docker. They also require manual setup of XFS quotas, since SLES doesn't ship XFS as the default file system. The steps on this page target SLES 15. 
+
+Before installing, make sure to cross-check your SLES version and Docker version against the [Support matrix](https://www.elastic.co/support/matrix#elastic-cloud-enterprise). The commands shown on this page are examples. Substitute the versions you've identified in the support matrix.
```
kunisen:
```suggestion
Before installing, make sure to cross-check the compatible SLES version and Docker version combination against the [Support matrix](https://www.elastic.co/support/matrix#elastic-cloud-enterprise). The commands shown on this page are examples. Substitute the versions you've identified in the support matrix.
```

Good to explicitly mention "compatible" so that user can understand more on "why" they need to cross check.

deploy-manage/deploy/cloud-enterprise/configure-host-ubuntu.md
```diff
@@ -49,10 +49,10 @@ Install a compatible Docker version on Ubuntu.
       $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
     ```
 
-4. Install the correct version of the `docker-ce` package. The following is an example of installing Docker 27.0. If you decide to install a different Docker version, make sure to replace with the desired version in the commands below.
+4. Install the correct version of the `docker-ce` package. The following command is an example of installing Docker {{ece-docker-version}}. To install a different Docker version, replace {{ece-docker-version}} with your preferred version from the [Support matrix](https://www.elastic.co/support/matrix#elastic-cloud-enterprise).
 
-    ```sh
-    sudo apt update && sudo apt install -y docker-ce=5:27.0.* docker-ce-cli=5:27.0.* containerd.io
+    ```sh subs=true
+    sudo apt update && sudo apt install -y docker-ce=5:{{ece-docker-version}}.* docker-ce-cli=5:{{ece-docker-version}}.* containerd.io
```
kunisen: @eedugon ~will this `{{ece-docker-version}}` be accidentally replaced by someone if they think it should be an environment variable in docs? If so, I would suggest we use a different way to express this, such as `<ece-docker-version>` or whatever that won't cause doc team's confusion. WDYT?~

Scratch that, it's probably good to use variable to make things easier to maintain, anyways.

deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
```diff
@@ -12,7 +12,7 @@ products:
 
 # Configure a RHEL host [ece-configure-hosts-rhel-centos]
 
-The following instructions show you how to prepare your hosts on Red Hat Enterprise Linux 8 (RHEL 8), 9 (RHEL 9), and Rocky Linux 8 and 9.
+Red Hat Enterprise Linux 8 and 9, along with Rocky Linux 8 and 9, run {{ece}} (ECE) on Podman rather than Docker. Use these steps to install and configure Podman, and to apply the SELinux, XFS, and kernel tuning that ECE expects on Red Hat-family hosts.
```
kunisen: I think I would remove `Red Hat-family hosts` but instead use:
* Either ` these supported hosts`
* Or `RHEL-compatible distributions`

Because `Red Hat-family` is not an official way to say Rocky Linux by RedHat / Oracle.
This avoids using Red Hat as an umbrella adjective for a competitor's product.

WDYT? @marciw @eedugon

kunisen (approved):
Thanks. It looks great in general, but only a few things we may want to make more explicitly clear.

deploy-manage/deploy/cloud-enterprise/configure-host-ubuntu.md
eedugon: Yes, when possible we prefer to use variables, and they will be rendered in the docs not as variables, but with the actual values.

On SuSE 15 SP4 doc we can't, except if we create a new variable for SuSE. In the near future, if we align and start supporting a similar docker version across ALL OSs then we would use a single variable in the 3 docs.

deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
eedugon: I'm ok with your suggestion @kunisen , thank, you can directly suggest it if you like and become a contributor of this PR ;) I like `on RHEL-compatible distributions`.

eedugon:
> Thanks. It looks great in general, but only a few things we may want to make more explicitly clear.

Btw, @kunisen , at a later stage we will try to align the OS preparation guides of the 3 OSs we support, as today each of the docs have a different structure, headings and order of steps. Our plan is to make this one the reference today in terms of clarity, structure and order, and then align the other 2 a bit more.

deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
eedugon:
```suggestion
Red Hat Enterprise Linux 8 and 9, along with Rocky Linux 8 and 9, run {{ece}} (ECE) on Podman rather than Docker. Use these steps to install and configure Podman, and to apply the SELinux, XFS, and kernel tuning that ECE expects on RHEL-compatible distributions.
```

kunisen:
Hi @eedugon 

> align the OS preparation guides of the 3 OSs we support

Do we still have this on the board or shall we merge this PR first and later when we have the plan in more affirmative state then we take further action?

eedugon:
> Do we still have this on the board or shall we merge this PR first and later when we have the plan in more affirmative state then we take further action?

My plan is to finish with this PR very soon, and then align the other install guides to this one at a later stage.
I hope to get this in a state for a final review between today and tomorrow.

eedugon (approved):
Approving to remove the lock, then I'll set new reviewers

deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
eedugon: done

eedugon:
@kunisen, I've added significant changes to the PR, please let me know your thoughts.

I've updated a lot the structure to ensure the steps are separated in clear and differentiated areas.
Also added a list of verification steps after reboot to ensure all is ready for the ECE installation, incuded a Next steps section, and some other changes.

In the near future I'll check with PMs and devs to see if we can align the 3 OSs documents so they have similar structure.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -415,14 +466,14 @@ Reboot the host and verify that the kernel parameters, Docker configuration, mou
 
     The output should show `/mnt/data` as an `xfs` file system with project quotas enabled.
 
-1. Verify the ownership and permissions of the data directories:
+1. Verify the ownership and permissions of the data directory:
 
     ```sh
-    stat -c "%U:%G %a %n" /mnt/data /mnt/data/docker
+    stat -c "%U:%G %a %n" /mnt/data
```
kunisen, line 471: @eedugon do we know why we want to drop this?

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -79,13 +85,47 @@ The following commands assume that you are logged in as the non-root user accoun
         sudo groupadd docker
         ```
 
+    1. (Optional) If you want to use the recommended `elastic` user and it doesn't exist yet, create it:
+
+        ```sh
+        sudo useradd -m -g elastic -G docker elastic
+        ```
+
+        The user that prepares the host must be able to run commands with `sudo`. Configure sudo access according to your organization's policies. For example, to allow the `elastic` user to run the commands in this guide without entering a password:
```
kunisen: I have some questions on this:
* Should we mention this earlier like under the first `Prepare the user account for ECE` block?
* Do we have to write it in such a detailed way? (I guess no because I saw you removed this by looking at the latest status - https://github.com/elastic/docs-content/pull/6598/changes)

kunisen (commented):
@eedugon thank you!

### [1]

I see some order change made in https://github.com/elastic/docs-content/pull/6598/changes/545cd4d46d60ce7cfa77476e8ad06fd0ee2054f4 and https://github.com/elastic/docs-content/pull/6598/changes/48ffa2fa041bd6149a5808a707f678b58920e4aa.

I think I am not in the position of reviewing the order change is good enough and won't impact on installation or not. 
My personal feeling is I can see they generally look good, but if you want to get 100% sign-off, I would recommend you reach out to ECE dev for a 2nd look.

### [2]

Other than that, the changes look good in general. 
There are some detailed places that I have questions, but no thing big.

---

For the next, if you are confident enough with the reordering per above [1], then I am fine with we merge the PR to have a much clearer OS prep docs :) 
Thanks again mate!

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: 1) We validate that docker root directory is `/mnt/data/docker` in a previous verification step.

2) Because `/mnt/data/docker` should belong to `root:root`, not to the `ECE user:elastic` :) So I didn't want the user to enter both directories and explain that the docker one should belong to root.

3) What we really care before installing ECE is that the user and group elastic are the owners of `/mnt/data` and with the right privileges. That would make `/mnt/data/elastic` to be created properly during the installation, so it's the main check IMO.

Anyway if you prefer we can also mention both directories in the command, but we should update the statement and tell the user that /mnt/data should belong to the ECE user but /mnt/data/docker should be `root:root`.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: I've removed all that part if I'm not mistaken. And I agreed, that level of detail shouldn't be needed. Our doc's main focus should be ECE preparations and requirements, not general knowledge like how to configure sudo.

eedugon:
> I think I am not in the position of reviewing the order change is good enough and won't impact on installation or not.

ordering is fine, the steps were unrelated and I've tried to give better sense to the steps and their order, although in some cases the order of certain actions is completely irrelevant.

Also I've tested everything step by step, and we'll be waiting for control-plane review and ack on this.
thanks!

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,153 @@ products:
...
-The following instructions explain how to prepare your hosts on SUSE Linux Enterprise Server 12 SP5 (SLES 12) or 15 (SLES 15).
+This guide explains how to prepare a SUSE Linux Enterprise Server (SLES) host for an {{ece}} (ECE) installation. It covers the operating system configuration required before you install ECE, including Docker installation, XFS quota configuration, and other host-specific settings.
 
+SLES hosts use `zypper` to install Docker and require manual XFS quota setup because XFS is not the default filesystem on SLES. The steps on this page target SLES 15 SP4.
+
+::::{warning}
+SLES 12 SP5 reached general support end of life on **October 31, 2024**. Use SLES 15 or later for new {{ece}} installations, and migrate existing SLES 12 SP5 hosts.
+::::
+
+* [Prerequisites](#ece-configure-host-suse-prerequisites)
```
shainaraskas: this page nav duplicates the right hand nav. do we expect people will be jumping around this guide much? if we expect people to have to do all of these steps, I would skip putting this full list in

if the steps are order-sensitive (or partially order-sensitive) we should make sure that the steps get numbered or dependencies are called out in the prose (the latter for the 'only sometimes the order matters' case)

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,153 @@ products:
...
+
+    1. Add the user to both groups:
 
         ```sh
-        sudo zypper install -y curl device-mapper lvm2 net-tools docker=24.0.7_ce-150000.198.2 net-tools
+        sudo usermod -aG elastic,docker $USER
         ```
 
-4. Disable nscd, as it interferes with Elastic’s services:
+1. Verify that the user that will run ECE has a UID and GID of at least 1000:
```
shainaraskas: mostly for my own info: why do we need this?

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,153 @@ products:
...
 
-## Set up OS groups and user [ece_set_up_os_groups_and_user]
+    Then set the user's primary group to `elastic`:
 
-1. If they don’t already exist, create the following OS groups:
+    ```sh
+    sudo usermod -g <elastic_group_gid> $USER
+    ```
+
+## Install Docker on SLES [ece-install-docker-sles12]
```
shainaraskas, line 95: need an intro sentence after the heading

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,153 @@ products:
...
+
+::::{include} /deploy-manage/deploy/_snippets/ece-supported-combinations.md
+::::
+
+1. Remove Docker and any previously installed Podman packages.
```
shainaraskas: there is some inconsistency with punctuation use throughout (or as far as I see so far :)). generally I think we need to intro our code snippets with colons

```suggestion
1. Remove Docker and any previously installed Podman packages:
```

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,95 +12,153 @@ products:
...
 ::::
 
-
 ::::{important}
-You must use XFS and have quotas enabled on all allocators, otherwise disk usage won’t display correctly.
+You must use XFS and have quotas enabled on all allocators. Otherwise, disk usage won't display correctly.
 ::::
 
-
 **Example:** Set up XFS on a single, pre-partitioned block device named `/dev/xvdg1`. Replace `/dev/xvdg1` in the following example with the corresponding device on your host.
```
shainaraskas: would try to kill this `example` syntax. use something like "For example," or "In the following example .."

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -109,49 +167,69 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
+1. Create the `/mnt/data/docker` directory for Docker storage:
 
     ```sh
-    sudo systemctl stop docker
+    sudo install -o $USER -g elastic -d -m 700 /mnt/data/docker
+    ```
+
+## Configure system settings [ece-update-config-sles]
+
+1. Stop the `nscd` service and prevent it from starting automatically (it can interfere with Elastic services):
```
shainaraskas:
```suggestion
1. Stop the `nscd` service and prevent it from starting automatically. This service can interfere with Elastic services.
```

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -109,49 +167,69 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
+
+    ```sh
+    sudo systemctl stop nscd
+    sudo systemctl disable nscd
     ```
 
-2. Enable cgroup accounting for memory and swap space.
+1. Enable cgroup accounting for memory and swap space.
```
shainaraskas: when there is a sub-procedure we often also want colons

```suggestion
1. Enable cgroup accounting for memory and swap space:
```

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -109,49 +167,69 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
 
-    1. In the `/etc/default/grub` file, ensure that the `GRUB_CMDLINE_LINUX=` variable includes these values:
+    1. In the `/etc/default/grub` file, ensure the `GRUB_CMDLINE_LINUX=` variable includes these values:
 
         ```sh
         cgroup_enable=memory swapaccount=1 cgroup.memory=nokmem
```
shainaraskas, line 215: are these copypasteable, or is this more like a list of things to check? if the latter, consider wrapping thme each onto their own line

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -109,49 +167,69 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
 
         ```sh
         sudo update-bootloader
         ```
 
-3. Configure kernel parameters
+1. Configure kernel parameters:
 
     ```sh
     cat <<EOF | sudo tee -a /etc/sysctl.conf
```
shainaraskas, line 227: can we swap to a modern docs link here? can we swap these with code annotations?

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -168,25 +246,16 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
 
-4. Adjust the system limits.
-
-    Add the following configuration values to the `/etc/security/limits.conf` file. These values are derived from our experience with the {{ecloud}} hosted offering and should be used for ECE as well. These settings apply to host-level processes and interactive user sessions (for example, SSH).
-
-    ::::{tip}
-    If you are using a user name other than `elastic`, adjust the configuration values accordingly.
-    ::::
-
+1. Adjust the system limits by adding the following configuration values to the `/etc/security/limits.conf` file. These values are based on the {{ecloud}} hosted offering and should be used for ECE as well. If needed, make sure to replace `elastic` with your user name.
```
shainaraskas: maybe the fact that this is based on ech is not so important anymore as the two platforms are less related. but we can skip this "hosted offering" language

```suggestion
1. Adjust the system limits by adding the following configuration values to the `/etc/security/limits.conf` file. These values are based on {{ech}} and should be used for ECE as well. If needed, make sure to replace `elastic` with your user name.
```

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -204,70 +273,52 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
-     "auths": {
-       "docker.elastic.co": {
-         "auth": "<auth-token>"
-       }
-     }
-    }
-    ```
-
-6. If you did not create the mount point earlier (if you did not set up XFS), create the `/mnt/data/` directory as a mount point:
+1. (Optional) Tune additional network kernel parameters for production workloads. Create a `70-cloudenterprise.conf` file in `/etc/sysctl.d/` and include these settings:
```
shainaraskas:
```suggestion
1. Optional: Tune additional network kernel parameters for production workloads. Create a `70-cloudenterprise.conf` file in `/etc/sysctl.d/` and include these settings:
```

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -204,70 +273,52 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
+    * If system-level values exceed 300 seconds, {{es}} automatically lowers them to 300 seconds.
+    * Values below 300 seconds are used as is.
 
+    For non-{{es}} connections such as the proxy layer, consider reducing the following TCP keepalive parameters to detect stale network sessions and prevent firewalls from dropping silent connections:
+    * `net.ipv4.tcp_keepalive_time`
+    * `net.ipv4.tcp_keepalive_intvl`
+    * `net.ipv4.tcp_keepalive_probes`
+    :::
 
 ## Configure the Docker daemon [ece-configure-docker-daemon-sles12]
```
shainaraskas, line 297: haven't been commenting on this, but pref a little tl;dr of what is being done / the role of the thing you're configuring below each heading

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -204,70 +273,52 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
 
-1. Edit `/etc/docker/daemon.json`, and make sure that the following configuration values are present. The `default-ulimits` section ensures that all containers created by the Docker daemon inherit the correct resource limits, which are not inherited from `/etc/security/limits.conf`.<br>
+1. Edit `/etc/docker/daemon.json`, and make sure the following configuration values are present:
```
shainaraskas:
```suggestion
1. Edit `/etc/docker/daemon.json` to make sure the following configuration values are present:
```

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -286,14 +337,32 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
       }
     }
     ```
+    1. The `max-size` and `max-file` options configure rotation for the `json-file` logs created by each container. Adjust these values to match your logging requirements.
+    2. The `default-ulimits` setting increases the maximum number of open file descriptors available to Docker containers.
+
+1. _If the Docker registry doesn't require authentication, skip this step._
```
shainaraskas: can we reword this to "if the docker registry requires authentication, then" or something?

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -307,76 +376,110 @@ You must use XFS and have quotas enabled on all allocators, otherwise disk usage
...
+Reboot the host and verify that the required system, storage, and Docker configuration has been applied successfully.
+
+1. Reboot your system to ensure that all configuration changes take effect:
+
+    ```sh
+    sudo reboot
+    ```
 
-4. Recommended: Tune your network settings.
+    Then log in again as your ECE user.
```
shainaraskas: is this process obvious to the user?

shainaraskas (approved):
I don't know enough about suse to validate this from a technical perspective, but this doc is looking really nice from a content/style perspective. 

I dropped some style comments throughout for you to look at, and some questions that I got stuck on that might not be super relevant to a seasoned user. 

since I'm out for the next bit, don't worry about waiting on me for another loop of reviews - it's basically good to go as is.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
kunisen: @eedugon @shainaraskas 

I didn't know why we removed the `4. Disable nscd, as it interferes with Elastic’s services:` but we do have a page for describing that. 
https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-sysconfig

Maybe we shouldn't delete that line?

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
kunisen: maybe this is related to https://github.com/elastic/docs-content/pull/6598#discussion_r3524137890?

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
kunisen: ++ we may also change the ECH value too and also there's no absolute connection to say "we use this on ECH thus we use the same on ECE". 

> two platforms are less related.

But IMHO they are still related - including the backend and UI. 
It's just the users who use ECE may not be using ECH at the same time because they are on different Elastic subscriptions, so it doesn't make an absolute logic sense or it doesn't make an absolute need to explain the way like this. 

From a user perspective, I would think it would be better to explain in the following perspective:
* It makes system more stable and reliable, and it is designed to handle traffic load for production in a nice way, etc.

But that's a PM call to say it in that way or not, so if we are unsure what to call it out, ++ on @shainaraskas's suggestion to remove it at all.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
shainaraskas: I think it's covered elsewhere:

https://github.com/elastic/docs-content/pull/6598#discussion_r3524158147

eedugon: Exactly. It was moved to a sub-section that fits much better with the nature of the step.

eedugon: @shainaraskas : 
> Verify that the user that will run ECE has a UID and GID of at least 1000:

If you are asking about that, it's an ECE limitation. But I don't remember exactly why, probably an ECE internal check to ensure it doesn't run as root 🤷

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: It's copypasteable! Better to leave it like that as the reader has to inject that entire string with spaces into the variable definition within the file.

We could offer an example of how the variable could look like.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
kunisen: We have https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-users-permissions mentioning this: 
> To install ECE: A user with a UID and GID greater than or equal to 1000 who is part of the docker group. You must not install ECE as the root user.

And IMHO, yes, that's more a best practice to have a non-root user (UID/GID >= 1000) on Linux distributions.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: I'm totally opened for any option here. As we already know, our OS guides are not aligned at all:

- RHEL : Doesn't mention anything about from where these `limits` come from <-- I'll follow this approach because that sentence doesn't add much benefit for a config file like this.

- Ubuntu and SuSE: `These values are derived from our experience with the Elastic Cloud hosted offering and should be used for Elastic Cloud Enterprise as well.`

I'm going to leave this shorter for the moment, aligned with the RHEL doc. Ubuntu doc will still have this longer sentence. And when we do the final alignment of all OS docs with devs and PMs we will decide the best way to present the `limits.conf` configuration.

@kunisen , I wouldn't add any prose to `sell` (stability & reliability) the values of this file. It's very common in the industry to just specify the limits and kernel parameters required by the applications without needing to explain all of them. And in this case `limits.conf` could not be as important as we think, as it doesn't apply to containers, only to pam-based sessions, and in ECE 95% of things run as containers.

So, IMO better to leave this simply as it is or without the ECH statement.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: Yes, how to log in as a certain user or become a certain user in the OS should be obvious to the reader.

Do you prefer to put that directly in the step?

> 1. Reboot your system and log in again as your ECE user to ensure that all configuration changes take effect:

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: Besides the best practice (2-999 = other system or services users, 1000+ = other users) I'd swear we have something hardcoded in ECE that requires this "practice", and that's the reason of the `must`, but maybe I'm wrong.... anyway I wouldn't change it :)

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
shainaraskas: no preference, just wondering :)

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: I've added a little extra prose to the relevant sections, the majority of them are independent.
Page nav removed. I agree it's better like this.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: solved ;)

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: included, together with the statement about data directories preparation and docker install steps should have been already performed.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: updated, hoping it's better now :)

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,181 +12,241 @@ products:
...
+        If you create a dedicated user, we recommend granting it `sudo` privileges and then logging in or switching to that user before continuing. This allows `$USER` to automatically resolve to the user that will install and run ECE.
+
+        Alternatively, you can continue using a different account with `sudo` privileges, and replace `$USER` with the name of the user that will install and run ECE in the remaining commands in this guide.
+        ::::
+
+    1. Add the user to both groups:
 
         ```sh
-        sudo zypper install -y curl device-mapper lvm2 net-tools docker=24.0.7_ce-150000.198.2 net-tools
+        sudo usermod -aG elastic,docker $USER
```
Frannie-Ludmilla: If the user decide to create `elastic` as $USER then the previous 
```bash
 sudo useradd -m -g elastic -G docker elastic
```
already sets the `elastic` user to have elastic,docker as groups.
What I am trying to say is that if the user takes the optional step, then this step could be skipped as these groups are already assigned.
This step is important if they decide to use an already existent user and we need to add (by appending) the needed groups.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
```diff
@@ -12,181 +12,241 @@ products:
...
-3. Add an entry to the `/etc/fstab` file for the new XFS volume. The default filesystem path used by ECE is `/mnt/data`.
+## Prepare the data directories [ece-prepare-data-directories-sles]
+
+Prepare the data directories used by {{ece}} and Docker. These steps create the `/mnt/data` and `/mnt/data/docker` directories, mount an XFS file system if applicable, and apply the required ownership and permissions.
+
+1. Create the `/mnt/data` directory if it doesn't already exist:
 
     ```sh
-    /dev/xvdg1	/mnt/data	xfs	defaults,pquota,prjquota,x-systemd.automount  0 0
+    sudo install -o $USER -g elastic -d -m 700 /mnt/data
```
Frannie-Ludmilla, line 173: TBH, since I didn't know for sure whether it actually did anything (apart from returning a 0 exit code), we may suggest to check that the newly created folder has elastic (owner) and elastic (group) as well as  drwx------ (700)
```
francescapesci@francesca-ece-suse:~> ls -ld /mnt/data
drwx------ 2 elastic elastic 6 Jul 14 16:13 /mnt/data
```

Frannie-Ludmilla: I see that https://github.com/elastic/docs-content/pull/6598/changes#diff-b13773bd6f750767482d21362e706bb0570cf173593672a4afcdfe7f1b3a8e0cR471 afterwards we test it as well, so this becomes a small nit 👍🏽

Frannie-Ludmilla (approved):
LGTM, maybe just a nit about a (possibly) duplicate action if the user did not exist.
Thank you for the env to test this 💻 !

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: yeah, the verification is at the end of the doc. But it's a good point, I'll check this a bit further.

deploy-manage/deploy/cloud-enterprise/configure-host-suse.md
eedugon: Yeah, that step is only needed if the user didn't create a user in the previous step, good point. I'll fix it.

kunisen:
Thank you all!

PR #6619, opened 2026-05-21
Add ES|QL note to runtime fields docs

leemthompo:
## Summary
- Adds a "Runtime fields in ES|QL" section to the runtime fields page
- Clarifies that ES|QL respects mapping-defined runtime fields but does not support defining new runtime fields at search time, pointing to `EVAL` as an alternative
- Distinguishes runtime fields from unmapped fields, linking to `SET unmapped_fields`

Companion to an upcoming elasticsearch repo PR that updates the ES|QL docs (EVAL, limitations, and the generated `unmapped_fields` setting description).

### DIFF
--- a/manage-data/data-store/mapping/runtime-fields.md
+++ b/manage-data/data-store/mapping/runtime-fields.md
@@ -61,7 +61,7 @@ Queries against runtime fields are considered expensive. If [`search.allow_expen
 
 ## Runtime fields in {{esql}} [runtime-esql]
 
-{{esql}} treats runtime fields defined in the index mapping like regular mapped fields. However, {{esql}} does not support defining new runtime fields at search time. To create computed columns at query time, use the [`EVAL`](elasticsearch://reference/query-languages/esql/commands/eval.md) command instead.
+In {{esql}}, runtime fields defined in the index mapping function as regular mapped fields. However, {{esql}} does not support defining new runtime fields at search time. To create computed columns at query time, use the [`EVAL`](elasticsearch://reference/query-languages/esql/commands/eval.md) command instead.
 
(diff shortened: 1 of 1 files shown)


manage-data/data-store/mapping/runtime-fields.md
```diff
@@ -59,6 +59,13 @@ Queries against runtime fields are considered expensive. If [`search.allow_expen
 ::::
 
 
+## Runtime fields in {{esql}} [runtime-esql]
+
+{{esql}} treats runtime fields defined in the index mapping like regular mapped fields. However, {{esql}} does not support defining new runtime fields at search time. To create computed columns at query time, use the [`EVAL`](elasticsearch://reference/query-languages/esql/commands/eval.md) command instead.
```
yetanothertw: This is not my area of expertise, so feel free to ignore, but I just wonder if we could make this a little more user-focused (as opposed to product-focused) to soften down the anthropomorphism? 

```suggestion
In {{esql}}, runtime fields defined in the index mapping function as regular mapped fields. However, {{esql}} does not support defining new runtime fields at search time. To create computed columns at query time, use the [`EVAL`](elasticsearch://reference/query-languages/esql/commands/eval.md) command instead.
```

I know it introduces passive voice, but in my books that's a lesser evil. 🙈

yetanothertw (approved):
Looks good, thanks Liam! 🪴 

Left a small suggestion, but definitely not a blocker.

manage-data/data-store/mapping/runtime-fields.md
leemthompo: @yetanothertw I'd counter with this alternative:

`In {{esql}}, runtime fields defined in the index mapping function as regular mapped fields.`

WDYT? 😄

yetanothertw: Love it, thanks!

PR #6626, opened 2026-05-21
Add Console tab to Dashboards API example (learning tutorial)

florent-leborgne:
## Summary

- Adds a **Console-first tab set** (Console + curl) to the "Recreate this dashboard with one API call" dropdown on the [Kibana data exploration learning tutorial](https://www.elastic.co/docs/explore-analyze/kibana-data-exploration-learning-tutorial) page, mirroring the pattern introduced for Lens chart pages in #5645.
- Updates `.github/scripts/verify-dashboards-api-example.py` to extract the JSON payload from the `console` block instead of the `curl` block, consistent with the validation approach used in the Lens API examples script.

## Validation

The updated script was run against a live Kibana serverless instance and confirmed that the Console payload creates a valid 11-panel dashboard.

## Test plan

- [ ] Check that the dropdown renders with Console tab first and curl tab second
- [ ] Check that all 11 callouts appear in both tabs with their numbered descriptions
- [ ] Run `verify-dashboards-api-example.py` against a Kibana instance to confirm the Console payload is valid

### DIFF
--- a/.github/scripts/verify-dashboards-api-example.py
+++ b/.github/scripts/verify-dashboards-api-example.py
@@ -79,20 +79,20 @@ EXPECTED_PANEL_COUNT = 11
 
 
 def extract_payload(markdown_path: Path) -> dict:
-    """Return the parsed JSON payload from the curl example in the markdown."""
+    """Return the parsed JSON payload from the Console tab example in the markdown."""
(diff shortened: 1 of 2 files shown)


reakaleek (approved):

leemthompo (approved):
💻

PR #6647, opened 2026-05-22
Add May 20 Kibana Serverless release notes

lcawl:
## Summary

This PR adds changelog snippets to the Elastic Cloud Serverless release note pages that are derived from https://github.com/elastic/kibana/pull/270524

This is a stop-gap until we can publish directly from the bundles.

## Generative AI disclosure

- [ ] Yes  
- [x] No


eedugon (approved):
LGTM

benironside (approved):
LGTM. Only note is that "ES|QL" is used instead of the {{esql}} attribute. Not sure if other attributes may be missing as well. I've noticed the LLM often messes that up for release notes specifically. I wonder if there's a way to improve its performance on attributes specifically 🤔

lcawl:
> LGTM. Only note is that "ES|QL" is used instead of the {{esql}} attribute. Not sure if other attributes may be missing as well. I've noticed the LLM often messes that up for release notes specifically. I wonder if there's a way to improve its performance on attributes specifically 🤔

Yeah, I removed that battle from my latest skills since it seemed like a nice-to-have not something necessary... especially if we have a future where these changelogs are used for other purposes too (e.g. blurbs in UI notification centres)

PR #6727, opened 2026-05-27
Adds explanation on when to use dense or sparse embeddings

kosabogi:
## Summary

This PR adds explanation on when to use dense and sparse vectors to the [Tutorial: Dense and sparse workflows using ingest pipelines](https://www.elastic.co/docs/solutions/search/vector/dense-versus-sparse-ingest-pipelines) page.

Related issue: https://github.com/elastic/docs-content-internal/issues/854

## Generative AI disclosure

- [x] Yes  
- [ ] No  
 Claude in Cursor


solutions/search/vector/dense-versus-sparse-ingest-pipelines.md
```diff
@@ -25,9 +25,40 @@ This guide shows how to implement semantic search in {{es}} with deployed NLP mo
...
+
+Use dense embeddings when you want to match content by overall meaning in vector space, not by overlapping keywords. Dense models map text (and other content types) to fixed-length vectors so that conceptually similar items sit close together. Dense vectors are compact arrays of floating-point values from an embedding model.
+
+Common use cases include:
+
+- [Retrieval augmented generation (RAG)](../rag.md): Retrieve document passages that answer a user's question, even when the question and the source text use different words.
+- [Natural language Q&A](/explore-analyze/machine-learning/nlp/ml-nlp-text-emb-vector-search-example.md): Match questions like "How do I reset my password?" to FAQ entries, product documentation, or policy pages.
+- [Recommendations and similarity](knn.md): Find related articles, products, or media. For example, you can surface articles like the current one or visually similar product images.
+
+Dense embeddings are a good choice when you need multilingual retrieval or a specific third-party embedding model you have already evaluated on your data.
```
seanhandley: I'd remove this sentence. Not all dense embeddings models are multilingual. We support Jina Embeddings v3, which is English only, for example.

Also, the need to use a model that's already been evaluated for a usecase could apply to a sparse embedding model too - it's a question of previous technical decisions and the need to accommodate them.

Could maybe say something like

> Dense embeddings are ideal when you care more about the semantic meaning of search terms than exact keyword matches - they excel at retrieving relevant results based on synonyms and paraphrasing of the original query to return results that reflect the user's intensions.

solutions/search/vector/dense-versus-sparse-ingest-pipelines.md
```diff
@@ -25,9 +25,40 @@ This guide shows how to implement semantic search in {{es}} with deployed NLP mo
...
+
+For more detail, refer to [Sparse vector search in {{es}}](sparse-vector.md).
+
+#### Dense embeddings [deployed-dense-embeddings]
+
+Use dense embeddings when you want to match content by overall meaning in vector space, not by overlapping keywords. Dense models map text (and other content types) to fixed-length vectors so that conceptually similar items sit close together. Dense vectors are compact arrays of floating-point values from an embedding model.
+
+Common use cases include:
+
+- [Retrieval augmented generation (RAG)](../rag.md): Retrieve document passages that answer a user's question, even when the question and the source text use different words.
```
seanhandley, line 54: Could be an opportunity here to mention Context Engineering more prominently than RAG?

RAG is still a useful concept but I think we're positioning ourselves in the market as a broader solution for context engineering as a whole.

seanhandley (commented):
Looks almost ready @kosabogi - just a couple of thoughts.

solutions/search/vector/dense-versus-sparse-ingest-pipelines.md
leemthompo: ++ there's some good definitions in this Anthropic blog post: https://www.anthropic.com/engineering/effective-context-engineering-for-ai-agents

this could be a good spot to link to agent builder for OOTB toolkit

Per the [glossary definition](https://www.elastic.co/docs/explore-analyze/ai-features/agent-builder/glossary#b): 

 "Agent Builder combines LLM reasoning with skills, tools, and best practices for context engineering and retrieval, so responses are accurately and efficiently grounded in your data."

kosabogi:
Thanks for the feedback, @seanhandley and @leemthompo! I’ve added context engineering as a use case for dense vectors. Let me know how it looks now.

seanhandley (approved):

leemthompo (approved):
Nice! I wonder if this is the right spot to mention that you can also combine these approaches. A production search setup might combine lexical, dense, and sparse in a tiered approach. But this page is set up as a "versus" so perhaps food for a follow-up...

kosabogi:
> Nice! I wonder if this is the right spot to mention that you can also combine these approaches. A production search setup might combine lexical, dense, and sparse in a tiered approach. But this page is set up as a "versus" so perhaps food for a follow-up...

Good idea! I opened an issue to address this later: https://github.com/elastic/docs-content/issues/6816

PR #6734, opened 2026-05-27
eck webhook default value clarified

### LINKED ISSUE #6160: [Website]: enable-webhook default value seems wrong.
### Before you submit

### Type of issue

None

https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/configure-validating-webhook

### What happened?

Hi, Team
In `operator.yaml`, the default value of `enable-webhook` is `true`.
but the Doc say that the default value is `false`. 
The explanation isn't clear on Doc, when is the default value of `enable-webhook false`?
or just Doc is wrong?

Image

### Additional info

_No response_

eedugon:
## Summary

Closes https://github.com/elastic/docs-content/issues/6160

Clarifies a bit better the default value of `false` of enable-webhook at code level in ECK, while at the same time our default installation enables it by default.


deploy-manage/deploy/cloud-on-k8s/configure-validating-webhook.md
```diff
@@ -45,13 +45,16 @@ You can customize almost all aspects of the webhook setup by changing the [opera
...
-| `enable-webhook` | false | This must be set to `true` to enable the webhook server. |
+| `enable-webhook` | false | Set to `true` to enable the webhook server. |
 | `manage-webhook-certs` | true | Set to `false` to disable auto-generating the certificate for the webhook. If disabled, you must provide your own certificates using one of the methods described later in this document. |
 | `webhook-cert-dir` | /tmp/k8s-webhook-server/serving-certs | Path to mount the certificate. |
 | `webhook-name` | elastic-webhook.k8s.elastic.co | Name of the `ValidatingWebhookConfiguration` resource. |
 | `webhook-secret` | elastic-webhook-server-cert | Name of the secret containing the certificate for the webhook server. |
 | `webhook-port` | 9443 | Port to listen for incoming validation requests. |
 
+:::::{note}
+Although the `enable-webhook` flag default is `false`, the default ECK installation manifests and Helm chart set it to `true`.
```
shainaraskas: can we remove "default" later in this sentence? we might rephrase to say "if you're editing x y z / installed using x y z methods, the setting is already present and set to true" 

I think a think that makes this confusing is we call this manual configuration but likely it's being applied in a layer over one of these defaults. we also say "If you installed ECK without the webhook" but what is the context that makes this possible? 

consider putting this default mismatch info directly in the description rather than as a note below the table

eedugon: > "If you installed ECK without the webhook" but what is the context that makes this possible?

You can install ECK without the webhook by tweaking the `operator.yaml` manifest file before installing through a manifest, or by tweaking the `values.yaml` before installing the helm chart. When doing that, you can decide to either "remove the webhook setting" (it would then default to false :) ), or set it to false explicitly.

I think it's a good practice when installing a production system to decide what exact configuration you want and not just deploy the quickstart or the default provided file, and some users could have installed ECK without the webhook.

Anyway, this PR aims to solve a possible confusion due to the fact that we provide default config files enabling the webhook but internally the code has a default value of disabling it. That can generates a bit of confusion when reading both sentences.

We can rephrase it to something like:

> Although the `enable-webhook` flag of the operator defaults to `false` when not provided, the standard ECK installation manifests and Helm chart set it to `true` explicitly.

But please suggest whatever you think it's better :)

What I want to ensure is that people doesn't get confused when reading these two sentences:
- The webhook is installed by default (through our manifests)
- `enable-webhook` default value is false

Oh, and if you believe we can remove the note completely because the other minor change we did already avoids the possible confusion then I'm totally ok with it too ;)

shainaraskas: that rephrase looks better to me.

shainaraskas (approved):
🏁  with the rephrase

deploy-manage/deploy/cloud-on-k8s/configure-validating-webhook.md
eedugon: Instead of the external "note", I've added the sentence directly in the table, in the place where the confusion could occur. I think it's a better approach.

eedugon:
> 🏁 with the rephrase

Sorry I requested a new review :) rephrase included but directly within the table. I think it's nicer and less disruptive.

deploy-manage/deploy/cloud-on-k8s/configure-validating-webhook.md
```diff
@@ -45,14 +45,13 @@ You can customize almost all aspects of the webhook setup by changing the [opera
 
 | Configuration option | Default value | Description |
 | --- | --- | --- |
-| `enable-webhook` | false | This must be set to `true` to enable the webhook server. |
+| `enable-webhook` | false | Set to `true` to enable the webhook server. Although the operator default is `false`, the standard ECK installation manifests and Helm chart explicitly set it to `true`. |
```
shainaraskas: nit: this is secondary info so a line break is nice

```suggestion
| `enable-webhook` | false | Set to `true` to enable the webhook server.<br><br>Although the operator default is `false`, the standard ECK installation manifests and Helm chart explicitly set it to `true`. |
```

shainaraskas (approved):
approved with a small suggest

PR #6767, opened 2026-05-29
Kibana --max-old-space-size setting clarification on ECK

kunisen:
## Summary

* To address https://github.com/elastic/docs-content-internal/issues/1251

As mentioned in https://github.com/elastic/docs-content-internal/issues/1251#issuecomment-4569664586, 

### [1] kibana memory - default memory calculation logic

By default, it sets `--max-old-space-size` automatically, already.

https://www.elastic.co/docs/release-notes/kibana
> Containers now set the default Node.js heap to 75% of available memory up to a maximum of 4096 Mb. Previously, this was set to 50% [#246073](https://github.com/elastic/kibana/pull/246073).

Before Kibana 9.4 (< 9.4), it's set to 50% of the container memory.
Starting in Kibana 9.4 (>= 9.4.0), it's set to 75% of the container memory (with max cap to 4096MB).

*The "Mb" (small b) is a mistake per [this](https://www.elastic.co/docs/deploy-manage/production-guidance/kibana-configure-memory): `The option accepts a limit in MB`.*

### [2] kibana memory - on ECH and ECE

In ECE and ECH, it is further overridden based on [our internal code](https://github.com/elastic/cloud/blob/4.1.0/scala-services/runner/src/main/resources/templates/allocator/kibana/kibana.sh#L60-L64), which is 800MB per 1GB. (MiB and GiB). 

This is the reason why we have `In orchestrated environments like ..., you should not override Kibana's default memory limit using --max-old-space-size.`

### [3] kibana memory - on ECK

However, in ECK, there's no such auto-adjust / override code piece. 
Instead, it uses the default logic.

So on ECK, the behavior is same as [1]:
* Before Kibana 9.4 (< 9.4), it's set to 50% of the container memory.
* Starting in Kibana 9.4 (>= 9.4.0), it's set to 75% of the container memory (with max cap to 4096MB).

### [4] One more different topic - ECK customize pods doc

https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/customize-pods gives the example of overriding kibana `--max-old-space-size=2048` to 2GiB which is based on https://github.com/elastic/cloud-on-k8s/blob/v3.4.0/deploy/eck-stack/examples/custom-elasticsearch-kibana.yaml#L70-L78, and has a pair of container memory setting be up-lifted to 2.5GiB. 

But this 2.5GiB is not correctly propagated in the doc. If we tweak the kibana memory setting without giving enough container memory (limits), then it may experience OOM. So we need to also put the container memory (limits) description in doc (in addition to the code).

---

So we need to tweak https://www.elastic.co/docs/deploy-manage/production-guidance/kibana-configure-memory and make it clear of below:
* to address [1], Kibana default memory logic difference before and after 9.4
* to address [2], on ECE and ECH, this setting is auto-adjusted
* to address [3], on ECK, this setting is not auto-adjusted

Also re https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/customize-pods,
* to address [4], we need to include the container memory limits pieces.

---

Thanks!

## Generative AI disclosure

- [ ] Yes  
- [x] No

### DIFF
--- a/deploy-manage/deploy/cloud-on-k8s/customize-pods.md
+++ b/deploy-manage/deploy/cloud-on-k8s/customize-pods.md
@@ -58,12 +58,22 @@ spec:
           env:
             - name: NODE_OPTIONS
               value:  "--max-old-space-size=2048"
+          resources:
+            requests:
+              memory: 1Gi
+              cpu: 0.5
+            limits:
+              memory: 2.5Gi
+              cpu: 2
 ```
+
+::::{note} 
+When manually setting `NODE_OPTIONS`, make sure you adjust the container's `resources.limits.memory` appropriately.
+::::
 
 ::::{note} 
 Configuration for other {{stack}} applications, like APM Server, or Beats, is identical to the {{kib}} configuration except for the `apiVersion` and `kind` fields.
 ::::
-
 
 The following example shows how it’s also possible to customize the init containers created as part of the Pods to initialize the filesystem or to manage the keystores.
 
--- a/deploy-manage/deploy/cloud-on-k8s/customize-pods.md
+++ b/deploy-manage/deploy/cloud-on-k8s/customize-pods.md
@@ -57,7 +57,7 @@ spec:
         - name: kibana
           env:
             - name: NODE_OPTIONS
-              value:  "--max-old-space-size=2048"
+              value:  "--max-old-space-size=2048"  <1>
           resources:
             requests:
               memory: 1Gi
@@ -67,9 +67,7 @@ spec:
               cpu: 2
 ```
 
-::::{note} 
-When manually setting `NODE_OPTIONS`, make sure you adjust the container's `resources.limits.memory` appropriately.
-::::
+1. When manually setting `--max-old-space-size`, make sure you adjust the container's `resources.limits.memory` appropriately.
 
 ::::{note} 
 Configuration for other {{stack}} applications, like APM Server, or Beats, is identical to the {{kib}} configuration except for the `apiVersion` and `kind` fields.
--- a/deploy-manage/deploy/cloud-on-k8s/customize-pods.md
+++ b/deploy-manage/deploy/cloud-on-k8s/customize-pods.md
@@ -60,7 +60,7 @@ spec:
               value:  "--max-old-space-size=2048"  <1>
           resources:
             requests:
-              memory: 1Gi
+              memory: 2Gi
               cpu: 0.5
             limits:
               memory: 2.5Gi
--- a/deploy-manage/production-guidance/kibana-configure-memory.md
+++ b/deploy-manage/production-guidance/kibana-configure-memory.md
@@ -30,5 +30,5 @@ Starting in {{kib}} version 9.4.0, containers set the default Node.js heap to 75
 
 In orchestrated environments like {{ech}}, {{ece}}, you should not override Kibana’s default memory limit using `--max-old-space-size`. Instead, set the desired {{kib}} memory size at the deployment level. This automatically adjusts the container’s memory allocation and ensures more consistent and predictable performance.
 
-In {{eck}}, the {{kib}} memory size is not automatically adjusted, but the default memory size logic mentioned above applies.
+In {{eck}}, the {{kib}} memory size is not automatically adjusted, but the default memory size logic applies.
 ::::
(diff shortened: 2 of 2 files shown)


deploy-manage/production-guidance/kibana-configure-memory.md
```diff
@@ -26,5 +26,9 @@ The option accepts a limit in MB:
 ```
 
 ::::{note}
-In orchestrated environments like {{ech}}, {{ece}}, or {{eck}}, you should not override Kibana’s default memory limit using `--max-old-space-size`. Instead, set the desired {{kib}} memory size at the deployment level. This automatically adjusts the container’s memory allocation and ensures more consistent and predictable performance.
+Starting in {{kib}} version 9.4.0, containers set the default Node.js heap to 75% of available memory, up to a maximum of 4096 MB. Previously, this was set to 50% regardless of the environment.
```
kunisen, line 29: btw this is based on https://www.elastic.co/docs/release-notes/kibana#:~:text=Containers%20now%20set%20the%20default%20Node.js%20heap%20to%2075%25%20of%20available%20memory%20up%20to%20a%20maximum%20of%204096%20Mb.%20Previously%2C%20this%20was%20set%20to%2050%25

deploy-manage/deploy/cloud-on-k8s/customize-pods.md
```diff
@@ -58,12 +58,22 @@ spec:
...
+              memory: 1Gi
+              cpu: 0.5
+            limits:
+              memory: 2.5Gi
+              cpu: 2
 ```
 
 ::::{note} 
-Configuration for other {{stack}} applications, like APM Server, or Beats, is identical to the {{kib}} configuration except for the `apiVersion` and `kind` fields.
+When manually setting `NODE_OPTIONS`, make sure you adjust the container's `resources.limits.memory` appropriately.
```
eedugon: I'd suggest to use a footnote here, instead of a note. I'd add a `<1>` on line 60, and create the footnote with a similar text, such as:

```
1. When manually setting `--max-old-space-size`, make sure you adjust the container's `resources.limits.memory` appropriately.

(I prefer `max-old-space-size` than NODE_OPTIONS because the latter could potentially be used to provide other options)

File: deploy-manage/production-guidance/kibana-configure-memory.md [hunk 3]
```diff
@@ -26,5 +26,9 @@ The option accepts a limit in MB:
...
+
+In orchestrated environments like {{ech}}, {{ece}}, you should not override Kibana’s default memory limit using `--max-old-space-size`. Instead, set the desired {{kib}} memory size at the deployment level. This automatically adjusts the container’s memory allocation and ensures more consistent and predictable performance.
+
+In {{eck}}, the {{kib}} memory size is not automatically adjusted, but the default memory size logic applies.
```
eedugon:
```suggestion
In {{eck}}, the {{kib}} memory size is not automatically adjusted, but the default memory size logic applies. You can override the default memory settings if they do not meet your requirements.
```

And maybe even including a link to ECK compute resources for an example. WDYT?

Note that `customize-pods.md` shouldn't be the main doc for this topic at ECK level, it should be "ECK Compute resources" or "ECK - Advanced kibana configuration", as a better place.

The customize-pods include an example, and in that example we add the reminder about resources, which is good, but if we link from `production-guidance` to ECK specific docs it should be towards a doc more focused on the topic (compute resources would be my preference)

eedugon (commented):
Two small comments, but I think this is very good already!

deploy-manage/production-guidance/kibana-configure-memory.md
kunisen: Thanks, it sounds great!

kunisen:
WDYT it now? @eedugon

eedugon (approved):
Looks great, thanks!

deploy-manage/deploy/cloud-on-k8s/customize-pods.md
```diff
@@ -57,14 +57,22 @@ spec:
         - name: kibana
           env:
             - name: NODE_OPTIONS
-              value:  "--max-old-space-size=2048"
+              value:  "--max-old-space-size=2048"  <1>
+          resources:
+            requests:
+              memory: 1Gi
```
pkoutsovasilis: let's go with 2Gi here, since this is aligned with the default
```suggestion
              memory: 2Gi
```

pkoutsovasilis (approved):
left one nit, but LGTM in general

deploy-manage/deploy/cloud-on-k8s/customize-pods.md
eedugon: true, it's the new default on ECK 3.4 (I'm creating a different PR for that)

kunisen: Thank you!

PR #6775, opened 2026-05-29
Update elastic-cloud-hosted-planning.md

nclough:
I'd like to add awareness for users who may need to restore from a snapshot that not all settings are restored

## Generative AI disclosure

- [x] Yes  
- [ ] No

### DIFF
--- a/deploy-manage/deploy/elastic-cloud/elastic-cloud-hosted-planning.md
+++ b/deploy-manage/deploy/elastic-cloud/elastic-cloud-hosted-planning.md
@@ -42,6 +42,9 @@ The data in your {{es}} clusters is also backed up every 30 minutes, 4 hours, or
 Clusters that use only one availability zone are not highly
 available and are at risk of data loss. To safeguard against data loss,
 you must use at least two availability zones.
+
+Note: Snapshots do not restore all configurations (e.g., custom repositories).
+Review the full [snapshot contents](https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore#snapshot-contents) list to understand recovery limitations.
 ::::
 
 


deploy-manage/deploy/elastic-cloud/elastic-cloud-hosted-planning.md
```diff
@@ -42,6 +42,9 @@ The data in your {{es}} clusters is also backed up every 30 minutes, 4 hours, or
 Clusters that use only one availability zone are not highly
 available and are at risk of data loss. To safeguard against data loss,
 you must use at least two availability zones.
+
+Note: Snapshots do not restore all configurations (e.g., custom repositories).
+Review the full [snapshot contents](https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore#snapshot-contents) list to understand recovery limitations.
```
shainaraskas, line 47: this note doesn't make sense to be embedded in this warning. I think it would make more sense integrated into line 39 where we have the other snapshot-related limitations.

We also need to use a relative link instead of an absolute one: `/deploy-manage/tools/snapshot-and-restore.md#snapshot-contents`


> The data in your {{es}} clusters is also backed up every 30 minutes, 4 hours, or 24 hours, depending on which snapshot interval you choose. These regular intervals provide an extra level of redundancy. We do support [snapshot and restore](/deploy-manage/tools/snapshot-and-restore.md), regardless of whether you use one, two, or three availability zones. However, with only a single availability zone and in the event of an outage, it might take a while for your cluster come back online. Using a single availability zone also leaves your cluster exposed to the risk of data loss, if the backups you need are not usable (failed or partial snapshots missing the indices to restore) or no longer available by the time that you realize that you might need the data (snapshots have a retention policy). In addition, snapshots [do not capture all configurations](/deploy-manage/tools/snapshot-and-restore.md#snapshot-contents), which can lead to an incomplete recovery.

shainaraskas (changes requested):
Thanks for your contribution! 

Right now, this doesn't render exactly as it should - we can't embed a note inside of a warning. I've made an alternative placement recommendation. The content is pretty dense right now so this recommendation isn't ideal either, but the entire section should be overhauled for readability (our task, not yours). the suggestion is probably good enough for now.

image

PR #6787, opened 2026-06-01
trial deployment size updated

eedugon:
## Summary
Closes https://github.com/elastic/docs-content-internal/issues/1207

Trial deployments are limited to 4GB per AZ (2 zones by default).

@lcawl : would you prefer to solve this in a different way?

## Generative AI disclosure

- [ ] Yes  
- [x] No


shainaraskas (commented):
this approach works but this file also needs to be fixed

https://www.elastic.co/docs/deploy-manage/deploy/elastic-cloud/create-an-organization

eedugon:
> deploy-manage/deploy/elastic-cloud/create-an-organization

That document says: `They include 8 GB of RAM spread out over two availability zones`.

That's a different narrative which in this case doesn't feel incorrect.

Maybe that's even better for the snippet file? :)

shainaraskas:
@eedugon I'm mostly raising it because that's the original problem that was reported in the issue - saying 8 when the number is 4 (even if the sentiment is accurate) was tripping someone up

eedugon:
> @eedugon I'm mostly raising it because that's the original problem that was reported in the issue - saying 8 when the number is 4 (even if the sentiment is accurate) was tripping someone up

Ok! I'll update it and align both sentences :) I mentioned it because the sentence that caused confusion didn't include the `spread out over two zones`, but I agree, let's make this easy!

shainaraskas (approved):

PR #6791, opened 2026-06-02
Update hardware prerequisites for ECE deployment

eedugon:
Trying to improve the format of the footnotes within a table that looked pretty bad.


deploy-manage/deploy/cloud-enterprise/ece-hardware-prereq.md
```diff
@@ -32,16 +32,16 @@ ECE installations with **spinning disks** are not supported when you run allocat
...
 
-2 For high availability, requires three hosts each of the capacities indicated, spread across three availability zones.
+2. For high availability, requires three hosts each of the capacities indicated, spread across three availability zones.
 
-3 For high availability, requires three hosts each of the capacities indicated (except for allocators), spread across three availability zones. For allocators, requires three or more hosts of the capacity indicated, spread across three availability zones.
+3. For high availability, requires three hosts each of the capacities indicated (except for allocators), spread across three availability zones. For allocators, requires three or more hosts of the capacity indicated, spread across three availability zones.
```
shainaraskas: is the "or more" really doing that much heavy lifting here? are the others not actually "three or more" as well?

wonder if this could just use the other footnote because this adds basically no value

deploy-manage/deploy/cloud-enterprise/ece-hardware-prereq.md
```diff
@@ -32,16 +32,16 @@ ECE installations with **spinning disks** are not supported when you run allocat
 | **Memory** | Coordinators | Directors | Proxies | Allocators |
 | --- | --- | --- | --- | --- |
 | Minimum to install | 8 GB RAM | 8 GB RAM | 8 GB RAM | 8 GB RAM<br> |
-| Minimum recommended | 16 GB RAM | 8 GB RAM | 8 GB RAM | 128 GB to 256 GB RAM1<br> |
-| **Small deployment**2 | 32 GB RAM | 32 GB RAM | 16 GB RAM | 128 GB RAM<br> |
-| **Medium deployment**2 | 32 GB RAM | 32 GB RAM | 16 GB RAM | 256 GB RAM<br> |
-| **Large deployment**3 | 128 GB RAM | 128 GB RAM | 16 GB RAM | 256 GB RAM<br> |
+| Minimum recommended | 16 GB RAM | 8 GB RAM | 8 GB RAM | 128 GB to 256 GB RAM (1)<br> |
+| **Small deployment** (2) | 32 GB RAM | 32 GB RAM | 16 GB RAM | 128 GB RAM<br> |
```
shainaraskas: as I think vlada suggested, we can use footnote markup if we want to keep these as footnotes

```suggestion
| **Small deployment**^1^ | 32 GB RAM | 32 GB RAM | 16 GB RAM | 128 GB RAM<br> |
```

deploy-manage/deploy/cloud-enterprise/ece-hardware-prereq.md
```diff
@@ -32,16 +32,16 @@ ECE installations with **spinning disks** are not supported when you run allocat
...
+1. Allocators must be sized to support your {{es}} clusters and {{kib}} instances. We recommend host machines that provide between 128 GB and 256 GB of memory. While smaller hosts might not pack larger {{es}} clusters and {{kib}} instances as efficiently, larger hosts might provide fewer CPU resources per GB of RAM on average. For example, running 64 * 2GB nodes on a 128GB host with 16 vCPUs means that each node will get 2/128 of the total CPU time. This is 1/4 core on average, and might not be sufficient. We recommend inspecting both what is the expected number and size of the nodes you plan to run on your hosts in order to understand which hardware will work best in your environment.
```
shainaraskas: this is a chunky footnote. I think this could become a "sizing allocators" paragraph or subsection

but this has been like this for years so ok to keep as footnote for ease

shainaraskas (approved):
approved to unblock w/ some formatting and succinctness recommendations

deploy-manage/deploy/cloud-enterprise/ece-hardware-prereq.md
eedugon: +10000 on that 🤷  the amount of love and improvements we could apply to this doc set is infinite

deploy-manage/deploy/cloud-enterprise/ece-hardware-prereq.md
eedugon: I think we should rephrase all footnotes in their own sub-sections (`### Sizing allocators` and `### High availability considerations` --> this one with links to the ECE HA doc and the doc where the small/med/large deployments are described).

The footnotes 2 and 3 are absolutely no-sense, first because the concept of `small / medium` (footnote 2) vs `large` (footnote 3) would never affect the HA requirements for the solution, which remains the same, neither for allocators, as allocators will always depend on the amount of deployments to support and the customer specific HA needs (for example allowing a hot-spare allocator to take over workloads in case of an allocator failure).

I think the idea to try to translate is that for coordinators directors and proxies 3 nodes is enough and ok for HA, but allocators could require more depending on the user needs.

Not sure how to write all that, but that's part of the idea besides the text.

I've also noticed that for `medium` and `large` standard layouts we suggest 3 possibilities for allocators:
- 3 big allocators (1 per zone)
- 6 smaller allocators (2 per zone)
- or even 12 smaller allocators (4 per zone).

Maybe that's the reason of the `3 or more` :-D 

image

^^ Also the way we present the `3 hosts for allocators` is also a bit fun :)

If you want we can close this PR and I can prepare a new one giving a bit of extra love and cross links between the [deployment layout descriptions](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/identify-deployment-scenario) and this [HW prereqs](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-hardware-prereq) doc.

shainaraskas: we could ship the footnote formatting fix if you wanted then follow up. this has been broken for a while so it's not urgent.

PR #6794, opened 2026-06-02
Change authentication method from Basic to Dev Tools

philippkahr:
Updated authentication method options in documentation.

Basic is kinda misleading when I look it up, I would love to have Dev Tools to know "ah that is where I need to copy from". Only thing I am wondering if it makes sense to make the Dev Tools tab the primary one

### DIFF
--- a/deploy-manage/monitor/autoops/cc-connect-self-managed-to-autoops.md
+++ b/deploy-manage/monitor/autoops/cc-connect-self-managed-to-autoops.md
@@ -106,7 +106,7 @@ Depending on your selected installation method, you might have to provide some o
 * **Preferred authentication method**: Choose one of the following:
   
   :::::{tab-set}
-  :group: api-key-or-basic
+  :group: api-key-or-devtools
 
   ::::{tab-item} API key
   :sync: api-key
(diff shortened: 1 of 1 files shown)


deploy-manage/monitor/autoops/cc-connect-self-managed-to-autoops.md
```diff
@@ -153,8 +153,8 @@ Depending on your selected installation method, you might have to provide some o
 
   ::::
 
-  ::::{tab-item} Basic
-  :sync: basic
+  ::::{tab-item} Dev Tools
```
shainaraskas, line 156: this is basic authentication because it relies on [basic auth](https://en.wikipedia.org/wiki/Basic_access_authentication) (u/p) being passed to auth a user with the requisite role. this is as opposed to API key auth, which requires passing a key.

the steps in this procedure (dev tools) are just creating the role (which could also be created through the UI)

shainaraskas (commented):
I don't believe this change should be made, because this is comparing the ways of preparing two auth methods that can be used by the autoops connection - API and basic. dev tools is just a means to creating the role used by the user for basic auth.

closing for now - let me know if you believe further changes should be made given this info.

PR #6799, opened 2026-06-02
Add new storage regions for AutoOps Cloud Connect

wajihaparvez:
## Summary
This PR adds GCP regions to the AutoOps Cloud Connect storage regions and moves the info to its own separate page.
~I also decided to keep the table on one page and link to it from the setup instructions because the setup page is getting very long.~

Closes https://github.com/elastic/opex-product/issues/784
## Generative AI disclosure

- [x] Yes, cursor in auto mode  
- [ ] No


shainaraskas (approved):
I won't block this, but I don't feel like the FAQ is an appropriate place for a happy-path piece of info. would pull this onto its own page or similar

otherwise g2g

wajihaparvez:
That's true, it would be better on its own page. Discussing options with Arnon based on how imp this info is for customers

arnonst:
@wajihaparvez Lets finalize this on Monday as the new regions are already live in production.

PR #6818, opened 2026-06-03
More S3-compatibility rewording

DaveCTurner:
Mention conformance with _documented_ behaviour of S3 as a requirement,
and split the last few imperative sentences into a separate paragraph.


deploy-manage/tools/snapshot-and-restore/s3-repository.md
```diff
@@ -212,7 +212,9 @@ Some storage systems offer an S3-compatible API, and the `s3` repository type al
...
+The `s3` repository type requires your storage API to be _fully_ compatible with AWS S3. This means it must exactly conform to every aspect of the documented behavior of AWS S3: it must support all the same API endpoints, with all the same parameters, return the same errors in case of failures, and offer consistency, performance, and reliability at least as good as S3 even when accessed concurrently by multiple {{es}} nodes. Some storage products, including some from very well-known storage vendors, claim to offer a S3-compatible API but in fact only support a subset of the features of AWS S3. Others offer an API which matches that of AWS S3 but with different behavior, especially when accessed concurrently. Such storage products are not fully S3-compatible. Using a storage product which is not fully S3-compatible puts your snapshot data at risk. If you are using such a system for your snapshots, consider using a [shared filesystem repository](shared-file-system-repository.md) based on a standardized protocol such as NFS to access your storage system instead.
```
eedugon:
```suggestion
The `s3` repository type requires your storage API to be _fully_ compatible with AWS S3. This means it must exactly conform to every aspect of the documented behavior of AWS S3: it must support all the same API endpoints, with all the same parameters, return the same errors in case of failures, and offer consistency, performance, and reliability at least as good as AWS S3 even when accessed concurrently by multiple {{es}} nodes. Some storage products, including some from very well-known storage vendors, claim to offer a S3-compatible API but in fact only support a subset of the features of AWS S3. Others offer an API which matches that of AWS S3 but with different behavior, especially when accessed concurrently. Such storage products are not fully S3-compatible. Using a storage product which is not fully S3-compatible puts your snapshot data at risk. If you are using such a system for your snapshots, consider using a [shared filesystem repository](shared-file-system-repository.md) based on a standardized protocol such as NFS to access your storage system instead.
```

eedugon (approved):
LGTM!

For the final sentence (`Don't report {{es}} issues...`) one alternative could be:

> When reporting an {{es}} issue for a storage system claiming S3 compatibility, provide evidence that the same issue reproduces with a genuine AWS S3 repository.

Just for your consideration if you want the sentence to be shofter.

deploy-manage/tools/snapshot-and-restore/s3-repository.md
eedugon: ^^ Difficult to see, but I'm adding `AWS` in the part that says `as good as S3`

PR #6821, opened 2026-06-03
ECE deployment scenarios - fix proxies location in medium deployment

eedugon:
Updated medium deployment requirements to include proxies together with control plane nodes.

Medium deployments, per the image and per the sentence about keep a close eye to proxy resources utilization, do NOT have dedicated proxies. The proxies are collocated with coordinators/directors.


shainaraskas (approved):
good catch

PR #6856, opened 2026-06-04
Cross-link nodes out of disk to watermark

stefnestor:
## Summary

👋 BabyPR, to link [fix data nodes out of disk](https://www.elastic.co/docs/troubleshoot/elasticsearch/fix-data-node-out-of-disk) over to [watermark errors](https://www.elastic.co/docs/troubleshoot/elasticsearch/fix-watermark-errors). The latter is linked from ES code. I'm not sure if the former needs to exist or if page could collapse into second, but very least should cross link :)

## Generative AI disclosure

- [ ] Yes  
- [X] No


troubleshoot/elasticsearch/fix-data-node-out-of-disk.md
```diff
@@ -12,7 +12,9 @@ products:
 
 # Fix data nodes out of disk [fix-data-node-out-of-disk]
 
-{{es}} is using data nodes to distribute your data inside the cluster. If one or more of these nodes are running out of space, {{es}} takes action to redistribute your data within the nodes so all nodes have enough available disk space. If {{es}} cannot facilitate enough available space in a node, then you can intervene in one of two ways:
+{{es}} is using data nodes to distribute your data inside the cluster. If one or more of these nodes are running out of space, {{es}} takes action to redistribute your data within the nodes so all nodes have enough available disk space. Refer also to [Watermark errors](/troubleshoot/elasticsearch/fix-watermark-errors.md) troubleshooting.
```
yetanothertw: I think we need to add a bit more context around the link we're including (like a type of bridge connecting the two concepts). 

Perhaps something like (feel free to adjust if anything is incorrect or inaccurate):

> If {{es}} is struggling to redistribute data as a node fills up, you may also encounter watermark errors, which are cluster-level alerts that trigger at specific disk usage thresholds (85%, 90%, and 95%) and progressively restrict shard allocation, rebalancing, and write operations. Refer to [Watermark errors](https://www.elastic.co/docs/troubleshoot/elasticsearch/fix-watermark-errors) for guidance on monitoring disk usage.

Also, in terms of placement on the page, I think I'd add this at the very end, just after you list the two steps on this page. Otherwise, it looks like we're jumping around before the reader even gets to evaluate the context. 
This might work as a tip type of admonition as well.

yetanothertw (commented):
Thanks for flagging this, Stef. The [Data nodes out of disk](https://www.elastic.co/docs/troubleshoot/elasticsearch/fix-data-node-out-of-disk) page seems to work like an intro to its two child pages, I wonder if those can be combined? (not part of this PR but as a future enhancement or something)

As for the link in this PR, I think we just need a small intro to the link and why we're placing it here (context).

troubleshoot/elasticsearch/fix-data-node-out-of-disk.md
stefnestor: "If one or more nodes are running out of space" is synonymous to industry concept of disk "watermarks" 🙂. You are guaranteed the escalation pattern will climb hitting watermarks (linked+this page) followed by full disk (this page). 

(I guess my point is that there's like 5 disk pages all basically about disk filling up and response options & the TLDR page we wrote and is linked from ES code was the "watermark errors" one, so I'd like to cross link them at very least. But I don't mean to prescribe for your team but would suggest I'm not sure why all these other duplicate pages exist.)

I don't think it answers your appropriate framing pain point, but example alternative combining your idea (thanks!) & that caveat:
```suggestion
{{es}} is using data nodes to distribute your data inside the cluster. If one or more of these nodes are running out of space, {{es}} takes action to redistribute your data within the nodes so all nodes have enough available disk space. This is known as hitting disk watermarks and you can refer to [troubleshooting watermark errors](/troubleshoot/elasticsearch/fix-watermark-errors.md) for additional information. If {{es}} cannot facilitate enough available space in a node, then disk capacity might be exhausted. In this case, you can intervene in one of two ways:
```

shainaraskas: > But I don't mean to prescribe for your team but would suggest I'm not sure why all these other duplicate pages exist.

someone created them long ago and we are carrying the debt. we have an [issue](https://github.com/elastic/docs-content-internal/issues/1067) open to clean this up, but it languishes in the backlog

stefnestor: Ah, thanks for the super-high-level context! 

I'm happy to take a stab at disk/watermark deduplication specifically after this PR if it's okay with y'all? 

And if y'all don't have technical requirements for these page links/anchors/references continuing to exist? (Specifically, I only know how to check ES errors linking to docs, but please LMK if there's any other dependencies I should be aware of.)

shainaraskas: thanks for considering taking this on. as you hinted at, merging/deleting pages has some additional considerations and constraints, and is also a bit of a PITA, so likely a member of the docs team that is familiar with these procedures will need to undertake the work.

yetanothertw: Thank you for explaining Stef. I've slightly tweaked the suggestion and applied it. Will approve now.

yetanothertw (approved):
Thank you for adding the missing context 🪴

PR #6923, opened 2026-06-15
Mention bucket policies in S3 docs

DaveCTurner:
We don't talk about bucket policies today and users may mistakenly
assume that this means they are free to impose additional restrictions
via such a policy. This isn't the case. This commit calls this out in
the docs.

### DIFF
--- a/deploy-manage/tools/snapshot-and-restore/s3-repository.md
+++ b/deploy-manage/tools/snapshot-and-restore/s3-repository.md
@@ -174,6 +174,9 @@ You may further restrict the permissions by specifying a prefix within the bucke
 
 The bucket needs to exist to register a repository for snapshots. If you did not create the bucket then the repository registration will fail.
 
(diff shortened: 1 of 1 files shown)


deploy-manage/tools/snapshot-and-restore/s3-repository.md
```diff
@@ -174,6 +174,9 @@ You may further restrict the permissions by specifying a prefix within the bucke
 
 The bucket needs to exist to register a repository for snapshots. If you did not create the bucket then the repository registration will fail.
 
+You can also impose similar restrictions using a [bucket policy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucket-policies.html).
+
+Do not impose any additional restrictions on access to your bucket, for instance by using more restrictive `Condition` clauses than the ones documented above.
```
shainaraskas: would it be helpful to the user to hint at why we don't want them to do this? what might be the outcome of adding such restrictions?

shainaraskas (approved):
lgtm with one question

PR #6924, opened 2026-06-15
Refine claim about MPU fallback for S3 registers

DaveCTurner:
Today we say that if S3-like storage doesn't support conditional writes
then `it will work correctly with {{es}}` (with some extra caveats).
This isn't true, we require overwrite protection to avoid repository
corruption even if linearizable registers work ok with unconditional
MPUs.

This commit adjusts the docs to match.


shainaraskas (approved):

shainaraskas:
(good to go, ty)

PR #6930, opened 2026-06-16
Add explicit workflow write permissions for GITHUB_TOKEN

theletterf:
## Summary
- add explicit `contents: write` and `pull-requests: write` permissions to `update-kube-stack-version.yml`
- add explicit `issues: write` permissions to `add-new-team-label.yml`
- scope write permissions to the owning jobs while keeping workflow-level permissions denied by default

Contributes to elastic/docs-actions#187.

## Test plan
- [x] Reviewed both workflows to confirm they write through `GITHUB_TOKEN`
- [x] Verified the diff only adds the required explicit permissions
- [ ] Optional: run the workflows in GitHub after merge or in a test branch if you want runtime confirmation

### DIFF
--- a/.github/workflows/add-new-team-label.yml
+++ b/.github/workflows/add-new-team-label.yml
@@ -4,6 +4,9 @@ on:
   issues:
     types: [opened]
 
+permissions:
+  issues: write
+
 jobs:
   add-needs-triage-label:
     name: Add `needs-team` label
--- a/.github/workflows/add-new-team-label.yml
+++ b/.github/workflows/add-new-team-label.yml
@@ -4,13 +4,14 @@ on:
   issues:
     types: [opened]
 
-permissions:
-  issues: write
+permissions: {}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/add-new-team-label.yml
```diff
@@ -4,6 +4,9 @@ on:
   issues:
     types: [opened]
 
+permissions:
```
reakaleek: I think the preferred pattern is to have a top-level

```yml
permissions: {}

# or 

permissions:
  contents: read
```

And only pass `write` permissions at the job level so that we don't leak unnecessary permissions to new jobs in the future.

reakaleek (approved):

.github/workflows/add-new-team-label.yml
theletterf: Good point!

reakaleek (approved):

PR #7060, opened 2026-06-22
Deprecate internal legacy monitoring and flag for removal in v10. 

mosche:
Relates to elastic/dev#2744, elastic/elasticsearch#151617


shainaraskas (approved):
I think we also need to make the same change here:

https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7060/deploy-manage/monitor/stack-monitoring/kibana-monitoring-legacy

otherwise lgtm 👍

PR #7067, opened 2026-06-22
Document kuery (KQL) filtering and improve Fleet API page structure

### LINKED ISSUE #3401: [Feedback]: Kuery is nowhere described or explictily mapped or matched.
https://www.elastic.co/docs/api/doc/kibana/operation/operation-get-fleet-package-policies

explains to use the `Kuery` but doesn't explain what it is, how it is used, what it needs to look like. Maybe there is some documentation somewhere, at the very least it should get back linked.

charlotte-hoblik:
## Summary

Adds KQL filtering guidance to the Fleet API page and brings the page in line with the how-to content type. Closes #3401.

Connected issue: https://github.com/elastic/docs-content/issues/3401

**kuery / KQL filtering (the issue)**
- New **Filter results with KQL** section: which list endpoints accept `kuery`, the saved-object field-prefix rule, a verified `package_policies` example, and links to the KQL reference. The API reference names `kuery` but never explains it or links out, this is the missing on-ramp.

 **Structure and navigation (applied the Elastic docs skills)**
- Rewrote the page opening to state purpose and scope, and added a `description` to the frontmatter (no longer duplicated by the intro).
- Added a **Before you begin** prerequisites section, plus **Next steps** and **Related pages** sections.
- Folded the standalone **Authentication** section into the prerequisites.

**Copy edits throughout**
- Active voice, concision, and punctuation; removed the relative-time word "currently"; replaced a directional "previous example" reference with a direct link; fixed an unclosed quote in a `curl` example.

- [x] Yes  
- [ ] No  
2. If you answered "Yes" to the previous question, please specify the tool(s) and model(s) used (e.g., Google Gemini, OpenAI ChatGPT-4, etc.).

Tool(s) and model(s) used:
Claude Opus 4.8
Elastic docs review skills


reference/fleet/fleet-api-docs.md
```diff
@@ -11,34 +12,39 @@ products:
...
+* Authentication credentials for {{kib}} API requests. API key authentication is recommended. For details, refer to [Authentication]({{kib-apis}}authentication).
+* A {{kib}} user with the required {{fleet}} and {{integrations}} privileges. For details, refer to [Roles and privileges](/reference/fleet/fleet-roles-privileges.md).
 
 
 ## Using the Console [using-the-console]
 
 You can run {{fleet}} API requests through the {{kib}} Console.
 
-1. Open the {{kib}} menu and go to **Management → Dev Tools**.
+1. From the {{kib}} menu, go to **Management → Dev Tools**.
```
bmorelli25: I don't think this wording aligns with our guidance on [referring to apps and pages](https://docs.elastic.dev/tech-writing-guidelines/ui-writing#referring-to-apps-and-pages).

reference/fleet/fleet-api-docs.md
```diff
@@ -461,24 +467,40 @@ curl -X GET 'http://<user>:<pass>@<kibana url>/api/fleet/agents?perPage=10000&se
 ```
 
 
+## Filter results with KQL [filter-results-with-kql]
+
+The {{fleet}} list endpoints for agents, agent policies, package policies, and enrollment tokens accept a `kuery` query parameter that takes a [{{kib}} Query Language (KQL)](elasticsearch://reference/query-languages/kql.md) string to filter results. To check whether another endpoint accepts `kuery`, refer to its parameters in the [Kibana API docs]({{kib-apis}}).
+
+Reference each field by the saved object type that stores it. {{fleet}} stores package policies as `ingest-package-policies` objects, so to filter package policies by integration name, query the `ingest-package-policies.package.name` field.
```
bmorelli25: I'm pretty sure this isn't true. Agents and enrollment tokens are not saved objects. They are stored in `.fleet-*` system indices so they must be referenced directly with no saved object prefix.

Your example for **package policies** works in [redacted] test cluster (I'll share w/you on Slack).
```
GET kbn:/api/fleet/package_policies?kuery=ingest-package-policies.package.name:activemq
```
✅ 
```
{
  "items": [
    {
      "id": "9xxxxa3-b6b2-4xxx-8f20-bb9822xxxxxe",
      "version": "WxxxxxxxxxQ==",
      "name": "activemq-1",
...
```
However, appending the saved object type to an **agent** call does not work:

```
GET kbn:/api/fleet/agents?kuery=ingest-agents.status:online
```

```
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": "[request query.kuery]: KQLSyntaxError: This key 'ingest-agents.status' does NOT exist in fleet-agents saved object index patterns"
}
```

Excluding the saved object type from an **agent** call does work:

```
GET kbn:/api/fleet/agents?kuery=status:online
```
✅ 
```
{
  "items": [
    {
      "id": "3xxxx1-4a0a-4b00-bxxxx20xxxxx9d4",
      "type": "PERMANENT",
      "active": true,
...
```

bmorelli25: @vishaangelova keep me honest here

reference/fleet/fleet-api-docs.md
```diff
@@ -489,6 +511,25 @@ These limitations apply for the manual rollback feature:
...
+* [Create an agent policy without using the UI](/reference/fleet/create-policy-no-ui.md) for automation use cases
+
+
+## Related pages [fleet-api-related-pages]
 
-- the rollback window has ended (typically more than seven days). When the rollback window ends, the files from the previous version are removed to free up disk space. 
+* [Kibana API docs]({{kib-apis}})
+* [Run API requests](/explore-analyze/query-filter/tools/console.md)
+* [Roles and privileges](/reference/fleet/fleet-roles-privileges.md)
+* [Fleet enrollment tokens](/reference/fleet/fleet-enrollment-tokens.md)
```
bmorelli25: This is the same link that's right above this section. I'm not sure how useful it is to list it twice.

bmorelli25 (changes requested):

reference/fleet/fleet-api-docs.md
vishaangelova: You mean this one: https://www.elastic.co/docs/contribute-docs/style-guide/ui-writing#referring-to-apps-and-pages

reference/fleet/fleet-api-docs.md
vishaangelova: Ok, so actually only these two Fleet endpoints are backed by saved objects:
- Agent policies (`GET /api/fleet/agent_policies`): stored as a `fleet-agent-policies` saved object, or, on deployments where space awareness is off, as the legacy `ingest-agent-policies` saved object ([#1](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/common/constants/agent_policy.ts#L8-L9) and [#2](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/server/services/agent_policy.ts#L203-L207)).
  - Use `fleet-agent-policies.<field>` in the KQL prefix; the legacy `ingest-agent-policies.<field>` also works because Fleet normalizes either prefix at runtime ([#3](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/server/services/agent_policy.ts#L183-L201))
- Package policies (`GET /api/fleet/package_policies`): stored as a `fleet-package-policies` saved object, or, on deployments where space awareness is off, as the legacy `ingest-package-policies` saved object ([#1](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/common/constants/package_policy.ts#L8-L10), and [#2](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/server/services/package_policy.ts#L291-L295)).
  - Use `fleet-package-policies.<field>`, but the legacy `ingest-package-policies.<field>` also works for the same reason as above ([#3](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/server/services/package_policy.ts#L338-L356)).

Agents (`GET /api/fleet/agents`) are stored in the `.fleet-agents` ES system index ([#1](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/common/constants/agent.ts#L8-L26)). So these should be referenced without a prefix (e.g., `status:online`). However, here `fleet-agents.status:online` also works because Fleet strips it from the kuery ([#2](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/server/services/agents/crud.ts#L85-L87)), but `ingest-agents.` does not exist, and that returns an error.

Enrollment API keys (`GET /api/fleet/enrollment_api_keys`) are stored in the `.fleet-enrollment-api-keys` ES system index ([#1](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/common/constants/enrollment_api_key.ts#L8)), but here no prefix should be used as the kuery is passed straight through to the index search (e.g., `policy_id:"<policy-id>"`).

charlotte-hoblik: Thanks so much, both of you, this was really helpful. I tested it all on a 9.4.2 cluster and your breakdown held up, with one surprise.

fleet-agent-policies.<field> doesn't actually work on 9.4.2. It 400s, and only ingest-agent-policies.<field> or a bare name:"..." work, even though the data is stored under the fleet-* types (confirmed with a type agg on .kibana*). 

In the doc I went with bare fields where possible, prefixed only nested fields like package.name, and added a tip that the KQLSyntaxError names the exact pattern to use.

@vishaangelova does fleet-agent-policies.<field> work for you on main? Thanks again.

bmorelli25: I can confirm that `fleet-agent-policies.name` 400s and the following actually work:

```
GET kbn:/api/fleet/agent_policies?kuery=ingest-agent-policies.name:"Elastic Cloud agent policy"
GET kbn:/api/fleet/agent_policies?kuery=name:"Elastic Cloud agent policy"
```

Tested on 9.4.3 (unreleased)

reference/fleet/fleet-api-docs.md
```diff
@@ -461,24 +467,51 @@ curl -X GET 'http://<user>:<pass>@<kibana url>/api/fleet/agents?perPage=10000&se
...
+
+```shell
+curl --request GET \
+  --url 'https://my-kibana-host:9243/api/fleet/package_policies?kuery=ingest-package-policies.package.name:nginx' \
+  --header 'Authorization: ApiKey yourbase64encodedkey' \
+  --header 'kbn-xsrf: xx'
+```
+
+::::{tip}
+The accepted saved object type prefix can differ across deployments and endpoints. If a query returns a `KQLSyntaxError` stating that a key does not exist in a saved object index pattern, use the index pattern named in the error as your prefix. For example, if the error names `ingest-agent-policies`, query `ingest-agent-policies.name`.
```
bmorelli25: I like this tip but the current wording only handles the KQLSyntaxError that names the index pattern (for example, `ingest-agents.status` on the agents endpoint). Forgetting the prefix on a nested field produces a different error that does not have this pattern:

```
GET kbn:/api/fleet/package_policies?kuery=package.name:nginx
```

```
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": "This type package is not allowed: Bad Request"
}
```

Here the parser reads package as a saved object type and rejects it, so there's no index pattern in the error to copy. Could we broaden the tip to cover both errors? Something like:

```suggestion
If a query returns an error like `This type <x> is not allowed` or a `KQLSyntaxError` about a missing key, the field needs the saved object type as a prefix. If the error names an index pattern, use that; otherwise prefix with the endpoint's saved object type (for example, `ingest-package-policies` for package policies).
```

bmorelli25 (approved):
ty! 🚢

reference/fleet/fleet-api-docs.md
vishaangelova: Hi @paul-tavares, pulling you in this thread to confirm if the way this works is by design:

On `GET /api/fleet/agent_policies`, the kuery validator rejects the `fleet-agent-policies.<field>` prefix (returns a 400 `KQLSyntaxError`), so only `ingest-agent-policies.<field>` or bare fields work. On `GET /api/fleet/package_policies` the validator was widened to accept both prefixes as part of https://github.com/elastic/kibana/pull/227862, but the equivalent change doesn't seem to have been made for agent policies. Wondering if this was intentional, so we can document the right prefix?

vishaangelova: Sorry, after checking with Paul, it seems maybe @criamico could help instead?

criamico: > On GET /api/fleet/agent_policies, the kuery validator rejects the fleet-agent-policies.<field> prefix (returns a 400 KQLSyntaxError), so only ingest-agent-policies.<field> or bare fields work. 

I just checked the code and verified that the [KQL validator](https://github.com/elastic/kibana/blob/8be46b300b632f4562d4df86a48793464cadfd1f/x-pack/platform/plugins/shared/fleet/server/types/rest_spec/agent_policy.ts#L44-L45) rejects any kuery prefixed with fleet-agent-policies. This was probably missed when doing the work for space awareness. It is actually a bug, so I'll open a ticket to track it.

criamico: Link to ticket for the KQL bug: https://github.com/elastic/kibana/issues/275810

reference/fleet/fleet-api-docs.md
charlotte-hoblik: Good catch. Broadened the tip to name both errors. Thanks!

PR #7097, opened 2026-06-24
docs: document workflow loop guardrails

rosomri:
## Summary

- Document loop-level guardrails for `foreach` and `while` workflow steps: `max-iterations`, `timeout`, `on-failure`, and `if`.
- Document iteration-level guardrails: `iteration-timeout` and `iteration-on-failure`.
- Update the flow-control overview and error-handling restrictions so they align with the new loop guardrail behavior.

## Testing

- Checked diagnostics for the edited docs files with `ReadLints`.
- Reviewed the final git diff before committing.


explore-analyze/workflows/authoring-techniques/pass-data-handle-errors.md
```diff
@@ -198,9 +198,9 @@ In the following example:
 
 ### Restrictions [workflows-on-failure-restrictions]
 
-- Flow-control steps (`if`, `foreach`) cannot have workflow-level `on-failure` configurations.
 - Fallback steps run only after all retries have been exhausted.
```
charlotte-hoblik, line 202:
```suggestion
- The `if` step cannot have an `on-failure` configuration. (`foreach` and `while` loops do support loop-level `on-failure` — see their step references.)
- Fallback steps run only after all retries have been exhausted.
```

Removing the original bullet also dropped a restriction that's still accurate for `if`:
in the workflow schema, `if` steps have no `on-failure` field , only `foreach`/`while` gained loop-level `on-failure` in 9.4. 
Re-adding it so the page doesn't silently lose that information.

explore-analyze/workflows/steps/foreach.md
```diff
@@ -39,6 +48,45 @@ steps:
...
+## Guardrails
+
+```{applies_to}
+stack: ga 9.4+
+serverless: ga
+```
+
+Use loop-level guardrails to control the `foreach` step as a whole:
+
+* `max-iterations` prevents unexpectedly large collections from creating runaway loops. A bare number is treated as `{ limit: N, on-limit: continue }`, so the loop succeeds when the limit is reached. Use the object form with `on-limit: fail` when reaching the cap should fail the workflow.
```
charlotte-hoblik:
```suggestion
* `max-iterations` caps how many items the loop processes and defaults to **2000** with `on-limit: continue`, so a larger collection is silently truncated unless you raise the limit or set `on-limit: fail`. A bare number is shorthand for `{ limit: N, on-limit: continue }`; use the object form with `on-limit: fail` to fail the workflow when the cap is reached.
```

foreach shares the same default `max-iterations: 2000` / `on-limit: continue` as while, so a foreach over a >2000-item array is silently truncated unless you raise the limit. `while.md` already warns about this: this brings the foreach guardrails section in line.

explore-analyze/workflows/steps/flow-control-steps.md
```diff
@@ -54,12 +54,16 @@ For expression syntax and additional examples, refer to [If step](/explore-analy
 
 ## `foreach` [foreach]
 
-Iterate over an array, running nested steps once per item. Inside the loop, the current item is available as `foreach.item`, the zero-based position as `foreach.index`, and the total count as `foreach.total`.
+Iterate over an array, running nested steps once per item. Inside the loop, the current item is available as `foreach.item`, the zero-based position as `foreach.index`, and the total count as `foreach.total`. Use `max-iterations`, `timeout`, `on-failure`, and `if` to guard the loop as a whole; use `iteration-timeout` and `iteration-on-failure` to guard each iteration.
```
charlotte-hoblik:
```suggestion
Iterate over an array, running nested steps once per item. Inside the loop, the current item is available as `foreach.item`, the zero-based position as `foreach.index`, and the total count as `foreach.total`. It also supports guardrails to cap iterations, set timeouts, and handle failures.
```

See comment on https://github.com/elastic/docs-content/pull/7097/changes#r3472914184

explore-analyze/workflows/steps/flow-control-steps.md
```diff
@@ -71,7 +75,7 @@ For the full parameter reference, refer to [Foreach step](/explore-analyze/workf
 
 ## `while` [while]
 
-Loop while a KQL condition evaluates to true. The `max-iterations` field caps the number of iterations and defaults to **2000**. The default `on-limit` behavior is `continue`, which means the step succeeds quietly when the cap is reached. To fail the workflow on the cap instead, use the object form with `on-limit: fail`.
+Loop while a KQL condition evaluates to true. The `max-iterations` field caps the number of iterations and defaults to **2000**. The default `on-limit` behavior is `continue`, which means the step succeeds quietly when the cap is reached. To fail the workflow on the cap instead, use the object form with `on-limit: fail`. Like `foreach`, `while` also supports loop-level `timeout`, `on-failure`, and `if`, plus per-iteration `iteration-timeout` and `iteration-on-failure`.
```
charlotte-hoblik:
```suggestion
Loop while a KQL condition evaluates to true. The `max-iterations` field caps the number of iterations and defaults to **2000**. The default `on-limit` behavior is `continue`, which means the step succeeds quietly when the cap is reached. To fail the workflow on the cap instead, use the object form with `on-limit: fail`. Like `foreach`, it also supports loop-level guardrails and per-iteration controls.
```

In an overview, listing every field tends to drift out of sync with the reference. Since each blurb already links the full step reference, I trimmed the enumerations to a
one-line nod. Ref: https://www.elastic.co/docs/contribute-docs/content-types/overviews

explore-analyze/workflows/steps/while.md
```diff
@@ -30,14 +30,31 @@ Use `while` for polling patterns: checking a status until it reaches `ready`, re
...
 
 :::{warning}
 `while` defaults to `max-iterations: 2000` with `on-limit: continue`, which means the step **succeeds quietly when the cap is reached**. If your loop needs to fail the workflow on hitting the cap, set `on-limit: fail` explicitly. Always think about the cap on loops that depend on external state to avoid runaway executions or silently truncated work.
 :::
 
+## Guardrails
+
+Use loop-level guardrails to control the `while` step as a whole:
+
+* `max-iterations` prevents runaway loops. The default is **2000** iterations. A bare number is treated as `{ limit: N, on-limit: continue }`, so the loop succeeds when the limit is reached. Use the object form with `on-limit: fail` when reaching the cap should fail the workflow.
```
charlotte-hoblik:
```suggestion
* `max-iterations` caps the number of iterations to prevent runaway loops.
```
The default (2000) and the bare-number-vs-object / `on-limit` behavior are already covered just below in the warning and the `max-iterations` shape section, so I trimmed this bullet to keep a single source of truth on the page. 
(Consolidating the pre-existing repetition between the warning and the shape section is out of scope here — happy to file that as a follow-up.)

charlotte-hoblik (commented):
Thank you, 🙂 clean change, and the `applies_to` tagging is spot on. I left a few inline suggestions.

One thing that's better as a manual move than a suggestion: 

Could we move the `## Guardrails` section to sit between `## Context variables` and `## Example`? Where it is now, its `ga 9.4+` badge also covers the pre-9.4 "expression types" intro that follows it.

charlotte-hoblik (approved):

Issue #1740, opened 2025-06-16
Restructure and split the 'kNN search' documentation into sub-pages


kosabogi:
The current [kNN search documentation page](https://www.elastic.co/docs/solutions/search/vector/knn) has grown significantly in length and complexity, which can make it difficult for users to navigate and digest the content.

Split the content into smaller, clearly structured sub-pages to make it easier to read and maintain.

PR #7138, opened 2026-06-29
Restructures kNN documentation into a clearer page hierarchy

kosabogi:
## Summary

Restructures the kNN documentation from one long page into a focused hierarchy. The new structure separates method selection, approximate and exact kNN, query construction, nested kNN, and performance optimization, while preserving existing published links through redirects.

### What's new

| Page / file | What changed |
| --- | --- |
| [Approximate kNN search](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/solutions/search/vector/knn/approximate-knn) | New focused page covering approximate kNN methods, a basic example, mapping defaults, indexing considerations, and vector index mode. |
| [Build approximate kNN search queries](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/solutions/search/vector/knn/build-search-queries) | Consolidates query-vector creation, semantic and similar-document search, filtering, similarity thresholds, hybrid search, multiple vector fields, and aggregations. |
| [Exact kNN search](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/solutions/search/vector/knn/exact-knn) | Moves exact kNN guidance into a dedicated page and documents both the `dense_vector` query and `script_score` approaches. |
| [Nested kNN search](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/solutions/search/vector/knn/nested-knn-search) | Consolidates nested vector search examples, including metadata and sibling filtering, inner hits, and chunked content retrieval. |
| [Optimize approximate kNN performance and accuracy](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/solutions/search/vector/knn/optimize-performance-accuracy) | Consolidates speed and accuracy tuning, HNSW and DiskBBQ controls, vector storage options, quantization, and oversampling and rescoring guidance. |

### What's changed

| Page / file | What changed |
| --- | --- |
| [kNN search in Elasticsearch](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/solutions/search/vector/knn) | Reworks the former long-form page into a landing page focused on use cases, prerequisites, method selection, and links to the new child pages. |
| [Tune approximate kNN search](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/deploy-manage/production-guidance/optimize-performance/approximate-knn-search) | Expands capacity-planning guidance with an interactive vector sizing calculator, Lucene vector-file reference, separate off-heap RAM and disk estimates, HNSW and DiskBBQ formulas, and file-preloading recommendations. |
| [Generative AI search reference architecture](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/deploy-manage/reference-architectures/genai-search-high-availability) | Points capacity planning to the updated calculator and disk and off-heap RAM formulas. |
| [Cross-cluster search](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/explore-analyze/cross-cluster-search) | Updates the approximate kNN limitation link to the new canonical page. |
| [NLP model reference](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/explore-analyze/machine-learning/nlp/ml-nlp-model-ref) | Updates the approximate kNN link to the new canonical page. |
| [Text embedding vector search example](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/explore-analyze/machine-learning/nlp/ml-nlp-text-emb-vector-search-example) | Updates the semantic kNN link to the new query-building page. |
| [Dense versus sparse ingest pipelines](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/solutions/search/semantic-search/dense-versus-sparse-ingest-pipelines) | Updates the dense-vector kNN link to the semantic search example on the new query-building page. |
| [Semantic search with inference](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/solutions/search/semantic-search/semantic-search-inference) | Updates the dense-vector semantic search link to the new query-building page. |
| [Vector search in Elasticsearch](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/solutions/search/vector) | Aligns the kNN guide link text with the restructured landing page. |
| [Vector search use cases](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7138/solutions/search/vector/vector-search-use-cases) | Updates the filtered kNN link to the new query-building page. |
| `redirects.yml` | Redirects anchors from the former monolithic kNN page to their new child pages and preserves renamed query-page anchors. |
| `solutions/toc.yml` | Adds the new kNN page hierarchy to the Search navigation. |

Related issue: [#1740](https://github.com/elastic/docs-content/issues/1740)  
Parent issue: [elastic/docs-content-internal#801](https://github.com/elastic/docs-content-internal/issues/801)


solutions/search/vector/knn/exact-knn.md
```diff
@@ -0,0 +1,91 @@
+---
+navigation_title: Exact kNN search
```
mayya-sharipova, line 2: Suggest to call "Exact NN" instead of "Exact kNN" in the name of files and in documentation, as we don't have "k" parameter in this query.

solutions/search/vector/knn/exact-knn.md
```diff
@@ -0,0 +1,91 @@
...
+          }
+        }
+      }
+    }
+    ```
+
+## Resources
+
+- [Approximate kNN search](approximate-knn.md): Learn how to map, index, and run fast, scalable approximate kNN search for most production workloads.
+- [Build search queries](build-search-queries.md): Learn how to construct approximate kNN queries for filtering, hybrid retrieval, semantic search, multiple vector fields, and similarity thresholds.
```
mayya-sharipova: I think better name for that is "kNN query patterns"

solutions/search/vector/knn/approximate-knn.md
```diff
@@ -0,0 +1,148 @@
+---
+navigation_title: Approximate kNN search
+description: Run fast, scalable approximate k-nearest neighbor (kNN) vector search in Elasticsearch using dense_vector fields and HNSW indexing.
```
mayya-sharipova: why only HNSW, we also have other methods.

Suggestion: " Run fast, scalable approximate k-nearest neighbor (kNN) vector search in Elasticsearch, including search methods, indexing considerations, and limitations."

solutions/search/vector/knn.md
```diff
@@ -1,1457 +1,77 @@
...
-  - id: elasticsearch
-  - id: cloud-serverless
 ---
 
-# kNN search in {{es}} [knn-search]
+# kNN search on {{es}} [knn-search]
 
-A *k-nearest neighbor* (kNN) search finds the *k* nearest vectors to a query vector using a similarity metric such as cosine or L2 norm.  
-With {{es}} kNN search, you can retrieve results based on semantic meaning rather than exact keyword matches.
+A *k-nearest neighbor* (kNN) search finds the *k* nearest vectors to a query vector using a similarity metric such as cosine or L2 norm. In {{es}}, kNN is the primary way to query [`dense_vector`](dense-vector.md) fields after you store embeddings.
```
lcawl, line 14: > In {{es}}, kNN is the primary way to query [`dense_vector`](dense-vector.md) fields after you store embeddings.

Seeing "primary" made me wonder what the other ways are, since only knn was mentioned in the parent page. Are there other methods that ought to have been linked from that parent?

solutions/search/vector/knn.md
```diff
@@ -1,1457 +1,77 @@
...
+  - Refer to [Approximate kNN search](knn/approximate-knn.md) for mapping, indexing, and search examples.
 
-### Choosing between approximate and exact kNN
+- **Exact, brute-force kNN**
+  - Best for small datasets, pre-filtered subsets, or when you need precise scoring without approximate indexing.
+  - Uses a [`script_score` query](elasticsearch://reference/query-languages/query-dsl/query-dsl-script-score-query.md) with a vector function to compute similarity against each matching document.
+  - Refer to [Exact kNN search](knn/exact-knn.md) for search examples.
 
 Approximate kNN offers low latency and good accuracy, while exact kNN guarantees accurate results but does not scale well for large datasets. With exact kNN, a `script_score` query must scan each matching document to compute the vector function, which can result in slow search speeds. However, you can improve latency by using a [query](../../../explore-analyze/query-filter/languages/querydsl.md) to limit the number of matching documents passed to the function. If you filter your data to a small subset of documents, you can get good search performance using this approach.
```
lcawl: I know you weren't updating this paragraph, but while reviewing I noticed two approaches were mentioned and I wasn't sure which "this approach" referred to by the end. Might be worth a small improvement?
```suggestion
Approximate kNN offers low latency and good accuracy, while exact kNN guarantees accurate results but does not scale well for large datasets. With exact kNN, a `script_score` query must scan each matching document to compute the vector function, which can result in slow search speeds. However, you can improve latency by using a [query](../../../explore-analyze/query-filter/languages/querydsl.md) to limit the number of matching documents passed to the function. If you filter your data to a small subset of documents, you can get good search performance using this approach for exact kNN.
```

solutions/search/vector/knn/approximate-knn.md
```diff
@@ -0,0 +1,148 @@
...
+
+Approximate kNN search uses graph-based or clustered index structures to find similar vectors quickly at scale. Use it for most production workloads where low latency matters more than perfect recall. This page covers approximate kNN search methods, running a basic approximate kNN search, indexing considerations, and limitations.
+
+::::{tip}
+If you use `semantic_text` fields, query them with a [`match` query](elasticsearch://reference/query-languages/query-dsl/query-dsl-match-query.md) for the simplest approach, or use the [`knn` query](elasticsearch://reference/query-languages/query-dsl/query-dsl-knn-query.md#knn-query-with-semantic-text) when you need more control over the search.
+::::
+
+
+::::{warning}
+Approximate kNN search has specific resource requirements. For instance, for HNSW, all vector data must fit in the node’s page cache for efficient performance. Refer to the [approximate kNN tuning guide](/deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md) for configuration tips.
```
lcawl, line 19: These backticks are a pet peeve for me and easy cleanup while you're moving this content, but also fine to ignore this if you have bigger fish to fry :)
```suggestion
Approximate kNN search has specific resource requirements. For instance, for HNSW, all vector data must fit in the node's page cache for efficient performance. Refer to the [approximate kNN tuning guide](/deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md) for configuration tips.
```

solutions/search/vector/knn/approximate-knn.md
```diff
@@ -0,0 +1,148 @@
...
+          },
+          "file-type": {
+            "type": "keyword"
+          }
+        }
+      }
+    }
+    ```
+
+2. Index your data with embeddings.
```
lcawl: > 2. Index your data with embeddings.  

Something for future consideration: It would be nice to have a link to more info about how/where users get these values, since this example alone doesn't explain how you come up with those values. I assume it's a "bring your own vectors" context so maybe even just calling that out?

solutions/search/vector/knn/approximate-knn.md
```diff
@@ -0,0 +1,148 @@
...
+::::
+
+## Indexing considerations for approximate kNN search [knn-indexing-considerations]
+
+
+For approximate kNN, {{es}} stores dense vector values per segment as an [HNSW graph](https://arxiv.org/abs/1603.09320) or per segment as clusters using [DiskBBQ](https://www.elastic.co/search-labs/blog/diskbbq-elasticsearch-introduction). Building these approximate kNN structures is compute-intensive, which means indexing vectors can be time-consuming. As a result, you might need to increase client request timeouts for index and bulk operations. The [approximate kNN tuning guide](/deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md) covers indexing performance, sizing, and configuration trade-offs that affect search performance.
+
+{applies_to}`stack: ga 9.2` In addition to search-time parameters, HNSW and DiskBBQ expose index-time settings that balance graph build cost, search speed, and accuracy. When defining your `dense_vector` mapping, use [`index_options`](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md#dense-vector-index-options) to set these parameters:
+
+::::{tip}
```
lcawl: > ::::{tip}

IMO this tip should follow the example, since it distance s the explanation (ending with colon) from the example it's referring to.

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,238 @@
...
+    "num_candidates": 50,
+    "boost": 0.1
+  },
+  "size": 10
+}
+```
+
+This search finds the global top `k = 5` vector matches, combines them with the matches from the `match` query, and finally returns the 10 top-scoring results. The `knn` and `query` matches are combined through a disjunction, as if you took a boolean *or* between them. The top `k` vector results represent the global nearest neighbors across all index shards.
+
+The score of each result is the sum of the `knn` and `query` scores. You can specify a `boost` value to give a weight to each score in the sum. In the preceding example, the scores will be calculated as
```
lcawl: Feels like there's a word and/or colon missing:
```suggestion
The score of each result is the sum of the `knn` and `query` scores. You can specify a `boost` value to give a weight to each score in the sum. In the preceding example, the scores will be calculated as follows:
```

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,238 @@
...
+The filter is applied **during** approximate kNN search to ensure that `k` matching documents are returned. In contrast, post-filtering applies the filter **after** the approximate kNN step and can return fewer than `k` results, even when enough relevant documents exist.
+::::
+
+### Filtering behavior and performance [filtering-behavior-and-performance]
+
+In approximate kNN search with an HNSW index, applying filters can decrease performance as the engine must explore more of the graph to gather enough candidates that satisfy the filter and reach `num_candidates`. This contrasts with conventional query filtering, where stricter filters often speed up queries.
+
+To avoid significant performance drawbacks, Lucene implements the following strategies per segment:
+
+* If the filtered document count is less than or equal to num_candidates, the search bypasses the HNSW graph and uses a brute force search on the filtered documents.
```
lcawl:
```suggestion
* If the filtered document count is less than or equal to `num_candidates`, the search bypasses the HNSW graph and uses a brute force search on the filtered documents.
```

solutions/search/vector/knn/nested-knn-search.md
```diff
@@ -0,0 +1,670 @@
...
+applies_to:
+  stack:
+  serverless:
+---
+
+# Nested kNN search [nested-knn-search]
+
+Nested kNN search lets you find the most relevant passage or chunk inside long documents by storing a separate vector for each nested section and returning parent documents ranked by their best match. This approach is useful when a single document is too long to embed as one vector, such as when a support portal needs to surface the most relevant paragraph from a long troubleshooting guide in response to a user question.
+
+This page covers when to use nested kNN search, a basic mapping and query example, filtering, inner hits, and chunked content retrieval. For other approximate kNN query patterns, refer to [Build search queries](build-search-queries.md).
```
lcawl: Nit, this sentence reads a little unevenly to me:
```suggestion
This page covers when to use nested kNN search, filtering, inner hits, and chunked content retrieval and provides a basic mapping and query example. For other approximate kNN query patterns, refer to [Build search queries](build-search-queries.md).
```

solutions/search/vector/knn/optimize-performance-accuracy.md
```diff
@@ -0,0 +1,367 @@
+---
+navigation_title: Optimize performance and accuracy
+description: Tune approximate kNN search in Elasticsearch for speed, recall, vector storage, quantization, and rescoring trade-offs.
+applies_to:
+  stack:
+  serverless:
+---
+
+# Optimize performance and accuracy [optimize-knn-performance-accuracy]
```
lcawl: Unless you've been advised to prune this title for SEO reasons, I'd be inclined to add context:
```suggestion
# Optimize approximate kNN search performance and accuracy [optimize-knn-performance-accuracy]
```

solutions/search/vector/knn/optimize-performance-accuracy.md
```diff
@@ -0,0 +1,367 @@
...
+
+Approximate kNN search balances query speed, result quality, and resource use, and small configuration changes can have a large impact on latency, recall, and storage costs in production.
+
+This page covers trade-offs among search speed, recall, indexing cost, vector storage, quantization, and rescoring for approximate kNN search.
+
+## Tune approximate kNN for speed or accuracy [tune-approximate-knn-for-speed-accuracy]
+
+To gather results, the kNN API first finds a `num_candidates` number of approximate neighbors per shard, computes similarity to the query vector, selects the top `k` per shard, and merges them into the global top `k` nearest neighbors.
+
+For HNSW indices, `num_candidates` is the main search-time speed/accuracy control:
```
lcawl: I wasn't sure if this meant "or", "and" or "divided by" at first glance, so I think it should be spelled out:
```suggestion
For HNSW indices, `num_candidates` is the main search-time speed or accuracy control:
```

solutions/search/vector/knn/optimize-performance-accuracy.md
```diff
@@ -0,0 +1,367 @@
...
+
+For DiskBBQ (`bbq_disk`) indices, you can also use `visit_percentage` to control the total percentage of vectors visited during search. `visit_percentage` accepts values from `0` to `100`, including decimal values such as `0.5` (half a percent):
+
+* A good starting value is `3` (3%).
+* Increase `visit_percentage` to improve recall and accuracy (at the cost of higher latency).
+* Decrease `visit_percentage` for faster queries (with a potential accuracy trade-off).
+
+When a search targets both HNSW and DiskBBQ indices, use `visit_percentage` with `num_candidates` to tune performance and recall across both index types.
+
+When [quantization](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md#dense-vector-quantization) is involved, `rescore_vector` is an additional speed/accuracy tuning tool. It reranks a larger candidate set using original vectors after approximate retrieval.
```
lcawl: Ditto re "/":
```suggestion
When [quantization](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md#dense-vector-quantization) is involved, `rescore_vector` is an additional speed or accuracy tuning tool. It reranks a larger candidate set using original vectors after approximate retrieval.
```

solutions/search/vector/knn/optimize-performance-accuracy.md
```diff
@@ -0,0 +1,367 @@
...
+
+* Increase `rescore_vector.oversample` to improve accuracy (at the cost of higher latency).
+* Decrease `rescore_vector.oversample` for faster queries (with a potential accuracy trade-off).
+* For detailed behavior and usage guidance, see [Oversampling and rescoring for quantized vectors](#dense-vector-knn-search-rescoring).
+
+## Approximate kNN using byte vectors [approximate-knn-using-byte-vectors]
+
+The approximate kNN search API also supports `byte` (int8) value vectors alongside `float` vectors. Use the [`knn` option]({{es-apis}}operation/operation-search#operation-search-body-application-json-knn) to search a `dense_vector` field with [`element_type`](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md#dense-vector-params) set to `byte` and indexing enabled. Byte vectors reduce memory footprint and can improve cache efficiency for large-scale vector similarity search.
```
lcawl: "also" seems unnecessary here:
```suggestion
The approximate kNN search API supports `byte` (int8) value vectors alongside `float` vectors. Use the [`knn` option]({{es-apis}}operation/operation-search#operation-search-body-application-json-knn) to search a `dense_vector` field with [`element_type`](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md#dense-vector-params) set to `byte` and indexing enabled. Byte vectors reduce memory footprint and can improve cache efficiency for large-scale vector similarity search.
```

solutions/search/vector/knn/optimize-performance-accuracy.md
```diff
@@ -0,0 +1,367 @@
...
+        "query_vector": [-5, 9],
+        "k": 10,
+        "num_candidates": 100
+      },
+      "fields": [ "title" ]
+    }
+    ```
+
+
+*Note*: In addition to the standard byte array, one can also provide a hex-encoded string value for the `query_vector` param. As an example, the search request above can also be expressed as follows, which would yield the same results
```
lcawl: "you" is better than "one" IMO and "param" seems like an unnecessary abbreviation. "Above" is also one of the words we try to avoid per https://www.elastic.co/docs/contribute-docs/style-guide/accessibility#terms
```suggestion
*Note*: In addition to the standard byte array, you can also provide a hex-encoded string value for the `query_vector` parameter. As an example, the preceding search request can also be expressed as follows, which would yield the same results:
```

solutions/search/vector/knn/optimize-performance-accuracy.md
```diff
@@ -0,0 +1,367 @@
...
+
+## Byte quantized kNN search [knn-search-quantized-example]
+
+If you want to provide `float` vectors but still get the memory savings of `byte` vectors, use the [quantization](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md#dense-vector-quantization) feature. Quantization allows you to provide `float` vectors, but internally they are indexed as `byte` vectors. Additionally, the original `float` vectors are still retained in the index.
+
+::::{note}
+The default index type for `float` vectors is either `bbq_hnsw` or `int8_hnsw`, depending on your product version and vector dimensions. Other element types, such as `byte`, default to plain `hnsw` with no quantization. Refer to [Dense vector field type](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md).
+::::
+
+You can use the default quantization strategy or specify an index option.
```
lcawl: > You can use the default quantization strategy or specify an index option.

Should this be numbered as well? If not, I think the subsequent numbered list needs an introductory sentence.

solutions/search/vector/knn/optimize-performance-accuracy.md
```diff
@@ -0,0 +1,367 @@
...
+    }
+  },
+  "fields": [ "title", "file-type" ]
+}
+```
+
+This example will:
+
+* Search using approximate kNN for the top 100 candidates.
+* Rescore the top 20 candidates (`oversample * k`) per shard using the original, non quantized vectors.
```
lcawl: Should this be hyphenated?
```suggestion
* Rescore the top 20 candidates (`oversample * k`) per shard using the original, non-quantized vectors.
```

solutions/search/vector/knn/optimize-performance-accuracy.md
```diff
@@ -0,0 +1,367 @@
...
+
+This setting only applies to newly indexed vectors. To apply the option to all vectors in the index, the vectors must be re-indexed or force-merged after changing the setting.
+
+### Additional rescoring techniques [dense-vector-knn-search-rescoring-rescore-additional]
+
+The following sections provide additional ways of rescoring:
+
+#### Use the `rescore` section for top-level kNN search [dense-vector-knn-search-rescoring-rescore-section]
+
+You can use this option when you don’t want to rescore on each shard, but on the top results from all shards.
```
lcawl: Maybe clearer like this?:
```suggestion
You can use this option when you want to rescore on the top results from all shards rather than on each shard.
```

solutions/search/vector/knn/optimize-performance-accuracy.md
```diff
@@ -0,0 +1,367 @@
...
+          "queryVector": [0.04283529, 0.85670587, -0.51402352, 0]
+        }
+      }
+    }
+  }
+}
+```
+
+1. The number of results to return
+2. The `knn` query to perform the initial search, this is executed per-shard
```
lcawl: One of the words we try to avoid per https://www.elastic.co/docs/contribute-docs/style-guide/word-choice
```suggestion
2. The `knn` query to perform the initial search, this is run per-shard
```

lcawl (approved):
Added suggestions but they're all minor so take or leave as you please

solutions/search/vector/knn/approximate-knn.md
mayya-sharipova: Resolved!

michaeljmarshall:
I pushed a commit to add documentation for the `dense_vector` query type, a function to compute exact kNN. I reviewed my changes locally using `docs-builder serve`. Happy to make changes if needed.

solutions/search/vector/knn/optimize-performance-accuracy.md
pmpailis: ++ this should read as a tradeoff

solutions/search/vector/knn/optimize-performance-accuracy.md
```diff
@@ -0,0 +1,397 @@
...
+
+# Optimize performance and accuracy [optimize-knn-performance-accuracy]
+
+Approximate kNN search balances query speed, result quality, and resource use, and small configuration changes can have a large impact on latency, recall, and storage costs in production.
+
+This page covers trade-offs among search speed, recall, indexing cost, vector storage, quantization, and rescoring for approximate kNN search.
+
+## Tune approximate kNN for speed or accuracy [tune-approximate-knn-for-speed-accuracy]
+
+To gather results, the kNN API first finds a `num_candidates` number of approximate neighbors per shard, computes similarity to the query vector, selects the top `k` per shard, and merges them into the global top `k` nearest neighbors.
```
pmpailis:
```suggestion
To gather results, the kNN API first finds a `num_candidates` number of approximate neighbors per segment, computes similarity to the query vector, selects the top `k` per shard, and merges them into the global top `k` nearest neighbors.
```

solutions/search/vector/knn/optimize-performance-accuracy.md
```diff
@@ -0,0 +1,397 @@
...
+
+For HNSW indices, `num_candidates` is the main search-time speed/accuracy control:
+
+* Increase `num_candidates` to improve recall and accuracy (at the cost of higher latency).
+* Decrease `num_candidates` for faster queries (with a potential accuracy trade-off).
```
pmpailis: Even though it's implied, maybe we could clearly note that `num_candidates` affects the exploration that we do. So the exploration stops once the available documents to explore cannot enter the `num_cands`-sized heap of top results, i.e. when the best available document to explore, is worst than the worst of what we have already collected.

wajihaparvez (approved):
Admin change lgtm

solutions/search/vector/knn/nested-knn-search.md
```diff
@@ -0,0 +1,440 @@
...
+                    "gte": "2019-05-01",
+                    "lte": "2019-05-05"
+                }
+            }
+        }
+    }
+}
+```
+
+With the top-level `creation_time` filter applied, only document `1` falls within the specified range, so the response contains a single hit.
```
john-wagster, line 188: linter rule here is wrong this is in reference to a hit

kosabogi:
Thank you for your suggestions and work on this, @mayya-sharipova, @michaeljmarshall, and @john-wagster, and for your review, @lcawl! For the next few days I'll be focusing more on the Vector DB-related docs changes, but once I'm done with those I'll come back to this and review everything together alongside Lisa's suggestions. After that, I think the PR should be good to go. Thanks a lot again!

pmpailis:
@kosabogi fyi i also have some minor updates on draft for the `optimize-performance-accuracy.md` will re-assess  and push them later today!

kosabogi:
> @kosabogi fyi i also have some minor updates on draft for the `optimize-performance-accuracy.md` will re-assess and push them later today!

Great, thanks for the heads-up!

solutions/search/vector/knn.md
kosabogi: Good callout! We've recently added the [Field types and queries](https://www.elastic.co/docs/solutions/search/vector#vector-queries-and-field-types) page to the Vector search docs, which could be a good candidate for navigating users to other query types. That said, I'm not sure if we should cover those here. @mayya-sharipova, could you please weigh in on this?

solutions/search/vector/knn/optimize-performance-accuracy.md
kosabogi: Yes, I agree this should be a numbered step too. Resolved it.

kosabogi:
Hey @wajihaparvez, some additional content was added on the admin side. Could you please take another look at the relevant commits? [Here](https://github.com/elastic/docs-content/pull/7138/changes/d9b6cfe29bba5fd1f9e8f22ac34f1d54360bb097..3a6cb468e04b810a85f857028d5b284bd02c03b6#diff-f95a9bde5dec343f74a420b8a4a62368e5348f33a4d030ceca539e0f134b3ee8) you can find those commits. Thank you!

deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
```diff
@@ -62,39 +62,125 @@ Another option is to use  [synthetic `_source`](elasticsearch://reference/elasti
...
 
-Here are estimates for different element types and quantization levels:
+The next sections list the Lucene files that make up a `dense_vector` field, then estimate the off-heap RAM for the files that must stay in cache.
 
-| `element_type` | `quantization` | Required RAM |
+### Vector files [vector-files-off-heap-ram]
+
+Each structure is a Lucene file (also reported under `off_heap.*_size_bytes` in [index stats]({{es-apis}}operation/operation-indices-stats)). Metadata files (`.vem`, `.vemf`, `.vemq`, `.vemb`) are small and you do not need to preload them.
+
+The **Off-heap RAM** column is whether that file is part of the working set you size in [Estimate off-heap RAM](#_estimate_off_heap_ram):
```
wajihaparvez: How about "column indicates" rather than "column is"?

deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
```diff
@@ -62,39 +62,125 @@ Another option is to use  [synthetic `_source`](elasticsearch://reference/elasti
...
+
+**What should stay in RAM, by index type:**
+
+- `flat`: the raw vectors.
+- `hnsw`: the raw vectors and the graph.
+- `int8_flat` / `int4_flat` / `bbq_flat`: the quantized codes only.
+- `int8_hnsw` / `int4_hnsw` / `bbq_hnsw`: the quantized codes and the graph.
+- `bbq_disk` (DiskBBQ): all centroids. Extra headroom to cache about 5–10% of the posting lists is a good target; the rest of the postings and the raw vectors stay on disk. This is why DiskBBQ can serve far more vectors per GiB of RAM.
+
+Those structures are the files marked **Yes** (and **Partial** for DiskBBQ cluster headroom) in [Vector files](#vector-files-off-heap-ram).
```
wajihaparvez: Not sure what "those structures" is referring to. Could use some clarification

deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
```diff
@@ -105,30 +191,103 @@ estimated\ bytes &= (1,000,000 \times 4 \times 16) + (1,000,000 \times 4 \times
 \end{align*}
```
wajihaparvez, line 200: Image

Not sure if this is just a docs preview issue, but this string is getting cut off with no way to scroll ^

deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
```diff
@@ -105,30 +191,103 @@ estimated\ bytes &= (1,000,000 \times 4 \times 16) + (1,000,000 \times 4 \times
...
+```{applies_to}
+stack: ga 9.3+
+```
+
+DiskBBQ keeps all centroids resident. Extra headroom to cache about 5–10% of the posting lists (cluster vectors) is a good target. Use the centroid and cluster formulas in [Index structure on disk](#_index_structure_on_disk).
 
 ```{math}
 \begin{align*}
-num\_clusters=\frac{num\_vectors}{vectors\_per\_cluster}=\frac{num\_vectors}{384}
+required\ RAM &= centroid\ bytes \\
```
wajihaparvez, line 227: Image

Same formatting issue as before. Cut off in standard window size with no way to scroll

deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
```diff
@@ -105,30 +191,103 @@ estimated\ bytes &= (1,000,000 \times 4 \times 16) + (1,000,000 \times 4 \times
...
 \end{align*}
 ```
 
+Start with all centroids plus about 5% of the posting lists in RAM and tune based on benchmark results. The useful fraction depends on your query patterns: queries that access overlapping clusters benefit from caching more.
+
+:::
+
+::::
+
+Data nodes should also leave a buffer for other ways that RAM is needed. For example your index might also include text fields and numerics, which also benefit from using filesystem cache. Run benchmarks with your dataset to confirm there is enough memory for good search performance. Nightly examples include the [`so_vector`](https://elasticsearch-benchmarks.elastic.co/#tracks/so_vector) and [`dense_vector`](https://elasticsearch-benchmarks.elastic.co/#tracks/dense_vector) tracks.
```
wajihaparvez: For example your index might ~also~ include text fields and numerics, which also benefit from using filesystem cache.

deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
```diff
@@ -105,30 +191,103 @@ estimated\ bytes &= (1,000,000 \times 4 \times 16) + (1,000,000 \times 4 \times
...
+| `quantization` | Bytes per vector |
+| --- | --- |
+| `int8` | `num_dimensions + 16` |
+| `int4` | `⌈num_dimensions / 2⌉ + 16` |
+| `bbq` | `⌈num_dimensions / 64⌉ × 8 + 14` |
+
 ```{math}
 \begin{align*}
-estimated\ quantized\ vector\ bytes = num\_vectors \times ((num\_dimensions/8 + 14 + 2) \times 2)
+estimated\ bytes = num\_vectors \times bytes\ per\ vector
```
wajihaparvez, line 273: Image

Formatting issue again, text overlapping with right-side nav. Could just be a preview issue but calling out to be safe

deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
```diff
@@ -105,30 +191,103 @@ estimated\ bytes &= (1,000,000 \times 4 \times 16) + (1,000,000 \times 4 \times
...
+| `bbq_disk` | Clusters (`.clivf`), `num_dimensions` less than 384 | `num_vectors × 2 × (⌈num_dimensions / 2⌉ + 16)` |
+
+For DiskBBQ, add centroid bytes and cluster bytes. The `× 2` on cluster vectors is a conservative upper bound: a vector may be written to a second cluster. Real cluster storage is between 1× and 2×.
+
+#### Total disk per replica
+
+```{math}
+\begin{align*}
+total\ disk = raw\ vector\ bytes + quantized\ disk + index\ structure\ bytes
+\end{align*}
```
wajihaparvez, line 296: Check formatting

wajihaparvez (approved):
@kosabogi New changes on deploy-manage LGTM. Suggested small style edits + possible formatting errors. Thank you!

florent-leborgne (approved):
LGTM for pages owned by ski-docs (link changes only)

mayya-sharipova (approved):
@kosabogi  
The Search Relevance Team contributed major sections of this PR, so it is good to go from our side.

Can you please drive it to the end and make it merged, as the PR is big, and it is difficult to track all the changes.

deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
kosabogi: interesting! in the current preview it looks okay to me: 

Screenshot 2026-09-07 at 8 59 01

deploy-manage/production-guidance/optimize-performance/approximate-knn-search.md
kosabogi: Thanks, @wajihaparvez! I think this was a temporary preview issue, the formulas look good to me now.

solutions/search/vector/knn.md
```diff
@@ -1,1457 +1,79 @@
...
-* **Recommendations**
-  * Product recommendations
-  * Collaborative filtering
-  * Personalized content discovery
+## Common use cases for kNN vector similarity search
 
-* **Analysis**
-  * Anomaly detection
-  * Pattern matching
+- **Search**
```
leemthompo, line 20: nit: We should always have a lead-in sentence before a list, table, code block etc. Otherwise jumping from heading to a list is a bit jarring :)

solutions/search/vector/knn.md
```diff
@@ -1,1457 +1,79 @@
...
 
 ## Prerequisites for kNN search [knn-prereqs]
 
 To run a kNN search in {{es}}:
 
-* Your data must be vectorized. You can [use an NLP model in {{es}}](../../../explore-analyze/machine-learning/nlp/ml-nlp-text-emb-vector-search-example.md) or generate vectors outside {{es}}.
-  * Use the [`dense_vector`](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md) field type for dense vectors.
-  * Query vectors must have the same dimension and be created with the same model as the document vectors.
-  * Already have vectors? Refer to [Bring your own dense vectors](bring-own-vectors.md).
+- Your data must be vectorized. You can [use an NLP model in {{es}}](../../../explore-analyze/machine-learning/nlp/ml-nlp-text-emb-vector-search-example.md) or generate vectors outside {{es}}.
```
leemthompo, line 37:
```suggestion
- Your data must be vectorized. You can use [`semantic_text`](/solutions/search/semantic-search/semantic-search-semantic-text.md) to have Elastic handle embedding automatically, use the [Elastic Inference Service](/explore-analyze/elastic-inference/elastic-inference-service.md) for managed inference, [deploy an NLP model](/explore-analyze/machine-learning/nlp/ml-nlp-text-emb-vector-search-example.md) on an ML node, or generate vectors outside of your Elastic deployment.
```

This reads a bit outdated, I think we need to add the various options here

solutions/search/vector/knn.md
```diff
@@ -1,1457 +1,79 @@
...
+  - Use the [`dense_vector`](elasticsearch://reference/elasticsearch/mapping-reference/dense-vector.md) field type for dense vectors.
+  - Query vectors must have the same dimension and be created with the same model as the document vectors.
+  - Already have vectors? Refer to [Bring your own dense vectors](bring-own-vectors.md).
```
leemthompo, line 40: this isn't a prerequisite, it's more like an aside, perhaps could be a TIP?

solutions/search/vector/knn.md
```diff
@@ -1,1457 +1,79 @@
...
-  * `read` to search the index
+- Required [index privileges](elasticsearch://reference/elasticsearch/security-privileges.md#privileges-list-indices):
+  - `create_index` or `manage` to create an index with a `dense_vector` field
+  - `create`, `index`, or `write` to add data
+  - `read` to search the index
 
 :::{tip}
 The default type of {{es-serverless}} project is suitable for this use case unless you plan to use uncompressed dense vectors (`int4` or `int8` quantization strategies) with high dimensionality.
-Refer to [](dense-vector.md#vector-profiles).
+To learn more, refer to [General purpose and vector optimized projects](dense-vector.md#vector-profiles).
```
leemthompo: Do we need to rethink this tip with an eye on VDB project type?

solutions/search/vector/knn.md
```diff
@@ -1,1457 +1,79 @@
...
+  - Best for most production workloads where low latency and scale matter more than perfect recall.
+  - Uses graph-based or clustered index structures to find similar vectors quickly without scoring every document in the index.
+  - Refer to [Approximate kNN search](knn/approximate-knn.md) for mapping, indexing, and search examples.
+
+- **Exact, brute-force kNN**
+  - Best for small datasets, pre-filtered subsets, or when you need precise scoring without approximate indexing.
+  - Scores every matching document to find the nearest neighbors across the complete set of matches, rather than selecting from an approximate candidate set.
+  - Supports two query methods:
+    - The [`dense_vector` query](elasticsearch://reference/query-languages/query-dsl/query-dsl-dense-vector-query.md) for standard exact vector scoring.
+    - The [`script_score` query](elasticsearch://reference/query-languages/query-dsl/query-dsl-script-score-query.md) when you need a custom scoring calculation.
```
leemthompo, line 63:
```suggestion
```

I think the query method detail (dense_vector vs script_score) belongs in `exact-knn.md`, not here.

solutions/search/vector/knn.md
```diff
@@ -1,1457 +1,79 @@
...
-* Relevance ranking based on natural language processing (NLP) algorithms
-* Product recommendations and recommendation engines
-* Similarity search for images or videos
-
-::::{tip}
-Check out our [hands-on tutorial](bring-own-vectors.md) to learn how to ingest dense vector embeddings into Elasticsearch.
-::::
+- **Approximate kNN**
+  - Best for most production workloads where low latency and scale matter more than perfect recall.
+  - Uses graph-based or clustered index structures to find similar vectors quickly without scoring every document in the index.
```
leemthompo, line 55: arguably, "graph-based or clustered index structures" is implementation detail that doesn't help a user decide which to use

solutions/search/vector/knn.md
```diff
@@ -1,1457 +1,79 @@
...
-Common use cases for kNN include:
-
-* Relevance ranking based on natural language processing (NLP) algorithms
-* Product recommendations and recommendation engines
-* Similarity search for images or videos
-
-::::{tip}
-Check out our [hands-on tutorial](bring-own-vectors.md) to learn how to ingest dense vector embeddings into Elasticsearch.
-::::
+- **Approximate kNN**
```
leemthompo, line 53: I'd consider not using a bulleted list here, it reads a little disjointed to me, particuarly the "learn more" bullet. Because this is two options with a decision rationale, I think it reads more naturally as prose.

solutions/search/vector/knn.md
```diff
@@ -1,1457 +1,79 @@
...
+  - Best for small datasets, pre-filtered subsets, or when you need precise scoring without approximate indexing.
+  - Scores every matching document to find the nearest neighbors across the complete set of matches, rather than selecting from an approximate candidate set.
+  - Supports two query methods:
+    - The [`dense_vector` query](elasticsearch://reference/query-languages/query-dsl/query-dsl-dense-vector-query.md) for standard exact vector scoring.
+    - The [`script_score` query](elasticsearch://reference/query-languages/query-dsl/query-dsl-script-score-query.md) when you need a custom scoring calculation.
+  - Refer to [Exact kNN search](knn/exact-knn.md) for search examples.
+
+Approximate kNN offers low latency and good accuracy, while exact kNN guarantees accurate results but does not scale well for large datasets. With exact kNN, every matching document must be scanned to compute the vector function, which can result in slow search speeds. However, you can improve latency by filtering your data to a small subset of documents.
+
+## Resources
```
leemthompo, line 68: This list is overwhelming and effectively duplicates the TOC. The natural next step from an overview page is the child pages. I'd drop the exhaustive list entirely, rename to "Next steps" and just keep just the two top-level links (approximate and exact) as a clear "where to go next" without the full list of grandchildren. 👪

solutions/search/vector/knn/build-search-queries.md
leemthompo, line 1: I wonder if we should rethink the name of this file?

We already have https://www.elastic.co/docs/solutions/search/querying-for-search which has a H1 called "Build search queries" so the SEO implications are probably not great.

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
+---
+navigation_title: Examples of using approximate kNN
+description: Examples of using approximate kNN in Elasticsearch search queries for query vector builders, filtering, similarity thresholds, hybrid search, multiple vector fields, and aggregations.
+applies_to:
+  stack:
+  serverless:
+---
+
+# Examples of using approximate kNN in search queries [build-approximate-knn-queries]
```
leemthompo, line 9: Related to the filename question, "Build the query vector" isn't an example, it's a concept/task. And the page covers filtering, hybrid search, aggregations, similarity thresholds which are distinct _features_ or concepts, not just examples. We might need to workshop the title more...

I fear this page isn't super coherent. We can determine if it's OK to make some minor changes and revisit in future PR, or whether it's work overhauling in this PR.

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+
+This page collects the query patterns you need most often with approximate kNN: building the query vector, restricting results with filters, requiring a minimum similarity, combining kNN with keyword search, searching several vector fields at once, and aggregating over kNN results.
+
+Except where a text embedding model is required, the examples run against the `image-index` mapping and sample data from [Approximate kNN search](approximate-knn.md#approximate-knn-example). That index stores an `image-vector` field with `l2_norm` similarity, a `title` text field, and a `file-type` keyword field.
+
+The examples use whichever of the three kNN forms suits the pattern: the top-level `knn` option, the [`knn` query](elasticsearch://reference/query-languages/query-dsl/query-dsl-knn-query.md), or the [`knn` retriever](elasticsearch://reference/elasticsearch/rest-apis/retrievers/knn-retriever.md). Refer to [Approximate kNN search methods](approximate-knn.md#approximate-knn-methods) to choose between them for your own searches. For vectors stored in `nested` fields, refer to [Nested kNN search](nested-knn-search.md).
+
+## Build the query vector [knn-build-query-vector]
```
leemthompo, line 17: This section duplicates content already covered in the reference at https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-knn-query#query-vector-builders-overview — this section should explain when to use it and link out, not re-document it.

This contributes to this page being a little disjointed and lacking consistency.

Also, "Build the query vector" is a concept that applies to all kNN, not just approximate, so we're at the wrong level of abstraction

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+
+Three builders are available, and all of them work with the `knn` option, the `knn` query, and the `knn` retriever:
+
+- `text_embedding` generates a vector from query text using a text embedding model deployed in {{es}}. Refer to [Perform semantic search](#knn-semantic-search).
+- `lookup` reuses a vector already stored in a document. Refer to [Find documents similar to an existing document](#knn-lookup-similar-documents). {applies_to}`stack: ga 9.4`
+- `embedding` generates a vector from text or a base64-encoded image using an {{infer}} endpoint, which makes multimodal search possible. Refer to [Multimodal search](../../multimodal-search.md) for the use case and [`embedding`](elasticsearch://reference/query-languages/query-dsl/query-dsl-knn-query.md#knn-query-builder-embedding) for the parameters. {applies_to}`stack: preview 9.4` {applies_to}`serverless: preview`
+
+### Perform semantic search [knn-semantic-search]
```
leemthompo, line 27: I see a few issues with this section:

- This is an example of text_embedding as a query_vector_builder type, not a guide to semantic search. The heading operates at the wrong level of abstraction.
- "Previously deployed text embedding model" assumes manual ML node deployment; doesn't account for managed inference endpoints via the Elastic Inference Service etc. In fact we shouldn't need to re-explain this part here, it should be stated in the pre-requisites. (The intro paragraph and the bullet list  restate concepts and requirements that belong upstream, not repeated in every example.)
- The tip says semantic search is better handled with semantic_text, which undermines the entire section's framing. If the recommended path is elsewhere, this section should just show the text_embedding query and link out, not explain semantic search from scratch.

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+  }
+}
+```
+
+1. The ID of the text embedding model that generates the query vector. Use the same model that produced the document embeddings in the target index. You can also provide a `deployment_id` as the `model_id` value.
+2. The query string from which the model generates the dense vector representation.
+
+For a walkthrough that covers deploying a model, generating document embeddings, and querying them, refer to this [end-to-end example](../../../../explore-analyze/machine-learning/nlp/ml-nlp-text-emb-vector-search-example.md).
+
+### Find documents similar to an existing document [knn-lookup-similar-documents]
```
leemthompo, line 67: "Find documents similar to an existing document" describes a use case, the actual subject is the `lookup` query vector builder.

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+```{applies_to}
+stack: ga 9.4
+```
+
+Use the `lookup` query vector builder when the vector you want to search with is already stored in a document. This is the pattern behind "more like this" and recommendation features: instead of embedding new input, you take the vector from an item the user is viewing and find its nearest neighbors.
```
leemthompo, line 72:
```suggestion
Use the [`lookup` query vector builder](elasticsearch://reference/query-languages/query-dsl/query-dsl-knn-query#knn-query-builder-lookup) when the vector you want to search with is already stored in a document. This is the pattern behind "more like this" and recommendation features: instead of embedding new input, you take the vector from an item the user is viewing and find its nearest neighbors.
```

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+            "values": ["2"]
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+## Filter kNN results [knn-search-filter-example]
```
leemthompo, line 127: This is the strongest content on the page and arguably deserves its own page

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+
+* If the number of documents matching the filter in a segment is no greater than the number of candidates the search would examine there, the search skips the HNSW graph and scores the filtered documents directly.
+* While exploring the graph, if the search visits more nodes than there are documents matching the filter, it stops traversing and scores the filtered documents directly.
+
+{applies_to}`stack: preview 9.1` For indices created in 9.1 or later, {{es}} also applies the `acorn` filter heuristic by default, which traverses only vectors that match the filter instead of comparing every vector it visits. This is generally faster at comparable recall, although you might need to raise `num_candidates` to hit exceptionally high recall targets. To change the heuristic, use [`index.dense_vector.hnsw_filter_heuristic`](elasticsearch://reference/elasticsearch/index-settings/index-modules.md#index-dense-vector-hnsw-filter-heuristic).
+
+For more ways to trade search speed against accuracy, refer to [Optimize performance and accuracy](optimize-performance-accuracy.md).
+
+## Set a similarity threshold [knn-similarity-search]
```
leemthompo, line 223: this section lacks visual hierarchy when you look at the rendered page

perhaps a couple of H3 subheadings would be a quick fix: something like "Understand threshold values by similarity metric" and "Derive a threshold from an observed score"

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+    }
+  },
+  "fields": ["title"],
+  "_source": false
+}
+```
+
+In this data set, the only document with `file-type = png` has the vector `[42, 8, -15]`. Its `l2_norm` distance from `[1, 5, -20]` is `41.412`, which is farther than the threshold of `36` allows. The filter leaves one candidate and the threshold rejects it, so this search returns no hits.
+
+## Combine approximate kNN with keyword search [combine_approximate_knn_with_other_features]
```
leemthompo, line 279: I think this heading is too narrow because "Combine approximate kNN with keyword search" excludes the linear retriever example which could combine kNN with any signal, not just keyword. "Hybrid search" might be the more appropriate term.

Maybe "`## Use kNN in hybrid search`" ?

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+
+Use hybrid retrieval when you want one ranked result list that reflects both how similar documents are to your query vector and how well they match specific words or phrases. For example, you might find images that look similar to a reference photo while also matching a title keyword like "mountain lake".
```
leemthompo, line 281: `hybrid-search.md` should link to these examples

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+
+This search retrieves the global top `k = 5` neighbors for `image-vector` and the global top `k = 10` for `title-vector`. These vector result sets are combined with the matches from the `match` query, and the top 10 overall documents are returned. Multiple `knn` clauses and the `query` clause are combined via a disjunction (boolean *OR*). The top `k` vector results represent the global nearest neighbors across all index shards.
+
+With the boosts configured above, a document is scored as:
+
+```
+score = 0.9 * match_score + 0.1 * knn_score_image-vector + 0.5 * knn_score_title-vector
+```
+
+As with hybrid retrieval, you can instead pass one `knn` retriever per field to an `rrf` or `linear` retriever, which spares you from balancing raw scores across fields. Refer to [Combine approximate kNN with keyword search](#combine_approximate_knn_with_other_features).
```
leemthompo, line 463:
```suggestion
As with hybrid retrieval, you can instead pass one `knn` retriever per field to an `rrf` or `linear` retriever, which spares you from balancing raw scores across fields. Refer to [](#combine_approximate_knn_with_other_features).
```

this will keep the link text future proof :)

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+
+## Aggregate over kNN results [knn-aggregations]
+
+You can use [aggregations](../../../../explore-analyze/query-filter/aggregations.md) with the `knn` option, but the buckets cover a different document set than you might expect. {{es}} computes aggregations over the documents that match the search, and for approximate kNN that means the top `k` nearest documents rather than everything in the index. If the request also includes a `query`, aggregations cover the combined set of `knn` and `query` matches.
```
leemthompo, line 467:
```suggestion
You can use [aggregations](/explore-analyze/query-filter/aggregations.md) with the `knn` option, but the buckets cover a different document set than you might expect. {{es}} computes aggregations over the documents that match the search, and for approximate kNN that means the top `k` nearest documents rather than everything in the index. If the request also includes a `query`, aggregations cover the combined set of `knn` and `query` matches.
```

nit: absolute links are easier to reason about in the source, and more robust to IA changes :)

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+        "field": "file-type"
+      }
+    }
+  }
+}
+```
+
+The `file-types` buckets describe the five nearest neighbors only, so treat the counts as a summary of the result set rather than of the index. To aggregate across all documents that match a filter, run a separate request without a `knn` clause.
+
+## Resources
```
leemthompo, line 492: again this is kinda overwhelming list, think about whether each link here adds navigation value that the TOC doesn't already provide

leemthompo (commented):
A few ideas here, I haven't provided a full review, feel free to be ruthless about what you tackle in this PR versus follow-ups :)

(I'll try to to put up a review against other pages tomorrow)

leemthompo:
💡 I don't think we're linking from this section to the [KNN search profiling page](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/search-profile#profiling-knn-search), which we definitely should do

kosabogi:
> @kosabogi The Search Relevance Team contributed major sections of this PR, so it is good to go from our side.
> 
> Can you please drive it to the end and make it merged, as the PR is big, and it is difficult to track all the changes.


Thanks for confirming! We're currently going through the content to make sure it aligns with our docs style guide, making small style corrections where needed and checking for consistency. We'll get it merged soon - thanks for your patience!

solutions/search/vector/knn/build-search-queries.md
kosabogi: I think it would be better to handle this page in a focused follow-up. The page has grown substantially as new content was added, and I agree that its structure no longer fully reflects its scope. For now, I renamed the file to avoid competing with the existing “Build search queries” page in search results. 

At this point, I’d prefer to get this PR merged soon so the updated content, especially the new additions, is available, and then revisit the page structure, title, and potential split in a follow-up. Does that sound reasonable?

solutions/search/vector/knn/build-search-queries.md
kosabogi: I agree that this content overlaps with the reference documentation, but I think resolving it properly requires broader consolidation rather than a local edit to this page. The original scope of this PR was only to break up the overly long kNN page and make the content easier to navigate. 

I’d keep this section as is for now so we can merge the updated and newly added content. I’d then address the cross-repository consolidation and the broader kNN content structure through a separate issue and focused follow-up.

solutions/search/vector/knn/build-search-queries.md
kosabogi: Good points! I reworked this section so it focuses specifically on the text_embedding query vector builder. I removed the repeated semantic search explanation and prerequisites, as well as the semantic_text tip. I also aligned the text_embedding and lookup headings so they’re at the same level of abstraction.

solutions/search/vector/knn/build-search-queries.md
Mikep86: ++. Should we document query vector builders in a dedicated page and link to that?

solutions/search/vector/knn/build-search-queries.md
Mikep86: ++

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+  "fields": ["title"],
+  "_source": false
+}
+```
+
+::::{note}
+The filter is applied **during** approximate kNN search to ensure that `k` matching documents are returned. In contrast, post-filtering applies the filter **after** the approximate kNN step and can return fewer than `k` results, even when enough relevant documents exist.
+::::
+
+### Put the filter where the kNN search can see it [knn-prefilter-vs-postfilter]
```
Mikep86, line 156: This heading is unclear. Can we rename it to make it clearer that it is about pre-filtering vs. post-filtering?

solutions/search/vector/knn/build-search-queries.md
```diff
@@ -0,0 +1,500 @@
...
+
+To limit the impact, Lucene falls back to brute force per segment in two cases:
+
+* If the number of documents matching the filter in a segment is no greater than the number of candidates the search would examine there, the search skips the HNSW graph and scores the filtered documents directly.
+* While exploring the graph, if the search visits more nodes than there are documents matching the filter, it stops traversing and scores the filtered documents directly.
+
+{applies_to}`stack: preview 9.1` For indices created in 9.1 or later, {{es}} also applies the `acorn` filter heuristic by default, which traverses only vectors that match the filter instead of comparing every vector it visits. This is generally faster at comparable recall, although you might need to raise `num_candidates` to hit exceptionally high recall targets. To change the heuristic, use [`index.dense_vector.hnsw_filter_heuristic`](elasticsearch://reference/elasticsearch/index-settings/index-modules.md#index-dense-vector-hnsw-filter-heuristic).
```
Mikep86, line 219: This is really still in tech preview, and applied by default?

solutions/search/vector/knn/build-search-queries.md
Mikep86: ++

solutions/search/vector/knn/build-search-queries.md
Mikep86, line 1: I agree with @leemthompo that this page is a bit disjoint as currently named and organized. "Build the query vector" applies to exact and approximate kNN search, and should be documented in a shared place. We should ensure this page directly documents information that applies only to approximate kNN.

Mikep86 (commented):
Heads up, I only reviewed `build-search-queries`

pmpailis:
Added a [commit](https://github.com/elastic/docs-content/pull/7138/commits/2c9e3e7f64a454eb303c3c5e62adc627f34aa555) to remove the vector sizing calculator for now as it's under review. Will revisit once we finalize its structure.

solutions/search/vector/knn/build-search-queries.md
kosabogi: I checked this in the code repository, and the default behavior described here matches the implementation. @mayya-sharipova @pmpailis, could you confirm whether the technical preview tagging is correct?

pmpailis:
@kosabogi just a heads up that we revisited the suggested RAM for `disk_bbq` and i'll push another update on this part.

kosabogi:
Thanks everyone for the reviews and technical input! I’ve addressed the feedback by:

- Moving shared query-vector guidance to the main kNN page.
- Renaming the query examples page to Approximate kNN query examples.
- Moving filtered approximate kNN guidance to its own page.
- Clarifying that the current nested examples cover approximate kNN.
- Improving headings, wording, links, navigation, anchors, and redirects.
- Adding a link to kNN search profiling.

I kept the existing SME-authored technical content mostly intact and limited the latest changes to moving content and small scope/navigation updates.

For a follow-up PR, I suggest:

- Reviewing how exact kNN applies to nested and filtered searches.
- Adding tested exact examples with SME input.
- Reconsidering the broader organization of the query examples.

cc @leemthompo @Mikep86 

@pmpailis  would you prefer me to wait until you push the DiskBBQ RAM update, or would it be okay to handle that in a follow-up PR?

pmpailis:
> would you prefer me to wait until you push the DiskBBQ RAM update, or would it be okay to handle that in a follow-up PR?

Thanks @kosabogi ! If it's ok I d suggest to wait to merge both at once. Was handling a couple of other tasks and hadn't find time for this until today. Will push an update by Monday :)

leemthompo (approved):
I think this is good to go now and represents a significant improvement over the giant pages that accumulated over time. Let's just be sure to doublecheck and triplecheck the redirects work as expected before merging and to verify manually also post-publication to be 100% sure. 

If there are any non-urgent outstanding comments or suggestions they can be handled in follow-ups.

Nice job on this hairy piece of refactoring 👍

PR #7154, opened 2026-06-30
Route ingest and experience issues to Team:SKI board

bmorelli25:
## Summary

Consolidates the Ingest and Experience triage routing into a single `Team:SKI` team label.

- Replaces `Team:Ingest` and `Team:Experience` routing with one `Team:SKI` entry that lands issues on [project board 1034](https://github.com/orgs/elastic/projects/1034).
- Recases the remaining team comparisons from lowercase `team:` to `Team:` so they match the actual GitHub label names. The comparison is case-sensitive (`==`), so the prior lowercase values would not have matched the real `Team:*` labels.

The `Team:SKI` label has already been created in this repo. The old `Team:Ingest` / `Team:Experience` labels are left in place (no longer routed).

## Test plan

- [ ] Label a test issue `Team:SKI` and confirm it lands on board 1034 and `needs-team` is removed.
- [ ] Spot-check another team label (e.g. `Team:Search`) still routes correctly after the recasing.


KOTungseth (approved):
I love a good consolidation

PR #7186, opened 2026-07-02
Fix index compatibility table header

### LINKED ISSUE #6819: [Website]: Incorrect snapshot version compatibility table
### Before you submit

### Type of issue

None

https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore#snapshot-restore-version-compatibility

### What happened?

I attempted to take a snapshot on ELK version 8.1.0 and restore it on version 9.2.4 which failed with the log `FATAL Error: Kibana 8.1.0 deployment detected. Please upgrade to Kibana 8.18.0 or newer before upgrading to 9.x series.`

I understand that this is expected and has been addressed [here](https://github.com/elastic/kibana/issues/217145), however the documentation appears to be wrong for this.

First of all, the compatibility table is somewhat confusing to read - both myself and a colleague initially read this transposed to how it is meant to be.
But secondly, there is no mention of the upgrade path via 8.18.0 - it appears the documentation has not been updated since issue 217145 was addressed.

### Additional info

_No response_

## Action
_added by @georgewallace_
- Clarify the table headers as it is confusing  `Index creation version` looks to be a header of the first column
- Look into the lack of clarifying 8.18 upgrade path

yetanothertw:
## Summary

Fixes #6819 which reports:

1. the docs don't mention that Kibana versions complicate snapshot restores:
> I attempted to take a snapshot on ELK version 8.1.0 and restore it on version 9.2.4 which failed with the log FATAL Error: Kibana 8.1.0 deployment detected. Please upgrade to Kibana 8.18.0 or newer before upgrading to 9.x series.

2. confusion around the way the table headings are with a single header row labeled only the left column (**Index creation version**).

## Fix for the 1st problem 

The PRs linked on the issue added a guard (that's present in the [current Kibana main](https://github.com/elastic/kibana/blob/main/src/core/packages/saved-objects/migration-server-internal/src/run_v2_migration.test.ts#L113-L122) and [here](https://github.com/elastic/kibana/blob/main/src/core/packages/saved-objects/server-internal/src/saved_objects_service.ts#L531C89-L532C1)) the minimum version seems to be [hardcoded here](https://github.com/elastic/kibana/blob/main/src/core/packages/saved-objects/server-internal/src/saved_objects_service.ts#L531C89-L532C1).

After restoring a snapshot from an 8.1.0 cluster, .kibana still has aliases like .kibana_8.1.0. Kibana 9.2.4 sees that, compares it to 8.18.0, and fails fatally before migrations run.

[PREVIEW AVAILABLE HERE](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7186/deploy-manage/tools/snapshot-and-restore#kib-compatibility)

### Why snapshot restore fails here

**My friend Claude reckons:**

Snapshot/restore can move Elasticsearch data, but it does not replace the required Kibana upgrade path:

1. Snapshot from 8.1.0 includes Kibana system indices (.kibana, .kibana_task_manager, etc.) and their version aliases.
2. Restore onto a 9.2.4 cluster brings those indices back as-is.
3. Kibana 9.2.4 starts, hits saved-object migration, sees .kibana_8.1.0, and refuses to proceed.

Elasticsearch version compatibility for data indices is separate from Kibana saved-object migration compatibility. A restore that ES accepts can still fail when Kibana starts.

### What the user should do instead

VALIDATED by Eng ([internal slack](https://elastic.slack.com/archives/C5UDK4KDJ/p1783078989114369)):

Supported path:

1. Restore the snapshot to an 8.18+ or 8.19 cluster (ideally latest 8.19 patch per upgrade docs).
2. Start Kibana 8.19 and let saved-object migrations complete.
3. Snapshot from 8.19 and restore to 9.2.4. (9.2.4 is used as an example as that's what the user reported they had).

Alternatively:

If the user only needs data and not the Kibana state:

1. Restore without the kibana feature state / .kibana* indices and accept a fresh Kibana setup.
2. Use reindex-from-remote for data-only migration (as noted in snapshot compatibility docs).

## Fix for the 2nd problem

The table is correct but ambiguous: since both axes use version numbers, but only one axis was named, I'm adding another row to name both dimensions. This matches an existing docs pattern (in the CCS table). This should reduce transpose misreads without changing the underlying compatibility data.
Unfortunately, there's tooling limitation we need to work around as cell merging is not possible, which is why I chose this option. 

[PREVIEW AVAILABLE HERE](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7186/deploy-manage/tools/snapshot-and-restore#index-compatibility)

This is one option:

image

This is another option (currently on this branch):

image

Both are quite similar. Not sure which one is clearer.

## Generative AI disclosure

- [x] Yes  
- [ ] No


sabarasaba (approved):
I think it makes sense and it matches what we discussed over slack

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -158,22 +158,32 @@ You can’t restore a snapshot to an earlier version of {{es}}. For example, you
...
-| 8.0–8.19                                 | ✅                                        | ✅       | ✅      | ❌       | ❌      | ❌  |
-| {{version.stack.base}}–{{version.stack}} | ✅                                        | ❌       | ❌      | ❌       | ❌      | ❌  | 
+The following table shows whether an index can be restored to a given cluster version. Find your index’s creation version in the left column and your cluster version across the top. For example, an index created in 6.8 can be restored to a {{version.stack.base}}–{{version.stack}} cluster (✅) but not to a 7.0–7.1 cluster (❌).
+
+:::{table}
+:widths: 3-2-1-2-2-1-1
+
+|                       |     |        |       |          |         |     |
+|-----------------------|-----|--------|-------|----------|---------|-----|
+|                       | Cluster version (restore target) <br> {{version.stack.base}}–{{version.stack}} |  8.3–8.19 | 8.0–8.2 |  7.2–7.17 | 7.0–7.1  | 6.8  |
```
jloleysens: I think this change makes sense. I often call it "stack" version to imply Kibana, Elasticsearch and everything else, but if "cluster" is consistent language then let's stick with that.

jloleysens (approved):

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -182,6 +192,37 @@ To ensure index compatibility, you can first restore the index to another cluste
...
+FATAL  Error: Kibana 8.1.0 deployment detected. Please upgrade to Kibana 8.18.0 or newer before upgrading to 9.x series.
+```
+
+Restoring an old snapshot directly to a 9.x cluster is not a shortcut around the normal [upgrade paths](/deploy-manage/upgrade.md#upgrade-paths). It is recommended you upgrade to the latest compatible minor release before a major upgrade, even when you use snapshots to move data between clusters.
+
+#### Move {{kib}} configuration and saved objects across major versions
+
+To make {{kib}} state from an older 8.x snapshot compatible with a 9.x cluster, restore and migrate to an intermediate 8.x cluster first, then snapshot and restore again on the target version:
+
+1. [Restore the snapshot](/deploy-manage/tools/snapshot-and-restore/restore-snapshot.md) to a cluster running {{kib}} 8.18.0 or later. If your target is 9.1.0 or later, use the latest 8.19.x patch release, as described in [Prepare to upgrade](/deploy-manage/upgrade/prepare-to-upgrade.md).
```
eedugon: Considering we are almost in 9.5 and that users shouldn't be aiming to `9.0.x` I'd suggest to change the way that's presented:

```suggestion
1. [Restore the snapshot](/deploy-manage/tools/snapshot-and-restore/restore-snapshot.md) to a cluster running {{kib}} 8.19.x, or, if your target is 9.0.x, to a cluster running `8.18.x`, as described in [Prepare to upgrade](/deploy-manage/upgrade/prepare-to-upgrade.md).
```

And tbh I would even remove the `9.0.0` exception, I don't think it's worthy to highlight it on this doc. We decided some time ago to not give too much visibility to the 8.18.x to 9.0.0 exceptional upgrade path, just the minimum necessary, which is in the upgrade official docs on the upgrade paths section.

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -182,6 +192,37 @@ To ensure index compatibility, you can first restore the index to another cluste
...
+
+```
+FATAL  Error: Kibana 8.1.0 deployment detected. Please upgrade to Kibana 8.18.0 or newer before upgrading to 9.x series.
+```
+
+Restoring an old snapshot directly to a 9.x cluster is not a shortcut around the normal [upgrade paths](/deploy-manage/upgrade.md#upgrade-paths). It is recommended you upgrade to the latest compatible minor release before a major upgrade, even when you use snapshots to move data between clusters.
+
+#### Move {{kib}} configuration and saved objects across major versions
+
+To make {{kib}} state from an older 8.x snapshot compatible with a 9.x cluster, restore and migrate to an intermediate 8.x cluster first, then snapshot and restore again on the target version:
```
eedugon:
```suggestion
To make {{kib}} state from an older 8.x snapshot compatible with a 9.x cluster, restore and migrate to an intermediate 8.19.x cluster first, then snapshot there, and restore finally on the target version:
```

From 8.x, to 9.x, upgrade to intermediate 8.x feels odd... aren't we already in 8.x? :) I suggest to keep 8.19.x there (yes, even if 8.18.x could be potentially the target in rare cases that want to upgrade to 9.0.x).

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -182,6 +192,37 @@ To ensure index compatibility, you can first restore the index to another cluste
...
+
+1. [Restore the snapshot](/deploy-manage/tools/snapshot-and-restore/restore-snapshot.md) to a cluster running {{kib}} 8.18.0 or later. If your target is 9.1.0 or later, use the latest 8.19.x patch release, as described in [Prepare to upgrade](/deploy-manage/upgrade/prepare-to-upgrade.md).
+2. Start {{kib}} on the intermediate cluster and wait for startup to complete. {{kib}} runs saved object migrations at startup, rewriting the `.kibana` system indices from their restored version (for example, 8.1.0) to a format that 9.x can read. {{es}} does not run these migrations during a snapshot restore.
```
eedugon: That sentence is great.
```suggestion
2. Start {{kib}} on the intermediate cluster and wait for startup to complete. During the startup, {{kib}} detects data from a previous compatible version and runs saved object migrations, the same as during a {{kib}} upgrade. This process rewrites `.kibana` system indices from their restored version (for example, 8.1.0) to the `8.19.x` format that is fully compatible with 9.x. {{es}} does not run these migrations during a snapshot restore.
```

For your consideration, feel free to ignore or find a better way.

Your sentence is great already.

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -182,6 +192,37 @@ To ensure index compatibility, you can first restore the index to another cluste
...
+3. [Take a new snapshot](/deploy-manage/tools/snapshot-and-restore/create-snapshots.md) of the migrated cluster, then restore that snapshot to your 9.x cluster. {{kib}} should start cleanly on the target version.
```
eedugon: Cleanly, cleanly... Kibana should start and apply the same migrations as the ones expected during an upgrade from 8.19.x to 9.x. I wouldn't call that a clean start. It would be the same type of start of the previous step, expecting data migrations to occur, so I'd refer to it in a similar way than the previous start, otherwise the user might expect this `clean` start to be different.

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -182,6 +192,37 @@ To ensure index compatibility, you can first restore the index to another cluste
...
+
+For example, to move state from an 8.1.0 snapshot to 9.2.4, restore the 8.1.0 snapshot to an 8.19.x cluster, let {{kib}} complete its startup migrations, snapshot the 8.19.x cluster, and restore that snapshot to 9.2.4.
```
eedugon:
```suggestion
For example, to restore {{kib}} state from an 8.1.0 snapshot to a 9.2.4 cluster, restore the 8.1.0 snapshot to an 8.19.x cluster, let {{kib}} complete its startup migrations, snapshot the 8.19.x cluster, restore that snapshot to 9.2.4, and let {{kib}} perform the latest data migrations.
```

Without the change we are suggesting that 8.19.x data restored into 9.x is perfectly fine for Kibana without migrations, and that's not the case. Same migrations between 8.1.0 and 8.19.x would occur between 8.19.x and 9.x.x if I'm not mistaken.

We shouldn't cause that impression if we are offering such level of details between 8.1.0 and 8.19.x

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -182,6 +192,37 @@ To ensure index compatibility, you can first restore the index to another cluste
...
+
+Alternatively, if you only need to recover data and can accept a fresh {{kib}} setup, restore the snapshot without the `kibana` feature state.
+
+:::{admonition} Roll back after a failed upgrade
+
+Restoring a snapshot from an earlier version to run a previous {{stack}} version is a different use case. That flow applies when you need to [roll back {{kib}}](/deploy-manage/upgrade/deployment-or-cluster/kibana-roll-back.md) after a failed upgrade, not when you are moving forward to a newer major version. In a rollback, you restore the `kibana` feature state from a snapshot taken before the failed upgrade and start {{kib}} on the earlier version you are rolling back to.
+:::
```
eedugon, line 231: I'm sorry but I don't understand the first part of this admonition!

> Restoring a snapshot from an earlier version to run a previous {{stack}} version

What does it mean? Earlier than what and to a previous version than what?

I have no clue about what we want to say with that.

Also take in mind the basic rule: `You can’t restore a snapshot to an earlier version of Elasticsearch. For example, you can’t restore a snapshot taken in 7.6.0 to a cluster running 7.5.0.`

A rollback would always consist of restoring the snapshot that was taken (or available) before the upgrade to the version that was running before the upgrade, but I don't see the relation with the admonition.

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -158,22 +158,32 @@ You can’t restore a snapshot to an earlier version of {{es}}. For example, you
 
 Any index you restore from a snapshot must also be compatible with the current cluster’s version. If you try to restore an index created in an incompatible version, the restore attempt fails.
```
eedugon, line 159: After this line, in the "INDEX COMPATIBILITY" section, I'd suggest to add some extra scope of what we mean exactly with compatibility, and that compatibility doesn't mean restored data will be valid, as that depends on the data owner expectations (such as Kibana data, but this could happen with other stack features data if they write in their own indices).

Let me explain:

When it comes to `snapshot compatibility` or `index compatibility` we talk about the capacity of Elasticsearch of restoring indices from snapshots created in older versions of Elasticsearch. Compatibility means that Elasticsearch is capable of restoring the original index and data (maybe as read-only in case of archive mode), but that does not mean that the data will be VALID for the application using it in all cases <-- that's the important thing to clarify.

For user data, compatibility will almost always imply that the restored data will be valid to the user, but for Kibana or stack features data this validity that cannot be guaranteed, as the application could expect different data format in different stack versions (such as Kibana). Because of that, users shouldn't use snapshot and restore as a way to "upgrade" except if they are focused only on the data they own.

I love the sentence you included:

> Restoring an old snapshot directly to a 9.x cluster is not a shortcut around the normal [upgrade paths](/deploy-manage/upgrade.md#upgrade-paths). It is recommended you upgrade to the latest compatible minor release before a major upgrade, even when you use snapshots to move data between clusters.

Maybe that sentence could be included as an important note in this index compatibility intro, more than just in the Kibana use case that you are documenting.

Anyway feel free to ignore this if you or the rest of the team believes it doesn't add value, the current PR is really good and adds value!

eedugon (approved):
I like a lot this PR, thanks a lot of adding this "Kibana compatibility" new section.

I've shared a few comments for your consideration but feel free to ignore if you don't feel they add much value.

deploy-manage/tools/snapshot-and-restore.md
yetanothertw: Yes, fixing this in a separate commit. Thank you!

deploy-manage/tools/snapshot-and-restore.md
yetanothertw: Thank you, I'll use your suggestion in a separate commit. I think I'll keep the exception mentioned here just in case it'll help someone on older versions. We know not everyone is using the latest.

deploy-manage/tools/snapshot-and-restore.md
yetanothertw: Updated in a separate commit, thank you for adding those details!

deploy-manage/tools/snapshot-and-restore.md
yetanothertw: lol bad choice of words there, I think I've oversimplified it.

I'll update, thank you

deploy-manage/tools/snapshot-and-restore.md
yetanothertw: Got it! Fixed in a new commit, thanks

deploy-manage/tools/snapshot-and-restore.md
yetanothertw: Sorry for making it so confusing. I've cleaned it up and here's what I was actually trying to say:

```
:::{admonition} Roll back after a failed upgrade
The steps described in [](#move-kibana-config-saved-objects) move {{kib}} state forward to a newer major version. Don't use these steps to roll back after a failed upgrade.

To roll back {{kib}}, restore the `kibana` feature state from a snapshot taken immediately before the failed upgrade, then start {{kib}} on the version that was running before that upgrade attempt. For more information, refer to [Roll back {{kib}}](/deploy-manage/upgrade/deployment-or-cluster/kibana-roll-back.md).
:::
```

yetanothertw: What do you think, @eedugon ?

deploy-manage/tools/snapshot-and-restore.md
yetanothertw: This is such a good comment and observation, you're totally right! We're missing crucial context, so I've tried to add some. Can you please review this again and let me know if this is what you had in mind, @eedugon ? 🙏

deploy-manage/tools/snapshot-and-restore.md
eedugon: Thanks! that new text is perfectly clear and easy to understand!

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -158,22 +158,38 @@ You can’t restore a snapshot to an earlier version of {{es}}. For example, you
...
-| 8.0–8.19                                 | ✅                                        | ✅       | ✅      | ❌       | ❌      | ❌  |
-| {{version.stack.base}}–{{version.stack}} | ✅                                        | ❌       | ❌      | ❌       | ❌      | ❌  | 
+Index compatibility in the context of the snapshot and restore process, indicates that {{es}} can restore the index and its data from the snapshot as a regular index or in a read-only form through [archive indices](/deploy-manage/upgrade/deployment-or-cluster/reading-indices-from-older-elasticsearch-versions.md). It does not mean that every application reading that data will treat it as valid on the target cluster version.
+
+For indices you own and control, a compatible restore usually means the data remains usable. For {{stack}} feature data, including {{kib}} saved objects and data that other features write to their own indices, {{es}} compatibility does not guarantee validity. Those applications can expect different data formats across {{stack}} versions, so a restore that {{es}} accepts can still fail or leave data in an unusable state.
```
eedugon:
```suggestion
For indices and data streams you own and control, a compatible restore usually means the data remains usable. For {{stack}} feature data, including {{kib}} saved objects and data that other features write to their own indices, {{es}} compatibility does not guarantee validity across versions. Those applications can expect different data formats across {{stack}} versions, so a restore that {{es}} accepts can still fail or leave data in an unusable state.
```

eedugon (approved):
Looks great IMO

PR #7253, opened 2026-07-07
Mark PRW endpoint as GA in 9.5 and serverless

felixbarny:
This endpoint is preview since 9.4 (when it has been initially introduced) and GA in serverless and 9.5.

Related to https://github.com/elastic/elasticsearch/pull/151359

### DIFF
--- a/manage-data/data-store/data-streams/tsds-ingest-prometheus-remote-write.md
+++ b/manage-data/data-store/data-streams/tsds-ingest-prometheus-remote-write.md
@@ -1,8 +1,8 @@
 ---
 navigation_title: "Prometheus remote write endpoint"
 applies_to:
-  stack: preview
-  serverless: preview
+  stack: preview =9.4, ga 9.5+
+  serverless: ga
 products:
   - id: elasticsearch
 ---


manage-data/data-store/data-streams/tsds-ingest-prometheus-remote-write.md
```diff
@@ -1,8 +1,8 @@
 ---
 navigation_title: "Prometheus remote write endpoint"
 applies_to:
-  stack: preview
-  serverless: preview
+  stack: preview =9.4, ga 9.5+
```
kkrik-es, line 4:
```suggestion
  stack: preview 9.4, ga 9.5+
```

kkrik-es (approved):

manage-data/data-store/data-streams/tsds-ingest-prometheus-remote-write.md
felixbarny: @lcawl could you consult on the syntax that should be used here? I saw some examples that did use the equals before the version.

leemthompo: @felixbarny you're right about the syntax here.

https://docs-v3-preview.elastic.dev/elastic/docs-builder/tree/main/syntax/applies#formats

The version metadata now supports three operators:

- `=X.x` applies to that version only, for example `=9.4`
- `X.x-Y.y` applies to a version range, for example `9.4-9.6`
- `X.x+` applies to that version and later, for example `9.5+`


I've also updated the ESQL docs README for this: https://github.com/elastic/elasticsearch/blob/main/docs/reference/query-languages/esql/README.md#add-version-metadata

That README should be considered our source of truth for this kind of stuff and we try to update it whenever necessary :-)

leemthompo (approved):
lgtm

PR #7277, opened 2026-07-08
Serverless admin user and API-only key workflow improvements

eedugon:
## Summary

Clarifies the confusion around the built-in `admin` user returned when creating
Serverless projects via the API (see elastic/docs-content-internal#546).

- **`manage-serverless-projects-using-api`**:
  - Reworked "Set up an API key" to specify the minimum roles (Organization owner, or
    Cloud resource Admin scoped to all projects) and the difference between Cloud API
    and Cloud + Elasticsearch/Kibana API access.
  - Added a note explaining what the built-in `admin` user is, why it exists, and
    recommending Elastic Cloud API keys with project access instead.
  - Reframed "Reset credentials" to clarify it resets the `admin` user, with use cases.
- **`differences-from-other-elasticsearch-offerings`**: added a footnote to the
  Native realm authentication row noting the single built-in `admin` user, so the
  comparison table no longer contradicts the API docs.

Closes https://github.com/elastic/docs-content-internal/issues/546

## Generative AI disclosure
- [x] Yes  - Cursor in auto mode
- [ ] No


ppf2:
I think this is a good direction if we have decided to advertise the built-in admin user/password.

That said, I would like to defer to @bytebilly on how we are communicating this, given that it is going into public documentation. Given that, the admin user is:
1. Considered a secret that is exposed in plaintext in our API response.  We may want to warn customers to not log the response from the API call.
2. A built-in user that represents an over privileged default identity. We may want to advise customers to store this user/password in a secret manager.
3. There is no default rotation of password for the admin user. We may want to advise customers to perform rotation themselves (I'm assuming there's a way for them to do this).

Now that I am writing this ... I wonder if we should also seek guidance from our Infosec team on any risks they see in documenting the details publicly 😅

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
```diff
@@ -36,12 +36,20 @@ The available APIs are grouped by project type:
 
 ## Set up an API key [general-manage-project-with-api-set-up-api-key]
 
-1. [Create an API key](https://www.elastic.co/docs/deploy-manage/api-keys/elastic-cloud-api-keys).
+To use the {{serverless-full}} API, you must authenticate your requests with an {{ecloud}} API key.
+
+1. As an **Organization owner**, [create an {{ecloud}} API key](/deploy-manage/api-keys/elastic-cloud-api-keys.md) with one of the following roles, so that it can manage projects:
+
+   - **Organization owner**: can create and manage projects without restriction.
```
wajihaparvez:
```suggestion
   - **Organization owner** to create and manage projects without restriction.
```
Suggesting so format is consistent between both bullet points

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
```diff
@@ -36,12 +36,20 @@ The available APIs are grouped by project type:
 
 ## Set up an API key [general-manage-project-with-api-set-up-api-key]
 
-1. [Create an API key](https://www.elastic.co/docs/deploy-manage/api-keys/elastic-cloud-api-keys).
+To use the {{serverless-full}} API, you must authenticate your requests with an {{ecloud}} API key.
+
+1. As an **Organization owner**, [create an {{ecloud}} API key](/deploy-manage/api-keys/elastic-cloud-api-keys.md) with one of the following roles, so that it can manage projects:
```
wajihaparvez: I find it confusing that this line says "As an Organization owner" and then presents Organization owner an one of two options for roles. Maybe I'm misunderstanding. How does that work?
Maybe adding this as a prereq would be more clear?

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
```diff
@@ -61,7 +69,7 @@ curl -H "Authorization: ApiKey $API_KEY" \
 1. Replace `My project` with a more descriptive name in this call.
 2. You can obtain a [list of available regions](#general-manage-project-with-api-list-available-regions). 
 
-The response from the create project request will include the created project details, such as the project ID, the credentials to access the project, and the endpoints to access different apps such as {{es}} and {{kib}}.
+The response from the create project request will include the created project details, such as the project ID, the endpoints to access different apps such as {{es}} and {{kib}}, and the credentials of a built-in `admin` user.
```
wajihaparvez:
```suggestion
The response from the create project request will include the created project details, including the project ID, the endpoints to access different apps such as {{es}} and {{kib}}, and the credentials of a built-in `admin` user.
```
To avoid the double "such as"

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
```diff
@@ -114,14 +128,18 @@ Example response:
...
+This endpoint resets the password of the built-in `admin` user that is returned when the project is created. Use it when you need to recover or rotate those credentials, for example if you didn't store the password returned at project creation, or if you want to rotate it as part of your security practices.
 
 ```bash
 curl -H "Authorization: ApiKey $API_KEY" \
     -XPOST \
     "https://api.elastic-cloud.com/api/v1/serverless/projects/elasticsearch/${PROJECT_ID}/_reset-credentials"
 ```
 
+:::{tip}
+For most use cases, we recommend authenticating to the {{es}} and {{kib}} APIs with an [{{ecloud}} API key with project access](/deploy-manage/api-keys/elastic-cloud-api-keys.md#project-access) rather than relying on the built-in `admin` user.
```
wajihaparvez: Would it help to xref to the **Create an Elastic Cloud Serverless project** section so that users can learn more from the note you added to it?

wajihaparvez (approved):
LGTM! Can have another look if the content needs to be changed

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
eedugon: Only Org owners can create API keys.

When creating an API key you can give the org owner role to the API key you are creating, or a different type of role.

Let me know if that's unclear in the text.

Note that the scope of this doc is not explaining users that only org owners can create API keys, but we could add it 🤷

deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings.md
```diff
@@ -71,11 +71,13 @@ This table compares core identity, access, and platform capabilities between {{e
...
 | **[API keys](/deploy-manage/api-keys.md)** | ✅ | ✅ | Available across {{ech}} and Serverless using deployment/project and cloud API key types. |
-| **[Native realm authentication](/deploy-manage/users-roles/cluster-or-deployment-auth/native.md)** | ✅ | ❌ | {{serverless-short}} does not support {{es}} authentication realms. User authentication is managed at the [organization level](/deploy-manage/users-roles/cloud-organization.md). |
+| **[Native realm authentication](/deploy-manage/users-roles/cluster-or-deployment-auth/native.md)** | ✅ | ❌ [^1^](#footnote-serverless-admin) | {{serverless-short}} does not support {{es}} authentication realms. User authentication is managed at the [organization level](/deploy-manage/users-roles/cloud-organization.md). |
```
bytebilly: I'd not mention this here, since this is a high-level overview on features and we should be very clear that native realm is not available.
From a technical perspective, the `admin` user is not even in the native realm.

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
bytebilly: FYI this is gonna change in the near future, and more users will be able to create Cloud API keys.

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
bytebilly: > and the credentials of a built-in `admin` user.

We should avoid calling it a user, since this is not a user as others (Elastic Cloud users).
I'm ok mentioning credentials are returned, but can we use a different term instead?

I think this could avoid confusion and false expectations for customers looking for native user support.

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
```diff
@@ -80,6 +88,12 @@ Example of `Create project` response:
 }
 ```
 
+:::{note}
+The `credentials` field contains a built-in `admin` user. This user exists so that API-first workflows can obtain an initial way to authenticate to the project's {{es}} and {{kib}} APIs without any UI interaction, such as logging in to {{kib}} through SSO.
```
bytebilly: Same here, can we avoid the term _user_?

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
```diff
@@ -114,14 +128,18 @@ Example response:
 
 ## Reset credentials [general-manage-project-with-api-reset-credentials]
 
-If you lose the credentials provided at the time of the project creation, you can reset the credentials by using the following endpoint:
+This endpoint resets the password of the built-in `admin` user that is returned when the project is created. Use it when you need to recover or rotate those credentials, for example if you didn't store the password returned at project creation, or if you want to rotate it as part of your security practices.
```
bytebilly: Same here, can we avoid the term user?

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
bytebilly: Same here, can we avoid the term user?

bytebilly (changes requested):
I'd like to consider framing the `admin` credentials as not being a _user_, since they're not from a high-level perspective. They're more similar to an API key, and they should be used for programmatic access only.

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
eedugon: totally agreed, I'll try to find a way to achieve it, thanks!

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
```diff
@@ -36,12 +35,20 @@ The available APIs are grouped by project type:
...
 ## Set up an API key [general-manage-project-with-api-set-up-api-key]
 
-1. [Create an API key](https://www.elastic.co/docs/deploy-manage/api-keys/elastic-cloud-api-keys).
+To create and manage projects with the {{serverless-full}} API, you must authenticate your requests with an {{ecloud}} API key.
+
+1. As an **Organization owner**, [create an {{ecloud}} API key](/deploy-manage/api-keys/elastic-cloud-api-keys.md) with one of the following roles:
+
+   - **Organization owner**
+   - **Cloud resource access** with the **Admin** role assigned to **all projects** of the relevant type ({{es}}, {{observability}}, or Security)
+
```
eedugon, line 44: @wajihaparvez , I've refined this a bit, let me know your thoughts:

- I've stated the objective (create and manage projects) in the intro instead of at the end of the step 1.
- That way step 1 gets more attention in the "with one of the following roles:", to ensure the user understand that both options are valid.
- Then in both roles I've removed the explanation around them. The org owner explanation was a bit obvious, and the second one maybe added some value but I consider it unneeded, and removing it helps with making the bullets symmetric.

let me know your thoughts.

Note that I don't like the pre-requisites section idea, because this sub-section is actually presented as a kind of pre-requisite for the rest of the document at the beginning:

> To try the examples in this section, start by [setting up an API key](#general-manage-project-with-api-set-up-api-key).

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
eedugon: sub-section removed

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
eedugon: Updated to:

> The response from the create project request will include the created project details, including the project ID, the endpoints to access different apps such as {{es}} and {{kib}}, and a set of default credentials.

Feel free to comment out in the new version.

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
eedugon: Note replaced in alignment with the previous change.

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
eedugon: Section removed, we just link to the reset credentials API endpoint in the previous note.

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
eedugon: section removed

bytebilly (approved):
LGTM 👍

deploy-manage/deploy/elastic-cloud/manage-serverless-projects-using-api.md
wajihaparvez: @eedugon Cool, I think this is more clear

eedugon:
@ppf2 , if possible we'd like to get your final review here after the latest updates.
Thanks!

ppf2:
@kunisen Would you mind lending a second pair of eyes on this? I know this is a topic that's close to your heart 🙂

kunisen:
Thanks @ppf2 for your comment 🙏 

### Overall

@bytebilly has already given LGTM, so I am also good with the statement :) 
Based on that, thinking in customer situation, I may have below two thoughts:

---

### [1]

@eedugon I have one question:
* Does `admin` have programmatic access to the project's Elasticsearch and Kibana APIs?

I guess no? (Because API key with ES/Kibana API access is fairly recent feature IIRC).

Then WDYT we add:
* This username/password credential also does not have the access to the project's Elasticsearch and Kibana APIs
* (or statements logically similar)

So that we can make it more clear that this credential actually is not at the same level of API keys, to make it more explanatory of why we recommend API keys.

(I can see it's also in line with what Seb said in [this comment](https://github.com/elastic/sdh-elasticsearch/issues/9510#issuecomment-3584683099) - `there won't be a need for the 'admin' basic auth user anymore`. So IMHO, good to emphasize it more explicitly.)

---

### [2]

> when no API key is available

@eedugon per the comment from Seb, I take the reason of keeping admin user is _"for backward compatibility reasons only (don't break the automation of existing customers)"_.

Do you think it's good to add this into the doc? Something like, 
* For historical reasons, there were situations of no API key scenarios. 
* But from now on, we recommend every use case to use API key. 
* This credential should only be used for for backward compatibility reasons, etc.

---

It doesn't mean I am against with the statement. 
The goal is to make things more clear based on Seb's comment, as well as the customer perspective. (given them more technical and logical evidence of why we recommend API key).

Hope it makes logical sense and thank you!

eedugon:
@kunisen , I'll allow @bytebilly to add more context if he considers necessary.

> Does admin have programmatic access to the project's Elasticsearch and Kibana APIs? I guess no?

Yes it does, not as an API key, but with user / password authentication. And precisely because of that we don't want to highlight the user and password or examples too much. We just say that they are default credentials to access the projects (which is true and valid), that they should be kept safe, and that they can be reset.

> per the comment from Seb, I take the reason of keeping admin user is "for backward compatibility reasons only (don't break the automation of existing customers)"

I don't see the reason to add such a context, but if @bytebilly wants we can somehow tweak this sentence and make it a bit longer:

> The default credentials returned in the `credentials` field serve as a fallback mechanism to access the project when no API key is available.

Update: I forgot about this:

> The goal is to make things more clear based on Seb's comment

That was the goal I had in mind initially, but it's not the real goal here, sorry for the confusion.

The goal is to make user our users use Elastic Cloud API keys to access Elasticsearch and Kibana programmatically, and don't use these credentials. The secondary goal is to reduce the visibility of these credentials as an option. But we have decided to not remove them entirely because they are still returned by the create project API.

bytebilly:
I'm not sure we want to mention backward compatibility, since in Serverless it's always been the case.

kunisen (approved):
Thanks both! Now I got the full picture and everything is clear on my side! 🙇  It looks great!

eedugon:
Perfect, merging then! Thanks all!

PR #7311, opened 2026-07-09
Add troubleshooting entry for combined index privilege patterns

### LINKED ISSUE #599: [Request]: Privileges created using Security Solution's alerts dataview pattern
### What documentation page is affected

Troubleshoot Detection Rules  (https://www.elastic.co/guide/en/security/current/ts-detection-rules.html)

### What change would you like to see?

The broader issue with this UI creating misconfigured roles is captured [here](https://github.com/elastic/kibana/issues/95009); this issue is requesting that we add a callout in our troubleshooting documentation that would allow users to diagnose/correct this situation.

As demonstrated in [the inciting issue](https://github.com/elastic/kibana/issues/211123), the user would likely see a

> This rule may not have the required read privileges

warning where they would otherwise not expect one, and as mentioned in the kibana issue the solution is to "itemize" their index privileges, i.e.: 
  * `auditbeat-*,filebeat-*: read/write` becomes:
     * `auditbeat-*: read/write` and
     * `filebeat-*: read/write`

### Additional info

_No response_

bmorelli25:
## Summary

Adds a troubleshooting entry to the [Troubleshoot detection rules](https://www.elastic.co/docs/troubleshoot/security/detection-rules) page for the spurious `This rule may not have the required read privileges` warning.

- Root cause: index privileges defined as a single comma-separated pattern (for example, `auditbeat-*,filebeat-*`) aren't granted to each pattern individually, so the rule's privilege check fails.
- Fix: itemize the privileges into separate role entries, then edit and save the rule to refresh its privileges.

Adds a new **Rule privileges** section with a dropdown keyed to the warning text.

Closes #599


troubleshoot/security/detection-rules.md
```diff
@@ -47,6 +47,25 @@ For **self-managed** {{stack}} deployments only, this message might display when
 ::::
 
 
+## Rule privileges [rule-privileges-ts]
+
+::::{dropdown} Warning about required read privileges for indices you can access
+:name: rule-privileges-combined-patterns-ts
+
+A rule might warn that it may not have the required read privileges for index patterns that your role clearly includes, such as:
```
mdbirnstiehl:
```suggestion
You might receive the following warning for a rule even when your role clearly includes privileges for the index patterns:
```
Not sure if this works completely, but the "might" and "may" were messing me up when trying to understand. Up to your discretion, of course!

mdbirnstiehl (approved):
One suggestion, otherwise LGTM

troubleshoot/security/detection-rules.md
bmorelli25: Thanks. This is better,

PR #7313, opened 2026-07-09
ADD SAML works ECH-agnostic

stefnestor:
## Summary
👋 howdy, team! 

This appears to be doc merge carry over, but Active Directory SAML works for all `stack` (same as top-of-page labels) and not just ECH as is referenced in first paragraph and throughout page.

Also ECH+ECE no longer have `elasticsearch.yml` per node tier, it's merged into one +3y.

TIA! Stef
## Generative AI disclosure
- [ ] Yes  
- [X] No  

Self note, re: 02114479


eedugon:
@copilot : analyze this PR and check the modified file to see if there are more areas that look attached to ECH or ECE, as the doc should be deployment agnostic. The current PR improves this by removing a few unneeded references to ECH details that don't add any value.

deploy-manage/users-roles/cluster-or-deployment-auth/saml-entra.md
```diff
@@ -5,11 +5,14 @@ mapped_pages:
 applies_to:
   stack: all
 products:
+  - id: elasticsearch
+  - id: cloud-enterprise
   - id: cloud-hosted
+  - id: cloud-kubernetes
```
eedugon: Maybe we have to add kibana there, and not sure what to do with cloud-enterprise, cloud-hosted, and cloud-kubernetes.

SAML combines elasticsearch and kibana, and this doc is applicable to all deployment types (`stack` already there), but from product point of view my first instinct would be to attach it to Elasticsearch and Kibana only.

@shainaraskas , WDYT?

eedugon:
```suggestion
  - id: elasticsearch
  - id: kibana
```

Updating list of products per internal discussion.

eedugon: Already discussed with @georgewallace , thanks!

eedugon:
@copilot , please revert the changes on lines 47, 51, and 55. We are ok with the original URL examples there.

eedugon (commented):
@stefnestor : Thanks for raising this, it's a good catch.

I've applied a few extra updates (through Copilot) to your PR with the same goal you've shared in the description. Let me know if you agree with them, and then we can do a final review.

deploy-manage/users-roles/cluster-or-deployment-auth/saml-entra.md
eedugon: @shainaraskas , I'd also like to get your view on this (and also learn a bit about products metadata =) ).

This is a SAML document, applicable to all deployment types (applies_to: stack).
From product point of view I'd say SAML touches both Elasticsearch and Kibana, that's why I'm suggesting that.

The original guide was written with ECH in mind and only had `cloud-hosted` in the products list (🤷 ). Does it make sense to include `product: cloud-hosted`, or ECE or ECK? It would look redundant in my opinion, but not sure what exact end goals we have with the `products` metadata.

shainaraskas: I think that you're right - we can use a minimal list of products. I've advocated with engineering to view elasticsearch/kib features as available across deployment types when it comes to searching and filtering (I realize it's more nuanced than that but I think you see where I'm going)

eedugon (approved):
LGTM!

PR #7104, opened 2026-06-24
Application connections tech preview (A.K.A. MCP OAuth)

shainaraskas:
## What this adds

- Fixes https://github.com/elastic/docs-content-internal/issues/1261

Docs for connecting MCP hosts  to the agent builder MCP server with OAuth, alongside the existing API key method.

**New pages:**

- [OAuth for MCP clients](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7104/deploy-manage/app-connections/oauth-clients): overview and how the flow works
- [Create an MCP client](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7104/deploy-manage/app-connections/create-oauth-client)
- [Connect an MCP host](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7104/deploy-manage/app-connections/connect-mcp-host)
- [Revoke a client or connection](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7104/deploy-manage/app-connections/revoke-oauth-client)
- [Manage application connections](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7104/deploy-manage/app-connections/manage-app-connections): org-level view in the Cloud Console

**Also:** split API key setup into its own [MCP server API keys](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7104/explore-analyze/ai-features/agent-builder/mcp-server-api-keys) page and trimmed [MCP server](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7104/explore-analyze/ai-features/agent-builder/mcp-server) to an overview plus an auth method comparison. Minor crosslink, terminology, and [permissions](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7104/explore-analyze/ai-features/agent-builder/permissions) updates.

## How it's organized, and why

- Setup and concepts live under **Application connections** in Deploy & Manage, since app connections are a Cloud/org concept and the org-level view is in the Cloud Console. This will allow the concepts to be easily expanded upon when we roll out support for CLI OAuth etc. 
- The MCP server itself and the OAuth vs. API keys comparison stay in Agent Builder docs, where the feature lives. **I'll also be crosslinking the OAuth page below the mcp server page in agent builder**, but can't do that until after this merges due to tooling limitations 
- One overview page on top, task pages (create, connect, revoke) beneath it.

## Open questions

[...]

PR #7329, opened 2026-07-13
Add OAuth clients docs URL stub

leemthompo:
## Summary

Adds a temporary hidden stub for /deploy-manage/app-connections/oauth-clients so elastic/kibana#277286 can link to a live docs URL before the full OAuth/application connections docs in #7104 merge.

## Changes

- Adds a hidden TOC entry for deploy-manage/app-connections/oauth-clients.md.
- Adds a minimal page with front matter and the oauth-clients anchor only.

## Follow-up

Replace this stub with the full content from #7104 when that PR is ready.


yetanothertw (approved):
👍

PR #6523, opened 2026-05-15
[Alerting V2][Serverless & 9.5][M2] Add experimental alerting features rules pages

nastasha-solomon:
## Summary

Adds rules documentation for the experimental alerting system, contributing to https://github.com/elastic/docs-content-internal/issues/919. Following the principle of accuracy over comprehensiveness, pages focus on stable concepts (for example, what rules do, when to choose each mode or creation path, and how detection and alerting work) rather than UI step-by-step procedures that are subject to change before GA.

## Navigation structure

The rules section is organized into four areas. Reference pages sit at the end of the section so the navigational flow moves from creating a rule to configuring it to managing it, before reaching advanced patterns and field references.

```
Rules          ← section landing page
├── Create a rule
│   ├── Create an ES|QL rule
│   │   └── YAML rule schema reference
│   ├── Create rules and action policies with Elastic Agent Builder
│   ├── Create rules using the rule builder
│   └── Create from Discover
├── Configure a rule
│   ├── Rule mode
│   ├── ES|QL query
│   ├── Severity
│   ├── Grouping
│   ├── Schedule and lookback
│   ├── Alert delay (Alert mode only)
│   ├── Recovery condition (Alert mode only)
│   ├── No-data handling
│   └── Tags and runbooks (Alert mode only)
├── View and manage rules
├── Review execution history
├── ES|QL query patterns
│   ├── Your first rule query
│   ├── Threshold queries
│   ├── No-data detection
│   ├── SLO burn rate detection
│   └── Persistent breach detection
└── Rule event and field reference
```

## Review requests

This PR needs a **technical** and **editorial** review. Instructions for each are below.

### 🔧 Technical reviewer

Please focus your review on accuracy of facts, names, and structure. Verify that field names, values, and descriptions in reference material match the current implementation. When reviewing conceptual content, check that definitions are technically accurate and reflect the current engineering design.

### ✏️ Editorial reviewer

This PR adds the full rules documentation set for the experimental alerting system. As you're reviewing, please focus on writing quality, clarity, and how the content serves a reader who's seeing this system for the first time. 

To understand the experimental alerting system basics and foundational concepts, refer to the overview and setup docs below:

**Overview and setup docs**

- [Alerting](https://www.elastic.co/docs/explore-analyze/alerting)
   - [Compare alerting systems](https://www.elastic.co/docs/explore-analyze/alerting/compare-alerting-systems)
   - [Experimental alerting system](https://www.elastic.co/docs/explore-analyze/alerting/system-overview)
[...]

PR #7375, opened 2026-07-15
[Alerting V2][Serverless & 9.5][M2] Fix forced-recovery step in tutorial

nastasha-solomon:
## Summary

Jason flagged that the tutorial's force delete step didn't reliably produce genuine recovery:

- **Deleting the degraded documents (`_delete_by_query`) left zero rows for `checkout`** (flagged in a [Slack convo](https://elastic.slack.com/archives/C0BHG8NKRV2/p1784073571482409?thread_ts=1784070494.253129&cid=C0BHG8NKRV2)), which the system treats as "no data" rather than genuine recovery. Fix: switched to an `_update_by_query` call that rewrites `latency_ms` to a healthy value (300ms), keeping a row for `checkout` just under the threshold so the rule evaluates a real recovery condition.
- **Waiting for the degraded data to age out of the lookback window isn't guaranteed to produce the same recovery** (flagged in PR review), but the tutorial's wording implied it would. Fix: made the **Force recovery** step mandatory instead of optional, and removed the wording that framed waiting as an equivalent path.

I also made some minor improvements to the tutorial overall:
- The "Key concepts" section was reframed to convey learning outcomes ("By completing this tutorial, you learned...") instead of recapping tutorial actions.
- The rule lookback window was shortened from 2 hours to 45 minutes so users don't need to wait as long to witness the automatic recovery behavior. 

**NOTE:** [kibana#277561](https://github.com/elastic/kibana/pull/277561) prevents the alert episode from switching to `inactive` at the end of the tutorial. That PR was recently merged, so the final lifecycle change should happen as the tutorial describes once the fix is available in serverless.

## Preview

[Create a rule and observe the alert lifecycle](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7375/explore-analyze/alerting/experimental-alerting-system/get-started/create-your-first-rule)

## Generative AI disclosure

- [x] Yes - Cursor + Claude 
- [ ] No


explore-analyze/alerting/experimental-alerting-system/get-started/create-your-first-rule.md
```diff
@@ -390,54 +384,43 @@ Open **Alerting v2 preview** using the navigation menu or the [global search fie
...
-
-Because the lookback window is 2 hours, this can take a while. The degraded data doesn't age out until roughly 2 hours after you loaded it. If you don't want to wait, follow the instructions in the next step to force the same outcome immediately.
+Select the episode to open its details. Use the metric trend to see how P95 latency compared to the threshold over the episode's lifetime, and confirm the grouping value (`checkout`) that triggered it.
 
 ::::
 
 ::::{step} (Optional) Force recovery immediately
 
-Run the following in **Dev Tools** to delete the degraded documents:
+Rather than waiting about 45 minutes for the degraded data to age out of the lookback window naturally, run the following in **Dev Tools** to rewrite the degraded documents' `latency_ms` values to a healthy level:
```
jasonrhodes: I would only note that this phrasing makes it sound like waiting those 45 minutes would _also_ lead to a recovery, but I can't say for sure if it would, I think it would be active until all of the docs fall out of the lookback window at which point it would become "no_data" again. Not sure if that makes you want to reword or not, just calling it out.

Also to be clear, there is a no data strategy option called "recover" which solves all of this for a user, but it doesn't test the "true recovery" flow in the same way that meeting the actual recovery condition does.

jasonrhodes (approved):
LGTM, for what it's worth you could also create a seed script that contains a spike and then a natural resolution in it, so that the user _can_ choose to wait for recovery to happen on its own without needing to overwrite the values. I think the overwrite method is fine, but if you wanted to allow for recovery without the extra query, I would just have

1. A script that seeds data in the recent past, but with healthy values - run this first
2. Check Discover, build the rule and see the data in the sandbox, again, no breaches in that view
3. Create the rule, no alerts are created
4. Second script that seeds data "right now" that starts off with a bad value, then gradually tapers off to a good value over ... 10 minutes? 5 minutes? 2 minutes? Whatever you think
5. The alert will fire almost immediately (or start pending -> active) stay active, then after that period of time, go to recovering, then inactive 

Again, not necessary, just sharing in case that was what you were hoping to achieve.

explore-analyze/alerting/experimental-alerting-system/get-started/create-your-first-rule.md
nastasha-solomon: That's a good point. There's no guarantee that waiting for 45 mins (or whatever the lookback time is) will produce a real recovery. As you mentioned, the rule might just report that there's no data to query and the episode would stay `Active`. 

I think the solution here is make the forced recovery step mandatory ([step 4](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7375/explore-analyze/alerting/experimental-alerting-system/get-started/create-your-first-rule#observe-episode-lifecycle)). That guarantees the rule's next scheduled run computes a P95 latency computes a P95 latency below the `WHERE p95_latency_ms > 2000` threshold, finds no breaching rows, finds no breaching rows, and then moves the episode's lifecycle state to `Inactive`. 

I'll revise this step and other areas in the tutorial that imply a natural recovery will happen if the user waits for the data to age out of the lookback window.

nastasha-solomon: Addressed in [`f0735b0` (this PR)](https://github.com/elastic/docs-content/pull/7375/commits/f0735b02f6e8230f42b4ced07dc4aaf24b737996).

explore-analyze/alerting/experimental-alerting-system/get-started/create-your-first-rule.md
```diff
@@ -319,7 +317,7 @@ Confirm the default settings:
 - **Recovery**: `Default recovery`
 - **Recovery delay**: `Immediate` (no delay, recovers on first non-breach)
 
-These default settings will produce the automatic recovery behavior this tutorial demonstrates. As soon as a scheduled run returns no breaching rows, the episode will close.
+These default settings will produce the automatic recovery behavior this tutorial demonstrates. As soon as a scheduled run's query returns no rows (because no service's P95 latency clears the `WHERE p95_latency_ms > 2000` threshold), the episode will close.
```
bmorelli25: I'm probably spending too much time on this but I'm confused and I want to make sure IUC.

On every run the rule is asking two q's about `checkout`:
1. Is there any `checkout` data to look at?
2. If yes, is `checkout` latency over or under 2000.

To (a) keep an alert open or (b) close the alert, the alert needs to answer Q2. BUT it can only answer Q2 if Q1 is `yes`. So there are three "states":

1. `checkout` has data, latency > 2000 → **breaching** → alert stays open
2. `checkout` has data, latency < 2000 → **not breaching** → alert closes
3. `checkou`t has no data → **can't tell** → alert sits there

tl;dr the alert only closes in the middle case (option2) because the rule has to see data to know.

This is why you've updated the dev tool command below from `delete` to `update`:

- `delete` deletes checkout's docs so the rule can't tell because there's nothing to eval (option3)
- `update` keeps the document but lowers their latency to 300. this is option2 and the alert can then close.

OKAY NOW TO MY CONFUSION

You state:
>As soon as a scheduled run's query returns no rows (because no service's P95 latency clears the `WHERE p95_latency_ms > 2000` threshold), the episode will close.

But this query will only list services that are actively breaching. 

And now that I've typed alllllll of this out I understand what this line is saying. You're saying as soon as no rows are returned (in the case of ``no service's P95 latency clears the `WHERE p95_latency_ms > 2000` threshold`` and only in the case of that), then the episode will close.

I think we can reword this to be clearer. Maybe I was just getting stuck on the no rows part. Idk maybe remove that?

> These default settings will produce the automatic recovery behavior this tutorial demonstrates. As soon as a scheduled run evaluates checkout and its P95 latency no longer clears the WHERE p95_latency_ms > 2000 threshold, the episode will close.

bmorelli25 (approved):
You can probably just ignore my comment. 🚢

explore-analyze/alerting/experimental-alerting-system/get-started/create-your-first-rule.md
nastasha-solomon: I went over that explanation a few times too and debated whether to include the line about no rows returning. 😅 While keeping that detail in makes the explanation more precise, it does feel too low-level for the tutorial, which is really just meant to demonstrate basic concepts about the experimental alerting system. Here, the basic concept that users need to understand is that the rule needs to find data indicating that the threshold is no longer being exceeded. I'll try to simplify it.

nastasha-solomon: Addressed in [`179a163` (this PR)](https://github.com/elastic/docs-content/pull/7375/commits/179a1632171e29bc720574ea6ad8e5aa3862ef29).

jasonrhodes: You all have successfully gone through the journey to understanding our no data handling logic! Keep this all in mind when you look at (a) no data strategy options, and (b) our insistence in some places on having a "base query". 

I realize these concepts are very confusing so I want to take this opportunity to use this exact example, while it's fresh in your minds, to explain how these concepts come together. Bear with me! :P 

Each rule declares a recovery strategy and a no data strategy. Default recovery strategy is called "no_breach", which means that the full "breach query" just no longer returns the result anymore. But how do we know if the result didn't come back because `p95_latency_ms` fell below the threshold (yay), or if the document just stopped appearing (because the service crashed, etc)? We need to remove the "condition" part of the query, the `WHERE p95_latency_ms > 2000`, and run JUST the "base query" -- if that one returns results for this alert (we don't have to check the value, we know it's below now), this alert has recovered. If we don't get an expected result from this base query, either, we're in "no data land". 

So recapping there, assuming the recovery strategy is set to the default "no_breach" setting and there is a specified "no_data" strategy, we need a base query that is separated from the full breach query in order to test a possible no breach recovery to make sure it isn't actually due to the data disappearing.

Possible no data strategies are:

* Keep last status - just leave it as active in our case above (this is the default and what led to the problem Nastasha found in the system the other day)
* Recover - we expect these situations, so if you detect no data, just treat it as a recovery
* Do nothing (arguably a confusing label) - this actually means "don't even check if there is no data", so we wouldn't even run the base query, and we'd also just recover in the above example, assuming the lack of results is just a recovery.

To round this massive post out, there are also a few other recovery strategies:

* "No Breach" - default from the UI, recovers when the breach query stops returning results
* "Custom recovery" - user has defined a different threshold value or combination of values, so we do a separate query to see if the row is returned. In this scenario, we MUST get a row back in the recovery query to recover. If we don't, we check the base query to see if the value is there and if so, we keep the alert active. If it's not in the base query, we flow back into the no data options.
* "None" (or not supplied at all) - here we just don't worry about recovery at all - this is the default in the API

bmorelli25: Okay this _really_ helps clarify things. Do we cover any of these recovery or no data strategies in the docs? @nastasha-solomon we should doc these if not.

I presume this is the `no_breach | query | none` and `last_known_status | emit | recover | none` in https://github.com/elastic/kibana/blob/main/x-pack/platform/packages/shared/response-ops/alerting-v2-schemas/src/rule_data_schema.ts?

jasonrhodes: Yep it's those. I think we removed "emit" for M2 and may or may not add it back, need to update that doc and maybe other places once we confirm

nastasha-solomon: The recovery and no data strategies are currently covered in https://github.com/elastic/docs-content/pull/6523 ([recovery thresholds](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/6523/explore-analyze/alerting/experimental-alerting-system/rules/configure-rule-thresholds) and [no data handling](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/6523/explore-analyze/alerting/experimental-alerting-system/rules/configure-no-data-handling)). I do need to adjust the recovery docs though because they don't explain when users would choose certain options. I can make those adjustments today and re-submit the rule docs for a quick validation, @jasonrhodes.

PR #7475, opened 2026-07-21
[ECK]: 3.5.0 document file-based secure settings hot-reload for Elasticsearch

pkoutsovasilis:
## Summary

Documents the opt-in file-based secure settings hot-reload for Elasticsearch 9.5+ on ECK 3.5, covering the annotation, compatible settings, and the important caveat about startup-required settings.

## Generative AI disclosure

- [x] Yes  
- [ ] No  

Tool(s) and model(s) used: Sonnet 4.6

### DIFF
--- a/deploy-manage/security/k8s-secure-settings.md
+++ b/deploy-manage/security/k8s-secure-settings.md
@@ -158,7 +158,7 @@ spec:
   - secretName: s3-credentials
 ```
 
-When the annotation is set and the cluster is running {{es}} 9.5+, ECK delivers all `spec.secureSettings` entries through {{es}} file-based settings instead of the keystore init container. Updating a source secret no longer triggers a rolling restart. Instead, {{es}} applies the updated credentials in place on each node.
+When the annotation is set and the cluster is running {{es}} 9.5+, ECK delivers all `spec.secureSettings` entries through {{es}} file-based settings instead of the keystore init container. Updating a source secret no longer triggers a rolling restart. Instead, {{es}} applies the updated credentials in place on each node. Note that enabling or disabling the annotation itself causes a one-time rolling restart, because it changes the Pod template by adding or removing the keystore init container.
 
 When the annotation is absent or the cluster is running {{es}} earlier than 9.5, the existing keystore init container path is used unchanged.
 
--- a/deploy-manage/security/k8s-secure-settings.md
+++ b/deploy-manage/security/k8s-secure-settings.md
@@ -168,7 +168,7 @@ When the annotation is set and the cluster is running {{es}} 9.5+, ECK delivers 
 When the annotation is absent or the cluster is running {{es}} earlier than 9.5, the existing keystore init container path is used unchanged.
 
 ::::{important}
-Only use this annotation when **all** entries in `spec.secureSettings` are hot-reloadable settings. Settings that must be present in the keystore at startup — such as OIDC `client_secret`, SAML key passphrases, and `xpack.watcher.encryption_key` — will cause {{es}} to fail to start if the keystore is absent. Typical reloadable settings include S3, GCS, and Azure repository credentials, and remote-cluster API keys.
+Only use this annotation when all entries in `spec.secureSettings` are [reloadable settings](/deploy-manage/security/secure-settings.md#reloadable-secure-settings). Settings that must be present in the keystore at startup, such as OIDC `client_secret`, SAML key passphrases, and `xpack.watcher.encryption_key`, will cause {{es}} to fail to start if the keystore is absent.
 ::::
 
 ## {{kib}} secure settings [k8s-kibana-secure-settings]
--- a/deploy-manage/security/k8s-secure-settings.md
+++ b/deploy-manage/security/k8s-secure-settings.md
@@ -15,12 +15,7 @@ With the help of ECK operator, you can specify {{es}} and {{kib}} [secure settin
 
 The secrets should contain a key-value pair for each secure setting you want to add. ECK watches the referenced secrets for changes and delivers them to your {{es}} or {{kib}} Pods. By default, each update triggers a rolling restart of the affected Pods to repopulate the keystore.
 
-```{applies_to}
-deployment:
-  eck: ga 3.5+
-```
-
-For {{es}} 9.5 and later, you can opt in to [updating secure settings without a restart](#k8s-es-secure-settings-hot-reload).
+You also can opt in to [updating secure settings without a restart](#k8s-es-secure-settings-hot-reload).
 
 ## {{es}} secure settings [k8s-es-secure-settings]
 
--- a/deploy-manage/security/k8s-secure-settings.md
+++ b/deploy-manage/security/k8s-secure-settings.md
@@ -128,6 +128,40 @@ stringData:
     }
 ```
 
+## {{es}} secure settings hot-reload [k8s-es-secure-settings-hot-reload]
+```{applies_to}
+deployment:
+  eck: ga 3.5
+```
+
+By default, every change to a `spec.secureSettings` source secret triggers a rolling restart of the {{es}} cluster because the keystore init container must re-run to repopulate the keystore. For {{es}} 9.5 and later, ECK supports an opt-in file-based delivery mechanism that eliminates this restart: secrets are written into the {{es}} file-based settings path and {{es}} reloads them in place when the file changes.
+
+To enable file-based delivery, add the following annotation to your {{es}} resource:
+
+```yaml subs=true
+apiVersion: elasticsearch.k8s.elastic.co/v1
+kind: Elasticsearch
+metadata:
+  name: quickstart
+  annotations:
+    eck.k8s.elastic.co/file-based-secure-settings: "true"
+spec:
+  version: {{version.stack}}
+  nodeSets:
+  - name: default
+    count: 3
+  secureSettings:
+  - secretName: s3-credentials
+```
+
+When the annotation is set and the cluster is running {{es}} 9.5+, ECK delivers all `spec.secureSettings` entries via the `cluster_secrets` field of the file-based settings mechanism instead of the keystore init container. Updating a source secret no longer triggers a pod restart — {{es}} reloads the credentials in place on each node.
+
+::::{important}
+Only use this annotation when **all** entries in `spec.secureSettings` are hot-reloadable settings. Settings that must be present in the keystore at startup — such as OIDC `client_secret`, SAML key passphrases, and `xpack.watcher.encryption_key` — will cause {{es}} to fail to start if the keystore is absent. Typical reloadable settings include S3, GCS, and Azure repository credentials, and remote-cluster API keys.
+
+When the annotation is absent or the cluster is running {{es}} earlier than 9.5, the existing keystore init container path is used unchanged.
+::::
+
 ## {{kib}} secure settings [k8s-kibana-secure-settings]
 
 Similar to {{es}} secure settings, you can use Kubernetes secrets to manage keystore settings for {{kib}}.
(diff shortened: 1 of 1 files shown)


deploy-manage/security/k8s-secure-settings.md
```diff
@@ -128,6 +128,40 @@ stringData:
...
+spec:
+  version: {{version.stack}}
+  nodeSets:
+  - name: default
+    count: 3
+  secureSettings:
+  - secretName: s3-credentials
+```
+
+When the annotation is set and the cluster is running {{es}} 9.5+, ECK delivers all `spec.secureSettings` entries via the `cluster_secrets` field of the file-based settings mechanism instead of the keystore init container. Updating a source secret no longer triggers a pod restart — {{es}} reloads the credentials in place on each node.
```
shainaraskas:
```suggestion
When the annotation is set and the cluster is running {{es}} 9.5+, ECK delivers all `spec.secureSettings` entries through {{es}} file-based settings instead of the keystore init container. Updating a source secret no longer triggers a rolling restart. Instead, {{es}} applies the updated credentials in place on each node.
```

deploy-manage/security/k8s-secure-settings.md
```diff
@@ -128,6 +128,40 @@ stringData:
...
+
+::::{important}
+Only use this annotation when **all** entries in `spec.secureSettings` are hot-reloadable settings. Settings that must be present in the keystore at startup — such as OIDC `client_secret`, SAML key passphrases, and `xpack.watcher.encryption_key` — will cause {{es}} to fail to start if the keystore is absent. Typical reloadable settings include S3, GCS, and Azure repository credentials, and remote-cluster API keys.
```
shainaraskas: we can link to the definition

```suggestion
Only use this annotation when all entries in `spec.secureSettings` are [reloadable settings](/deploy-manage/security/secure-settings.md#reloadable-secure-settings). Settings that must be present in the keystore at startup, such as OIDC `client_secret`, SAML key passphrases, and `xpack.watcher.encryption_key`, will cause {{es}} to fail to start if the keystore is absent.
```

deploy-manage/security/k8s-secure-settings.md
```diff
@@ -128,6 +128,40 @@ stringData:
...
+  nodeSets:
+  - name: default
+    count: 3
+  secureSettings:
+  - secretName: s3-credentials
+```
+
+When the annotation is set and the cluster is running {{es}} 9.5+, ECK delivers all `spec.secureSettings` entries via the `cluster_secrets` field of the file-based settings mechanism instead of the keystore init container. Updating a source secret no longer triggers a pod restart — {{es}} reloads the credentials in place on each node.
+
+::::{important}
```
shainaraskas: movin' this up as mentioned below

```suggestion
When the annotation is absent or the cluster is running {{es}} earlier than 9.5, the existing keystore init container path is used unchanged.

::::{important}
```

deploy-manage/security/k8s-secure-settings.md
```diff
@@ -128,6 +128,40 @@ stringData:
...
+Only use this annotation when **all** entries in `spec.secureSettings` are hot-reloadable settings. Settings that must be present in the keystore at startup — such as OIDC `client_secret`, SAML key passphrases, and `xpack.watcher.encryption_key` — will cause {{es}} to fail to start if the keystore is absent. Typical reloadable settings include S3, GCS, and Azure repository credentials, and remote-cluster API keys.
+
+When the annotation is absent or the cluster is running {{es}} earlier than 9.5, the existing keystore init container path is used unchanged.
```
shainaraskas, line 163: this doesn't need to be in the important note
```suggestion
```

deploy-manage/security/k8s-secure-settings.md
```diff
@@ -128,6 +128,40 @@ stringData:
     }
 ```
 
+## {{es}} secure settings hot-reload [k8s-es-secure-settings-hot-reload]
```
shainaraskas:
```suggestion
## Update {{es}} secure settings without a restart [k8s-es-secure-settings-hot-reload]
```

shainaraskas: not really your problem, but the IA of this page is getting a little funky. we can try to group the elasticsearch-related settings into a single h2 for clarity

```
## {{es}} secure settings

Reference one or more Kubernetes secrets from `spec.secureSettings` so ECK can inject secure settings into your {{es}} Pods. You can [add secrets to the resource](#k8s_basic_usage), [map secret keys to keystore setting names](#k8s_projection_of_secret_keys_to_specific_paths), and [update secure settings without a rolling restart](#k8s-es-secure-settings-hot-reload).

### Reference secrets in the Elasticsearch resource
### Project secret keys to specific paths
### Update secure settings without a restart
```

shainaraskas: I think this new behavior invalidates the overview. probably needs to be softened

> The secrets should contain a key-value pair for each secure setting you want to add. ECK automatically injects these settings into the keystore on each Elasticsearch or Kibana Pod before it starts. The ECK operator continues to watch the secrets for changes and will update the Elasticsearch or Kibana keystores when it detects a change.
>
> To allow the operator to inject the settings into the application, you must reference your secrets in the spec.secureSettings field of your Elasticsearch or Kibana object definition. Next, you’ll find examples for both Elasticsearch and Kibana.

shainaraskas (approved):
👋 some comments for your consideration :)

deploy-manage/security/k8s-secure-settings.md
pkoutsovasilis: Folded into the IA restructure in a533066

pkoutsovasilis: Changed in a533066

pkoutsovasilis: Softened in 1b937e0

deploy-manage/security/k8s-secure-settings.md
pkoutsovasilis: Applied in 181c123

deploy-manage/security/k8s-secure-settings.md
pkoutsovasilis: Applied in 181c123

deploy-manage/security/k8s-secure-settings.md
pkoutsovasilis: Applied in 181c123

deploy-manage/security/k8s-secure-settings.md
pkoutsovasilis: Changed in 534b51a

pkoutsovasilis:
thanks for the initial review @shainaraskas , please take another look

deploy-manage/security/k8s-secure-settings.md
```diff
@@ -13,11 +13,20 @@ products:
...
-The secrets should contain a key-value pair for each secure setting you want to add. ECK automatically injects these settings into the keystore on each {{es}} or {{kib}} Pod before it starts. The ECK operator continues to watch the secrets for changes and will update the {{es}} or {{kib}} keystores when it detects a change.
+The secrets should contain a key-value pair for each secure setting you want to add. ECK watches the referenced secrets for changes and delivers them to your {{es}} or {{kib}} Pods. By default, each update triggers a rolling restart of the affected Pods to repopulate the keystore.
 
-To allow the operator to inject the settings into the application, you must reference your secrets in the `spec.secureSettings` field of your {{es}} or {{kib}} object definition. Next, you’ll find examples for both {{es}} and {{kib}}.
+```{applies_to}
+deployment:
+  eck: ga 3.5+
+```
+
+For {{es}} 9.5 and later, you can opt in to [updating secure settings without a restart](#k8s-es-secure-settings-hot-reload).
```
shainaraskas: we can't float an applies to tag in the middle of a paragraph here. we have two options: inline all version info, or rely on the target link to carry the version limitations. I prefer the latter (they have to read the instructions to do the work anyway).


````suggestion
In ECK 3.5 and later, you can opt in to [updating secure settings without a restart](#k8s-es-secure-settings-hot-reload) for {{es}} 9.5+ clusters.
````

````suggestion
You also can opt in to [updating secure settings without a restart](#k8s-es-secure-settings-hot-reload).
````

we also technically support inline tags, but prefer those for bulleted lists (it gets messy if we just let them go anywhere). syntax looks like this

```
{applies_to}`eck: ga 3.5` For {{es}} 9.5 and later, you...
```

**Review by shainaraskas — APPROVED** [review]
thanks for the reorg 🙏 one small thing to address

File: deploy-manage/security/k8s-secure-settings.md (hunk 6 above)
Comment (pkoutsovasilis) [reply deploy-manage/security/k8s-secure-settings.md]: Ok makes sense, went with the latter in d009857

File: deploy-manage/security/k8s-secure-settings.md [hunk 7]
```diff
@@ -128,6 +132,40 @@ stringData:
     }
 ```
 
+### Update secure settings without a restart [k8s-es-secure-settings-hot-reload]
+```{applies_to}
+deployment:
+  eck: ga 3.5+
```
pebrc, line 138: Nit: `eck: ga 3.5+` is non-idiomatic. The repo convention is `ga X.Y` with no trailing `+` (e.g. `ga 9.1`, `ga 3.3`, `ga 9.5` across `deploy-manage/`), and the version already means "and later" in the `applies_to` model, so the `+` is redundant.

```suggestion
  eck: ga 3.5
```

deploy-manage/security/k8s-secure-settings.md
```diff
@@ -128,6 +132,40 @@ stringData:
...
+spec:
+  version: {{version.stack}}
+  nodeSets:
+  - name: default
+    count: 3
+  secureSettings:
+  - secretName: s3-credentials
+```
+
+When the annotation is set and the cluster is running {{es}} 9.5+, ECK delivers all `spec.secureSettings` entries through {{es}} file-based settings instead of the keystore init container. Updating a source secret no longer triggers a rolling restart. Instead, {{es}} applies the updated credentials in place on each node.
```
pebrc: Optional enhancement: this correctly states that source-secret *updates* no longer trigger a rolling restart. Worth noting that first enabling (or disabling) the annotation itself changes the Pod template — the keystore init container is removed — so the initial toggle does cause one rolling restart. Adding a sentence here would set the right expectation for a reader who assumes zero restarts.

pebrc (approved):

deploy-manage/security/k8s-secure-settings.md
shainaraskas: repo convention has been changed so it's easier to think about "current" values when it's paired with a ranged value representing older releases, e.g. eck 3.0-3.4, eck 3.5+ 

both are accepted - we don't enforce the `+` for backwards compatibility and because it's the "common" case

pkoutsovasilis: ok no objection changing it, two words before doing so. In my understanding 3.5+ means >=3.5.0. The next release that we are about to make is 3.5.0, so not sure what repo convention means in this context. Now I am seeing in other no ECK related files as well this being used like that. Maybe I am mistaken though, so before I press the apply suggestion, are we sure we wanna change this?

deploy-manage/security/k8s-secure-settings.md
pkoutsovasilis: Added in 5129e0ebd5d9fa67ab53b6dac8494808c9c9b9e2

deploy-manage/security/k8s-secure-settings.md
shainaraskas: it's something that didn't exist until somewhat recently, and we're terrible at enforcing it. in this case, I'd keep the `+` because this will be the functionality until the functionality changes again. however, you're right that it doesn't match the rest of the tags, because we haven't done any cleanup exercise to make the tags consistent (and don't intend to). this is more of an ergonomic/legibility benefit than anything else, so I'm happy either way.

pkoutsovasilis:
@shainaraskas @pebrc  since the upcoming release of ECK is nearby are we ok with this PR? 🙂

pkoutsovasilis:
merging as this will be part of another PR to merge in `main`

PR #7484, opened 2026-07-21
docs: rewrite SAML SSO guide and split into dedicated sub-pages

eedugon:
Closes elastic/docs-content-internal#427

Main preview link: https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7484/deploy-manage/users-roles/cluster-or-deployment-auth/saml

## Summary

The original `saml.md` had grown into a single monolithic page with 43 sections covering every aspect of SAML configuration — prerequisites, configuration steps, advanced features, reference details, and edge cases — with no clear separation between required and optional content, and no logical ordering that reflected an actual configuration workflow.

This PR restructures the guide into an actionable how-to and moves some parts to dedicated sub-pages.

## Scope of changes

**`saml.md` (main guide)** has been significantly reworked: restructured as a stepper-based how-to, language improved for clarity, redundancies removed, technical inaccuracies corrected, and advanced sections reordered by relevance.

**Sub-pages extracted from the original content** have been left with the original text largely intact. They are a starting point and will be improved in follow-up work:
- `saml-attribute-mapping.md` — full attribute mapping reference
- `saml-signing-encryption.md` — signing and encryption guide with certificate generation examples
- `saml-without-kibana.md` — SAML REST API flow for custom applications

## Structural changes to `saml.md`

- Core configuration wrapped in a stepper with 5 explicit steps: configure IdP → create ES realm → configure Kibana → configure role mappings → test authentication
- Advanced configuration section introduced with a framing paragraph so users understand upfront that its subsections are optional and situational
- Advanced sections reordered by relevance: signing and encryption first, then SLO, then authentication context, then SP metadata

## Notable content corrections

**Multiple Kibana instances** — Rewritten to make the actual condition explicit: each distinct public URL requires its own SAML realm, regardless of how many nodes or replicas exist behind it. Correctly scoped to self-managed and ECK deployments.

**`req_authn_context` section** — Now includes an MFA example using the Entra ID URI with a footnote clarifying that exact values are IdP-specific.

**SP metadata curl command** — Fixed from `http://localhost:9200` (incorrect, SAML requires HTTPS) to `https://elasticsearch.example.com:9200` with a footnote.

**SLO section** — Simplified by removing redundant sub-headings that duplicated content already in the IdP requirements section.


deploy-manage/users-roles/cluster-or-deployment-auth/saml-entra.md
```diff
@@ -89,7 +89,7 @@ Follow these steps to configure SAML with Microsoft Entra ID as an identity prov
         * For `idp.metadata.path`, we’ve shown the format to construct the URL. This value should be identical to the `App Federation Metadata URL` setting that you made a note of in the previous step.
 
         :::{admonition} For organizations with many group memberships
-        If you configure [`attributes.groups`](/deploy-manage/users-roles/cluster-or-deployment-auth/saml.md#saml-es-user-properties) to read the list of Azure AD groups from the SAML assertion, be aware that users who belong to many groups may exceed Azure AD’s size limit for SAML tokens. In that case, the `groups` attribute will be omitted.
+        If you configure [`attributes.groups`](/deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md#saml-es-user-properties) to read the list of Azure AD groups from the SAML assertion, be aware that users who belong to many groups may exceed Azure AD’s size limit for SAML tokens. In that case, the `groups` attribute will be omitted.
```
shainaraskas: little style edit while we're here
```suggestion
        If you configure [`attributes.groups`](/deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md#saml-es-user-properties) to read the list of Azure AD groups from the SAML assertion, users who belong to many groups might exceed Azure AD’s size limit for SAML tokens. In that case, the `groups` attribute is omitted.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
-
-To configure SAML, you need to perform the following steps:
-
-1. [Configure the prerequisites](#prerequisites)
-2. [Create one or more SAML realms](#saml-create-realm)
-3. [Configure role mappings](#saml-role-mapping)
-4. [Configure {{kib}} to use SAML as the authentication provider](#saml-configure-kibana)
-
-Additional steps outlined in this document are optional.
+Because this feature is designed with {{kib}} in mind, most sections of this guide assume {{kib}} is used. To learn how a custom web application could use the SAML REST APIs to authenticate users to {{es}} without {{kib}}, refer to [SAML without {{kib}}](#saml-no-kibana).
```
shainaraskas: suggest we just link out to the dedicated page now (annoying to click and then likely just want to click again)


```suggestion
Because this feature is designed with {{kib}} in mind, most sections of this guide assume {{kib}} is used. To learn how a custom web application could use the SAML REST APIs to authenticate users to {{es}} without {{kib}}, refer to [SAML without {{kib}}](/deploy-manage/users-roles/cluster-or-deployment-auth/saml-without-kibana.md).
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
+## How it works [saml-how-it-works]
 
-In SAML terminology, the {{stack}} is operating as a *Service Provider*.
+In SAML terminology, your identity system is the *Identity Provider* (IdP): it authenticates users and issues SAML assertions about them. The {{stack}} acts as a *Service Provider* (SP): {{kib}} initiates and coordinates the SSO flow, and {{es}} validates the SAML assertions and issues session tokens.
 
-The other component that is needed to enable SAML single-sign-on is the *Identity Provider*, which is a service that handles your credentials and performs that actual authentication of users.
+To enable SSO, register the {{stack}} as a known SP within your IdP, and configure {{es}} and {{kib}} to trust and communicate with your IdP. When SAML is enabled in {{kib}}, unauthenticated users are redirected to the IdP login page by default.
+ 
+The {{stack}} implements SAML SSO with the SAML 2.0 Web Browser SSO profile. Because the flow is browser-based, the SAML realm is not suitable for standard REST clients. If you configure SAML for {{kib}}, also add a realm for API access, such as the [native realm](/deploy-manage/users-roles/cluster-or-deployment-auth/native.md).
```
shainaraskas, line 51: gorgeous

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
-* If you want to support [Single Logout](/deploy-manage/users-roles/cluster-or-deployment-auth/saml.md#saml-logout), a `<SingleLogoutService>` with binding of HTTP-Redirect (`urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect`)
-
-### Signing requirements
-
-The {{stack}} requires that all messages from the IdP are signed. For authentication `<Response>` messages, the signature can be applied to either the response itself, or to the individual assertions. For `<LogoutRequest>` messages, the message itself must be signed, and the signature should be provided as a URL parameter, as required by the `HTTP-Redirect` binding.
-
-## Prerequisites
-
-Before you set up SAML single-sign on, you must have a [SAML IdP](#saml-guide-idp) configured.
+* If you want [Single Logout](#saml-logout): a `<SingleLogoutService>` with binding of HTTP-Redirect
```
shainaraskas: while we spell it out before, I think it is nice for the bullets to stand alone and i'd keep the urn here too

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
 
-If you're using a self-managed cluster, then perform the following additional steps:
+All messages from the IdP must be signed. For `<Response>` messages, the signature can be on the response itself or on individual assertions. For `<LogoutRequest>` messages, the signature must be provided as a URL parameter, as required by the `HTTP-Redirect` binding.
 
-* Enable TLS for HTTP.
+### Prerequisites for self-managed clusters
```
shainaraskas, line 75: can we put an applies_to tag here so people are clear on if this is vanilla self-managed or ALLLL self-managed options? I know the note below specified what it doesn't apply to, but nice to be clear

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
 
-    If your {{es}} cluster is operating in production mode, you must configure the HTTP interface to use SSL/TLS before you can enable SAML authentication. For more information, see [Encrypt HTTP client communications for {{es}}](/deploy-manage/security/set-up-basic-security-plus-https.md#encrypt-http-communication).
```
shainaraskas, line 98: does this only apply to prod mode

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
+If you're using a self-managed cluster:
 
-    If you started {{es}} [with security enabled](/deploy-manage/deploy/self-managed/installing-elasticsearch.md), then TLS is already enabled for HTTP.
+* Enable TLS for the {{es}} HTTP interface. SAML requires HTTPS. For more information, see [Encrypt HTTP client communications for {{es}}](/deploy-manage/security/set-up-basic-security-plus-https.md#encrypt-http-communication). If you started {{es}} [with security enabled](/deploy-manage/security/self-setup.md), TLS is already enabled.
```
shainaraskas: can we swap the CTA and caveat here so people don't click through unless they need to? 

we might loosen this up to say "in most cases", because we start clusters w/ security enabled for all stack versions supported by this docset. enabling security is not an intentional choice now so this framing might be confusing



```suggestion
* Enable TLS for the {{es}} HTTP interface. SAML requires HTTPS. [In most installation scenarios](/deploy-manage/security/self-setup.md), TLS is already enabled. For manual setup steps, refer to [Encrypt HTTP client communications for {{es}}](/deploy-manage/security/set-up-basic-security-plus-https.md#encrypt-http-communication).
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
+  xpack.security.authc.token.enabled: true
+  ```
 
-    {{ech}}, {{ece}}, and {{eck}} have TLS enabled by default.
+:::{note}
+{{ech}}, {{ece}}, and {{eck}} enable HTTPS and the token service by default.
+:::
 
-* Enable the token service.
+## Configuration steps
```
shainaraskas, line 94: prefer a sentence between headings/before numbered lists

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
+:::::{stepper}
 
-    ```yaml
-    xpack.security.authc.token.enabled: true
-    ```
+::::{step} Configure your identity provider
+:anchor: saml-configure-idp
 
-    {{ech}}, {{ece}}, and {{eck}} have TLS enabled by default.
+Register the {{stack}} as a SAML service provider in your IdP. The exact steps vary by provider, but you generally need to:
```
shainaraskas: sentences where the first clause is before a colon and the "second" clause is a list after a colon is not great for translation/some ESL readers, so we can "complete" the sentence

```suggestion
Register the {{stack}} as a SAML service provider in your IdP. The exact steps vary by provider, but you generally need to do the following:
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
+3. Set the **entity ID** to a URI that uniquely identifies this {{kib}} instance as a Service Provider. We recommend using the {{kib}} base URL (for example, `https://kibana.example.com`).
+4. Set the **logout URL** to your {{kib}} base URL followed by `/logout` if your IdP supports Single Logout (for example, `https://kibana.example.com/logout`).
+5. Identify the user attributes your IdP can include in SAML assertions. Consult your IdP documentation or local admin. This varies between providers. These attribute URIs will be used to configure the attribute mapping in the {{es}} realm.
+6. Note the **IdP metadata URL** or download the metadata file. You will need this for the {{es}} configuration.
 
-SAML authentication is enabled by configuring a SAML realm within the authentication chain for {{es}}.
+:::{tip}
+If your IdP supports SP metadata import, you can generate an SP metadata file after [configuring the {{es}} realm](#saml-create-realm) and import it into your IdP to register the {{stack}} as a Service Provider automatically. Refer to [Generate SP metadata](#saml-sp-metadata).
```
shainaraskas: shall we frontload this so people don't follow this process if they want to generate the SP metadata?

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
+:::
 
-This realm has a few mandatory settings, and a number of optional settings. The available settings are described in detail in [Security settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md):
-* [SAML realm settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#ref-saml-settings)
-* [SAML realm signing settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#ref-saml-signing-settings)
-* [SAML realm encryption settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#ref-saml-encryption-settings)
-* [SAML realm SSL settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#ref-saml-ssl-settings)
+:::{note}
+If your IdP requires signed outgoing SAML messages (authentication requests or logout requests), you will also need to provide your {{es}} signing certificate to the IdP at this stage. Refer to [Signing and encryption](#saml-enc-sign) for how to generate certificates and configure signing in {{es}}.
```
shainaraskas: if you move the tip out, you could feasibly take this out of a note and just make it a para after the core steps

or, maybe better, make it step 7: 

`7. Optional: ...`

```suggestion
If your IdP requires signed outgoing SAML messages (authentication requests or logout requests), then you also need to provide your {{es}} signing certificate to the IdP at this stage. Refer to [Signing and encryption](#saml-enc-sign) for how to generate certificates and configure signing in {{es}}.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
     * If you're using {{eck}}, then install the file as [custom configuration files](/deploy-manage/deploy/cloud-on-k8s/custom-configuration-files-plugins.md#use-a-volume-and-volume-mount-together-with-a-configmap-or-secret).
```
shainaraskas:
```suggestion
    * If you're using {{eck}}, then install the file as a [custom configuration file](/deploy-manage/deploy/cloud-on-k8s/custom-configuration-files-plugins.md#use-a-volume-and-volume-mount-together-with-a-configmap-or-secret).
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
 
-::::{dropdown} Common settings
-
-xpack.security.authc.realms.saml.saml1
-:   Defines a new `saml` authentication realm named "saml1". See [Realms](/deploy-manage/users-roles/cluster-or-deployment-auth/authentication-realms.md) for more explanation of realms.
+1. Controls realm priority. Assign SSO realms higher order values than password-based realms (native, LDAP). If you're using {{eck}}, set `order` to a value greater than the file realm (default `-100`) and native realm (default `-99`).
 
-order
-:   The order of the realm within the realm chain. Realms with a lower order have highest priority and are consulted first. We recommend giving password-based realms such as file, native, LDAP, and Active Directory the lowest order (highest priority), followed by SSO realms such as SAML and OpenID Connect. If you have multiple realms of the same type, give the most frequently accessed realm the lowest order to have it consulted first.
+Configure each setting as follows. For the full list of available settings, refer to [SAML realm settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#ref-saml-settings).
```
shainaraskas: can we contextualize these as required / core / commonly used settings so the link to the larger reference makes more sense? or, if all of the settings are here, "for more details about these settings ... "

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
...
 
-These attributes might include information like:
+`sp.entity_id`
+:   A unique identifier for your {{kib}} instance, expressed as a URI. Must match exactly the entity ID you [configure in your IdP](#saml-configure-idp). The comparison is case-sensitive. We recommend using the {{kib}} base URL.
 
-* The user’s username
-* The user’s email address
-* The user’s groups or roles
+`sp.acs`
+:   The URL of the Assertion Consumer Service within {{kib}} that receives authentication responses from your IdP, using the HTTP-POST binding. For example, `https://kibana.example.com/api/security/saml/callback`. This URL must be reachable from users' browsers — it does not need to be directly accessible by {{es}} or the IdP. If {{kib}} is behind a reverse proxy, use the public-facing URL.
```
shainaraskas:
```suggestion
:   The URL of the Assertion Consumer Service within {{kib}} that receives authentication responses from your IdP, using the HTTP-POST binding. For example, `https://kibana.example.com/api/security/saml/callback`. This URL must be reachable from users' browsers, but does not need to be directly accessible by {{es}} or the IdP. If {{kib}} is behind a reverse proxy, use the public-facing URL.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
...
 
-Attributes in SAML are [usually](#saml-attribute-mapping-nameid) named using a URI such as `urn:oid:0.9.2342.19200300.100.1.1` or `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/upn`, and have one or more values associated with them.
+`sp.logout`
+:   The URL of the Single Logout service within {{kib}} that receives logout messages from your IdP. For example, `https://kibana.example.com/logout`. Required for [SAML Single Logout](#saml-logout). If not configured, {{es}} refuses all `<LogoutRequest>` messages from the IdP.
 
-These attribute identifiers vary between IdPs, and most IdPs offer ways to customize the URIs and their associated value.
+`attributes.principal` (required)
```
shainaraskas, line 159: why do we have `required` and `recommended` on only two settings here?

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
...
+The attribute URIs must exactly match what your IdP sends in the assertion. A mismatch silently breaks authentication: the login might appear to succeed but the user's principal or group memberships will be missing or incorrect. Verify the exact URIs with your IdP administrator or by inspecting a captured SAML assertion.
+:::
 
-In order for SAML attributes to be useful in {{es}}, {{es}} and the IdP need to have a common value for the names of the attributes. This is done manually, by configuring the IdP and the SAML realm to use the same URI name for each logical user attribute.
+The most important mappings are `attributes.principal` and `attributes.groups`: `attributes.principal` is required and determines the username {{es}} assigns to the authenticated user, and `attributes.groups` is recommended if you want to assign roles based on IdP group memberships.
 
-The recommended steps for configuring these SAML attributes are as follows:
+For a complete reference of mappable user properties, special attribute names, and advanced patterns, refer to [Map SAML attributes to {{es}} user properties](/deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md).
```
shainaraskas, line 183: we might also say this is a detailed guide to set this up (it carries the best practices / approaches we used to store in this doc)

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
...
+If you run multiple {{kib}} instances connected to the same cluster or deployment (for example, behind a load balancer), apply this SAML provider configuration to every instance. If each instance is accessed through a different URL, each needs its own SAML realm. Refer to [Multiple {{kib}} instances and URLs](#_operating_multiple_kib_instances).
 :::
 
-*friendlyName*
-:   A SAML attribute may have a *friendlyName* in addition to its URI based name. For example the attribute with a name of `urn:oid:0.9.2342.19200300.100.1.1` might also have a friendlyName of `uid`. You may use these friendly names within an attribute mapping, but it is recommended that you use the URI based names, as friendlyNames are neither standardized or mandatory.
+#### Allow both SAML and basic authentication [saml-kibana-basic]
 
-The example below configures a realm to use a persistent nameid for the principal, and the attribute with the friendlyName "roles" for the user’s groups.
+To let some users (for example, local administrators) log in with a username and password, enable a basic provider alongside SAML:
```
shainaraskas: there are _a lot_ of parentheticals in this doc. was trying to ignore them but will state once my preference to avoid them as a rule

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
...
+xpack.security.authc.providers:
+  saml.saml1:
+    order: 0
+    realm: "saml1"
+    description: "Log in with SSO"
+  basic.basic1:
+    order: 1
 ```
 
+Users will see a login selector and can choose their preferred method. Users who log in with basic authentication must have credentials in a configured {{es}} realm (such as native or LDAP).
```
shainaraskas:
```suggestion
When multiple authentication providers are confgired, users see a login selector and can choose their preferred method. Users who log in with basic authentication must have credentials in a configured {{es}} realm such as native or LDAP.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
...
 
-### Mappable {{es}} user properties [saml-es-user-properties]
+You can also adjust [session timeout settings](/deploy-manage/security/kibana-session-management.md) (`xpack.security.session.idleTimeout` and `xpack.security.session.lifespan`) to match your security requirements.
```
shainaraskas: this sentence feels a little throwaway here. is it part of the multiple auth providers case? perhaps needs to move up

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
...
+::::{step} Configure role mappings
+:anchor: saml-role-mapping
 
-mail
-:   *(Optional)* The user’s email address. It will be used in {{kib}}'s profile page to display user details.
+SAML authentication identifies users to the {{stack}}, but does not automatically grant them any access. You must map SAML users to {{es}} roles before they can do anything.
 
-dn
-:   *(Optional)* The user’s X.500 *Distinguished Name*.
+Role mappings connect SAML user identities to {{es}} roles, but the roles themselves must exist first. You can use built-in roles such as `superuser` or `kibana_admin` for initial testing. For production, create [custom roles](/deploy-manage/users-roles/cluster-or-deployment-auth/defining-roles.md) with appropriate access to your data and [{{kib}} features](/deploy-manage/users-roles/cluster-or-deployment-auth/kibana-privileges.md#kibana-feature-privileges).
```
shainaraskas: > appropriate access to your data

this is more than data right? it's also es features

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
...
 
+You can create role mappings in the **Role Mappings** page in {{kib}} or with the [role mapping API]({{es-apis}}operation/operation-security-put-role-mapping). For the general concepts, UI workflow, and rule syntax, refer to [Map external users and groups to roles](/deploy-manage/users-roles/cluster-or-deployment-auth/mapping-users-groups-to-roles.md). The examples below show common SAML patterns.
```
shainaraskas:
```suggestion
You can create role mappings in the **Role Mappings** page in {{kib}} or with the [role mapping API]({{es-apis}}operation/operation-security-put-role-mapping). For the general concepts, UI workflow, and rule syntax, refer to [Map external users and groups to roles](/deploy-manage/users-roles/cluster-or-deployment-auth/mapping-users-groups-to-roles.md). The following examples show common SAML patterns.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
...
-```
+For a complete example including certificate generation and configuration for PEM, PKCS#12, and JKS formats, refer to [Configure SAML signing and encryption](/deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md).
 
-1. This configuration ensures that only SAML authentication requests will be sent signed to the Identity Provider.
+For available settings, refer to [SAML realm signing settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#ref-saml-signing-settings) and [SAML realm encryption settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#ref-saml-encryption-settings).
 
+### SAML Single Logout [saml-logout]
 
-### Configuring {{es}} for encrypted messages [_configuring_es_for_encrypted_messages]
+The SAML protocol supports Single Logout (SLO), which ends both the {{kib}} session and the IdP session when a user logs out. Support for SLO varies between identity providers. Consult your IdP's documentation to determine what Logout services it offers.
```
shainaraskas:
```suggestion
The SAML protocol supports Single Logout (SLO), which ends both the {{kib}} session and the IdP session when a user logs out. Support for SLO varies between identity providers. Consult your IdP's documentation to determine what logout services it offers.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
...
+    - "http://schemas.microsoft.com/claims/multipleauthn"  # <1>
+```
 
-`encryption.keystore.secure_password`
-:   The passphrase for the keystore, if the file is encrypted. This is a secure setting that must be uploaded to your [{{es}} keystore](/deploy-manage/security/secure-settings.md).
+1. This example uses the Entra ID URI for MFA. The exact URI values depend on your IdP. Some standard SAML 2.0 values such as `urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport` are widely supported, but IdP-specific URIs are common. Consult your IdP's documentation.
 
-:::
-::::
+{{es}} supports only the `exact` comparison method. The Authentication Response must include one of the specified values; otherwise, authentication fails. For more details, refer to [`req_authn_context_class_ref`](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#ref-saml-settings).
```
shainaraskas: nit: link lands a little far from the target. might want to contextualize the link a little more or inject an anchor into the target page

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -159,388 +144,253 @@ idp.metadata.path
...
 
-## Generate SAML metadata for the Service Provider [saml-sp-metadata]
+### Generate SP metadata [saml-sp-metadata]
 
-Some Identity Providers support importing a metadata file from the Service Provider. This will automatically configure many of the integration options between the IdP and the SP.
+The SP metadata is an XML document that describes your {{stack}} as a SAML Service Provider. It contains your entity ID, ACS URL, logout URL, and any signing or encryption certificates configured in the realm. Some Identity Providers can import this file to configure the integration automatically, instead of requiring you to enter each value individually when [registering the {{stack}} in your IdP](#saml-configure-idp).
```
shainaraskas:
```suggestion
SP metadata is an XML document that describes your {{stack}} as a SAML Service Provider. It contains your entity ID, ACS URL, logout URL, and any signing or encryption certificates configured in the realm. Some Identity Providers can import this file to configure the integration automatically, instead of requiring you to enter each value individually when [registering the {{stack}} in your IdP](#saml-configure-idp).
```

alt:

```suggestion
The SP metadata file is an XML document that describes your {{stack}} as a SAML Service Provider. It contains your entity ID, ACS URL, logout URL, and any signing or encryption certificates configured in the realm. Some Identity Providers can import this file to configure the integration automatically, instead of requiring you to enter each value individually when [registering the {{stack}} in your IdP](#saml-configure-idp).
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -568,152 +418,25 @@ kubectl cp elasticsearch-sample-es-default-0:/usr/share/elasticsearch/saml-elast
...
-### Operating multiple {{kib}} instances [_operating_multiple_kib_instances]
-
-If you want to have multiple {{kib}} instances that authenticate against the same {{es}} cluster, then each {{kib}} instance that is configured for SAML authentication, requires its own SAML realm.
-
-Each SAML realm must have its own unique Entity ID (`sp.entity_id`), and its own *Assertion Consumer Service* (`sp.acs`). Each {{kib}} instance will be mapped to the correct realm by looking up the matching `sp.acs` value.
-
-These realms may use the same Identity Provider, but are not required to.
-
-The following is example of having 3 different {{kib}} instances, 2 of which use the same internal IdP, and another which uses a different IdP.
+The following is an example of having 3 different {{kib}} instances, 2 of which use the same internal IdP, and another which uses a different IdP.
```
shainaraskas:
```suggestion
The following is an example of settings for a cluster with three different {{kib}} instances, two of which use the same internal IdP, and another which uses a different IdP.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -731,126 +454,20 @@ xpack.security.authc.realms.saml.saml_eng:
...
+:::
 
-For SP-initiated Single Logout, the IdP may send back a logout response which can be verified by {{es}} using the [SAML complete logout API]({{es-apis}}operation/operation-security-saml-complete-logout).
+### SAML without {{kib}} [saml-no-kibana]
 
+If you're building a custom web application that needs to authenticate users against {{es}} using SAML (without {{kib}}), you can use the SAML REST APIs to implement the full authentication flow directly. This is relevant when you have a custom portal, a microservice-based architecture, or any scenario where the browser-based SSO flow must be handled by your own application code rather than by {{kib}}.
 
+The application acts as an authentication proxy: it drives the SP-initiated or IdP-initiated SSO flow, exchanges the SAML response for an {{es}} access token and refresh token, and uses that token as a `Bearer` credential for subsequent requests.
 
+For the complete implementation guide (including service account setup, SP-initiated and IdP-initiated flows, and logout handling), refer to [SAML without {{kib}}](/deploy-manage/users-roles/cluster-or-deployment-auth/saml-without-kibana.md).
```
shainaraskas:
```suggestion
For the complete implementation guide, including service account setup, SP-initiated and IdP-initiated flows, and logout handling, refer to [SAML without {{kib}}](/deploy-manage/users-roles/cluster-or-deployment-auth/saml-without-kibana.md).
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+description: Understand how SAML attributes map to Elasticsearch user properties and metadata, and how to configure attribute mappings in your SAML realm.
+applies_to:
+  stack: all
+products:
+  - id: elasticsearch
+---
+
+# Map SAML attributes to {{es}} user properties [saml-attributes-mapping]
+
+When a user connects to {{kib}} through your Identity Provider, the Identity Provider will supply a SAML Assertion about the user. The assertion will contain an *Authentication Statement* indicating that the user has successfully authenticated to the IdP and one or more *Attribute Statements* that will include *Attributes* for the user.
```
shainaraskas:
```suggestion
When a user connects to {{kib}} through your Identity Provider, the Identity Provider supplies a SAML assertion about the user. The assertion contains an *Authentication Statement* indicating that the user has successfully authenticated to the IdP and one or more *Attribute Statements* that include *Attributes* for the user.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+These attributes might include information like:
+
+* The user's username
+* The user's email address
+* The user's groups or roles
+
+Attributes in SAML are [usually](#saml-attribute-mapping-nameid) named using a URI such as `urn:oid:0.9.2342.19200300.100.1.1` or `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/upn`, and have one or more values associated with them.
+
+These attribute identifiers vary between IdPs, and most IdPs offer ways to customize the URIs and their associated value.
```
shainaraskas: maybe?
```suggestion
These attribute identifiers vary between IdPs, and most IdPs offer ways to customize the URIs and their associated values.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+{{es}} does not require specific attribute URIs, but the values configured in the SAML realm must exactly match the identifiers sent by the IdP.
+
+{{es}} uses these attributes to infer information about the user who has logged in, and they can be used for [role mapping](/deploy-manage/users-roles/cluster-or-deployment-auth/saml.md#saml-role-mapping).
+
+## How attributes appear in user metadata [saml-user-metadata]
+
+By default users who authenticate through SAML will have some additional metadata fields.
```
shainaraskas:
```suggestion
By default, users who authenticate through SAML have some additional metadata fields.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+* `saml_nameid` will be set to the value of the `NameID` element in the SAML authentication response
+* `saml_nameid_format` will be set to the full URI of the NameID's `format` attribute
+* Every SAML Attribute that is provided in the authentication response (regardless of whether it is mapped to an {{es}} user property), will be added as the metadata field `saml(name)` where "name" is the full URI name of the attribute. For example `saml(urn:oid:0.9.2342.19200300.100.1.3)`.
+* For every SAML Attribute that has a *friendlyName*, it will also be added as the metadata field `saml_friendlyName` where "name" is the friendly name of the attribute. For example `saml_mail`.
```
shainaraskas: avoid future (assume this all predated your work)


```suggestion
* `saml_nameid` is set to the value of the `NameID` element in the SAML authentication response
* `saml_nameid_format` is set to the full URI of the NameID's `format` attribute
* Every SAML Attribute that is provided in the authentication response, regardless of whether it is mapped to an {{es}} user property, is added as the metadata field `saml(name)` where "name" is the full URI name of the attribute. For example, `saml(urn:oid:0.9.2342.19200300.100.1.3)`.
* For every SAML attribute that has a *friendlyName*, it is also added as the metadata field `saml_friendlyName` where "name" is the friendly name of the attribute. For example, `saml_mail`.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+This behavior can be disabled by adding `populate_user_metadata: false` as a setting in the SAML realm.
+
+## Map attributes
+
+In order for SAML attributes to be useful in {{es}}, {{es}} and the IdP need to have a common value for the names of the attributes. This is done manually, by configuring the IdP and the SAML realm to use the same URI name for each logical user attribute.
```
shainaraskas:
```suggestion
For SAML attributes to be useful in {{es}}, {{es}} and the IdP need to have a common value for the names of the attributes. This is done manually, by configuring the IdP and the SAML realm to use the same URI name for each logical user attribute.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+1. Consult your IdP to see what user attributes it can provide. This varies greatly between providers, but you should be able to obtain a list from the documentation, or from your local admin.
+2. Review the list of [user properties](#saml-es-user-properties) that {{es}} supports, and decide which of them are useful to you, and can be provided by your IdP. At a *minimum*, you must configure a mapping for the required `principal` user property.
+3. Configure your IdP to "release" those attributes to your {{kib}} SAML service provider. This process varies by provider: some will provide a user interface for this, while others might require that you edit configuration files.
+
+   Because {{es}} does not require that any specific URIs are used, you can use any URIs as recommended by the IdP or your local administrator.
+4. Configure the SAML realm in {{es}} to associate the [{{es}} user properties](#saml-es-user-properties) to the URIs that you configured in your IdP. Refer to [Create a SAML realm](/deploy-manage/users-roles/cluster-or-deployment-auth/saml.md#saml-create-realm) for a sample configuration with `principal` and `groups`.
```
shainaraskas, line 52: hmmm wonder if we should make this example a snippet 🤔

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Special attribute names [saml-attribute-mapping-nameid]
+
+In general, {{es}} expects that the configured value for an attribute will be a URI, such as `urn:oid:0.9.2342.19200300.100.1.1`. However, there are some additional names that can be used:
```
shainaraskas:
```suggestion
In general, {{es}} expects that the configured value for an attribute is a URI, such as `urn:oid:0.9.2342.19200300.100.1.1`. However, there are some additional names that can be used:
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+`nameid`
+:   This uses the SAML `NameID` value (all leading and trailing whitespace removed) instead of a SAML attribute. SAML `NameID` elements are an optional, but frequently provided, field within a SAML Assertion that the IdP can use to identify the Subject of that Assertion. Sometimes the `NameID` will relate to the user's login identifier (username) within the IdP, but often they will be internally generated identifiers that have no obvious meaning outside of the IdP.
+
+`nameid:persistent`
+:   This uses the SAML `NameID` value (all leading and trailing whitespace removed), but only if the NameID format is `urn:oasis:names:tc:SAML:2.0:nameid-format:persistent`. A SAML `NameID` element has an optional `Format` attribute that indicates the semantics of the provided name. It is common for IdPs to be configured with "transient" NameIDs that present a new identifier for each session. Since it is rarely useful to use a transient NameID as part of an attribute mapping, the `nameid:persistent` attribute name can be used as a safety mechanism that will cause an error if you attempt to map from a `NameID` that does not have a persistent value.
```
shainaraskas:
```suggestion
:   This uses the SAML `NameID` value (all leading and trailing whitespace removed), but only if the NameID format is `urn:oasis:names:tc:SAML:2.0:nameid-format:persistent`. A SAML `NameID` element has an optional `Format` attribute that indicates the semantics of the provided name. It is common for IdPs to be configured with "transient" NameIDs that present a new identifier for each session. Because it is rarely useful to use a transient NameID as part of an attribute mapping, the `nameid:persistent` attribute name can be used as a safety mechanism that will cause an error if you attempt to map from a `NameID` that does not have a persistent value.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+:::{note}
+Identity Providers can be either statically configured to release a `NameID` with a specific format, or they can be configured to try to conform with the requirements of the SP. The SP declares its requirements as part of the Authentication Request, using an element which is called the `NameIDPolicy`. If this is needed, you can set the `nameid_format` setting in the SAML realm to request that the IdP releases a `NameID` with a specific format.
+:::
+
+*friendlyName*
+:   A SAML attribute can have a *friendlyName* in addition to its URI based name. For example the attribute with a name of `urn:oid:0.9.2342.19200300.100.1.1` might also have a friendlyName of `uid`. You can use these friendly names within an attribute mapping, but it is recommended that you use the URI based names, as friendlyNames are neither standardized nor mandatory.
```
shainaraskas:
```suggestion
:   A SAML attribute can have a *friendlyName* in addition to its URI based name. For example, the attribute with a name of `urn:oid:0.9.2342.19200300.100.1.1` might also have a friendlyName of `uid`. You can use these friendly names within an attribute mapping, but it is recommended that you use the URI based names, as friendlyNames are neither standardized nor mandatory.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+`nameid:persistent`
+:   This uses the SAML `NameID` value (all leading and trailing whitespace removed), but only if the NameID format is `urn:oasis:names:tc:SAML:2.0:nameid-format:persistent`. A SAML `NameID` element has an optional `Format` attribute that indicates the semantics of the provided name. It is common for IdPs to be configured with "transient" NameIDs that present a new identifier for each session. Since it is rarely useful to use a transient NameID as part of an attribute mapping, the `nameid:persistent` attribute name can be used as a safety mechanism that will cause an error if you attempt to map from a `NameID` that does not have a persistent value.
+
+:::{note}
+Identity Providers can be either statically configured to release a `NameID` with a specific format, or they can be configured to try to conform with the requirements of the SP. The SP declares its requirements as part of the Authentication Request, using an element which is called the `NameIDPolicy`. If this is needed, you can set the `nameid_format` setting in the SAML realm to request that the IdP releases a `NameID` with a specific format.
```
shainaraskas: a link to the nameid_format docs in this context might be helpful

alt, overall, a link to the ref being like "for detailed information about these settings, refer to __" might be useful here because down the line it will have type info and info on whether a setting is static or dynamic etc

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+`mail`
+:   *(Optional)* The user's email address. It will be used in {{kib}}'s profile page to display user details.
+
+`dn`
+:   *(Optional)* The user's X.500 *Distinguished Name*.
+
+## Extract partial values from SAML attributes [saml-attribute-patterns]
+
+There are some occasions where the IdP's attribute can contain more information than you want to use within {{es}}. A common example of this is one where the IdP works exclusively with email addresses, but you want the user's `principal` to use the `local-name` part of the email address. For example if their email address was `james.wong@staff.example.com`, then you might want their principal to be `james.wong`.
```
shainaraskas:
```suggestion
An IdP's attribute might contain more information than you want to use within {{es}}. A common example of this is one where the IdP works exclusively with email addresses, but you want the user's `principal` to use the `local-name` part of the email address. For example, if user's email address is `james.wong@staff.example.com`, then you might want their principal to be `james.wong`.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+This can be achieved using the `attribute_patterns` setting in the {{es}} realm, as demonstrated in the realm configuration below:
```
shainaraskas:
```suggestion
You can extract a partial value from an attribute using the `attribute_patterns` setting in the {{es}} realm, as demonstrated in the realm configuration below:
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+  order: 2
+  idp.metadata.path: saml/idp-metadata.xml
+  idp.entity_id: "https://sso.example.com/"
+  sp.entity_id:  "https://kibana.example.com/"
+  sp.acs: "https://kibana.example.com/api/security/saml/callback"
+  attributes.principal: "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress"
+  attribute_patterns.principal: "^([^@]+)@staff\\.example\\.com$"
+```
+
+In this case, the user's `principal` is mapped from an email attribute, but a regular expression is applied to the value before it is assigned to the user. If the regular expression matches, then the result of the first group is used as effective value. If the regular expression does not match then the attribute mapping fails.
```
shainaraskas:
```suggestion
In this example, the user's `principal` is mapped from an email attribute, but a regular expression is applied to the value before it is assigned to the user. If the regular expression matches, then the result of the first group is used as the effective value. If the regular expression does not match, then the attribute mapping fails.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
...
+
+In this example, the email address must belong to the `staff.example.com` domain, and then the local-part (anything before the `@`) is used as the principal. Any users who try to login using a different email domain will fail because the regular expression will not match against their email address, so their principal attribute — which is mandatory — will not be populated.
```
shainaraskas:
```suggestion
Using this expression, the email address must belong to the `staff.example.com` domain, and then the local-part (anything before the `@`) is used as the principal. Any users who try to log in using a different email domain will fail because the regular expression will not match against their email address, so their principal attribute — which is mandatory — will not be populated.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
```diff
@@ -0,0 +1,129 @@
+---
+navigation_title: SAML attribute mapping
+description: Understand how SAML attributes map to Elasticsearch user properties and metadata, and how to configure attribute mappings in your SAML realm.
+applies_to:
+  stack: all
+products:
+  - id: elasticsearch
+---
+
+# Map SAML attributes to {{es}} user properties [saml-attributes-mapping]
```
shainaraskas, line 10: can we contextualize this page in the context of the larger saml story? "For basic SAML realm configuration, refer to [main page]

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
shainaraskas, line 1: VERY NIT: it is a bit of an anti-pattern for an overview page to be a full-blown tutorial page like this. I think, ideally, we'd see this be an overview of the basic concepts and the child pages (including a core tutorial), and then link down to the tutorial

this structure is _much nicer now_, but it doesn't really follow an ideal parent-child pattern. you could argue that this is ok because the other procedures are sub- or dependent procedures, but someone might want them individually 🤷

deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md
```diff
@@ -0,0 +1,175 @@
...
+
+# Configure SAML signing and encryption [saml-enc-sign]
+
+The {{stack}} supports generating signed SAML messages (for authentication or logout), verifying signed SAML messages from the IdP (for both authentication and logout) and can process encrypted content.
+
+You can configure {{es}} for signing, encryption or both, using a single key or individual keys.
+
+The {{stack}} uses X.509 certificates with RSA private keys for SAML cryptography. These keys can be generated using any standard SSL tool, including the `elasticsearch-certutil` tool.
+
+Your IdP might require that the {{stack}} have a cryptographic key for signing SAML messages, and that you provide the corresponding signing certificate within the Service Provider configuration (either within the {{stack}} SAML metadata file, or manually configured within the IdP administration interface).
```
shainaraskas: many parentheses here

deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md
```diff
@@ -0,0 +1,175 @@
...
+
+While most IdPs do not expect authentication requests to be signed, it is commonly the case that signatures are required for logout requests. Your IdP will validate these signatures against the signing certificate that has been configured for the {{stack}} Service Provider.
```
shainaraskas:
```suggestion
While most IdPs do not expect authentication requests to be signed, signatures are often required for logout requests. Your IdP validates these signatures against the signing certificate that is configured for the {{stack}} Service Provider.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md
```diff
@@ -0,0 +1,175 @@
...
+
+Encryption certificates are rarely needed, but the {{stack}} supports them for cases where IdPs or local policies mandate their use.
+
+## Generate certificates and keys [saml-sign-generate-certs]
+
+{{es}} supports certificates and keys in either PEM, PKCS#12 or JKS format. Some Identity Providers are more restrictive in the formats they support, and will require you to provide the certificates as a file in a particular format. You should consult the documentation for your IdP to determine what formats they support.
```
shainaraskas:
```suggestion
{{es}} supports certificates and keys in either PEM, PKCS#12 or JKS format. Some Identity Providers are more restrictive in the formats they support, and require you to provide the certificates as a file in a particular format. You should consult the documentation for your IdP to determine what formats they support.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md
```diff
@@ -0,0 +1,175 @@
...
+  self:
+```
+
+Using the [`elasticsearch-certutil` tool](elasticsearch://reference/elasticsearch/command-line-tools/certutil.md), you can generate a signing certificate with the following command. Because PEM format is the most commonly supported format, the example generates a certificate in that format.
+
+```sh
+bin/elasticsearch-certutil cert --self-signed --pem --days 1100 --name saml-sign --out saml-sign.zip
+```
+
+This will do the following:
```
shainaraskas:
```suggestion
This command does the following:
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md
```diff
@@ -0,0 +1,175 @@
...
+
+* generate a certificate and key pair (the `cert` subcommand)
+* create the files in PEM format (`--pem` option)
+* generate a certificate that is valid for 3 years (`--days 1100`)
+* name the certificate `saml-sign` (`--name` option)
+* save the certificate and key in the `saml-sign.zip` file (`--out` option)
+
+The generated zip archive will contain 3 files:
```
shainaraskas:
```suggestion
The generated zip archive contains three files:
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md
```diff
@@ -0,0 +1,175 @@
...
+
+* `saml-sign.crt`, the public certificate to be used for signing
+* `saml-sign.key`, the private key for the certificate
+* `ca.crt`, a CA certificate that is not needed, and can be ignored.
+
+Encryption certificates can be generated with the same process.
+
+## Sign outgoing SAML messages [saml-sign-configure]
+
+By default, {{es}} will sign *all* outgoing SAML messages if a signing certificate and key has been configured.
```
shainaraskas: will stop commenting on `will`

```suggestion
By default, {{es}} signs *all* outgoing SAML messages if a signing certificate and key has been configured.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md
```diff
@@ -0,0 +1,175 @@
...
+          ...
+          signing.saml_messages: AuthnRequest # <1>
+          ...
+```
+
+1. This configuration ensures that only SAML authentication requests will be sent signed to the Identity Provider.
+
+## Configure {{es}} for encrypted messages [saml-encryption-configure]
+
+The {{es}} {{security-features}} support a single key for message decryption. If a key is configured, then {{es}} attempts to use it to decrypt `EncryptedAssertion` and `EncryptedAttribute` elements in Authentication responses, and `EncryptedID` elements in Logout requests.
```
shainaraskas:
```suggestion
{{es}} supports a single key for message decryption. If a key is configured, then {{es}} attempts to use it to decrypt `EncryptedAssertion` and `EncryptedAttribute` elements in Authentication responses, and `EncryptedID` elements in Logout requests.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md
```diff
@@ -0,0 +1,175 @@
...
+
+{{es}} rejects any SAML message that contains an `EncryptedAssertion` that cannot be decrypted.
+
+If an `Assertion` contains both encrypted and plain-text attributes, then failure to decrypt the encrypted attributes will not cause an automatic rejection. Rather, {{es}} processes the available plain-text attributes (and any `EncryptedAttributes` that could be decrypted).
```
shainaraskas:
```suggestion
If an `Assertion` contains both encrypted and plain-text attributes, then failure to decrypt the encrypted attributes does not cause an automatic rejection. Rather, {{es}} processes the available plain-text attributes (and any `EncryptedAttributes` that could be decrypted).
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md
```diff
@@ -0,0 +1,175 @@
...
+
+:::{tip}
+* In self-managed clusters, file path settings are resolved relative to the {{es}} config directory. {{es}} will automatically monitor this file for changes and will reload the configuration whenever it is updated.
+* If you're using {{ech}} or {{ece}}, then you must upload any certificate or keystore files before they can be referenced in the configuration. For {{ech}}, upload them [as a custom bundle](/deploy-manage/deploy/elastic-cloud/upload-custom-plugins-bundles.md). For {{ece}}, follow the equivalent [ECE procedure](/deploy-manage/deploy/cloud-enterprise/add-custom-bundles-plugins.md). In both cases, you can add the files to your existing SAML bundle.
+* If you're using {{eck}}, then install the files as [custom configuration files](/deploy-manage/deploy/cloud-on-k8s/custom-configuration-files-plugins.md#use-a-volume-and-volume-mount-together-with-a-configmap-or-secret).
+:::
```
shainaraskas: snippet? I think you have this 3x

deploy-manage/users-roles/cluster-or-deployment-auth/saml-without-kibana.md
```diff
@@ -0,0 +1,112 @@
...
+description: Implement SAML authentication for a custom web application using the Elasticsearch SAML REST APIs, without Kibana.
+applies_to:
+  stack: all
+products:
+  - id: elasticsearch
+---
+
+# SAML without {{kib}} [saml-no-kibana]
+
+The SAML realm in {{es}} is designed to allow users to authenticate to {{kib}} and as such, most of the [SAML SSO guide](/deploy-manage/users-roles/cluster-or-deployment-auth/saml.md) makes the assumption that {{kib}} is used. This page describes how a custom web application could use the relevant SAML REST APIs in order to authenticate users to {{es}} with SAML.
```
shainaraskas:
```suggestion
The SAML realm in {{es}} is designed to allow users to authenticate to {{kib}} and as such, most of the [SAML SSO guide](/deploy-manage/users-roles/cluster-or-deployment-auth/saml.md) makes the assumption that {{kib}} is used. This page describes how a custom web application can use SAML REST APIs in order to authenticate users to {{es}} with SAML.
```

deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md
```diff
@@ -0,0 +1,175 @@
+---
+navigation_title: SAML signing and encryption
+description: Configure Elasticsearch to sign outgoing SAML messages and decrypt incoming encrypted assertions, including certificate generation for PEM, PKCS#12, and JKS formats.
+applies_to:
+  stack: all
+products:
+  - id: elasticsearch
+---
+
+# Configure SAML signing and encryption [saml-enc-sign]
```
shainaraskas, line 10: could use some prereq context here - good to treat each of these as "page one" like someone landed from google, so they can find the entrypoint

deploy-manage/users-roles/cluster-or-deployment-auth/saml-without-kibana.md
```diff
@@ -0,0 +1,112 @@
...
+
+::::{note}
+This page assumes that you are familiar with the SAML 2.0 standard and more specifically with the SAML 2.0 Web Browser Single Sign On profile.
+::::
+
+Single sign-on realms such as OpenID Connect and SAML use the Token Service in {{es}} and in principle exchange a SAML or OpenID Connect Authentication response for an {{es}} access token and a refresh token. The access token is used as credentials for subsequent calls to {{es}}. The refresh token enables the user to get new {{es}} access tokens after the current one expires.
+
+## SAML realm [saml-no-kibana-realm]
```
shainaraskas, line 20: should these be numbered steps? 

1. Create a SAML realm in {{es}}
2. Create a service account user

deploy-manage/users-roles/cluster-or-deployment-auth/saml-without-kibana.md
```diff
@@ -0,0 +1,112 @@
...
+
+You must create a SAML realm and configure it accordingly in {{es}}. See [Create a SAML realm](/deploy-manage/users-roles/cluster-or-deployment-auth/saml.md#saml-create-realm).
+
+## Service account user for accessing the APIs [saml-no-kibana-user]
+
+The realm is designed with the assumption that there needs to be a privileged entity acting as an authentication proxy. In this case, the custom web application is the authentication proxy handling the authentication of end users (more correctly, "delegating" the authentication to the SAML Identity Provider). The SAML related APIs require authentication and the necessary authorization level for the authenticated user. For this reason, you must create a Service Account user and assign it a role that gives it the `manage_saml` cluster privilege. The use of the `manage_token` cluster privilege will be necessary after the authentication takes place, so that the service account user can maintain access to refresh access tokens on behalf of the authenticated users or to subsequently log them out.
```
shainaraskas, line 26: perhaps OOS for this PR, but this page could use a plain language pass. these paras are a little wall-of-texty

> (more correctly, "delegating" the authentication to the SAML Identity Provider).

😬

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
+  sp.logout: "https://kibana.example.com/logout"
   attributes.principal: "urn:oid:0.9.2342.19200300.100.1.1"
   attributes.groups: "urn:oid:1.3.6.1.4.1.5923.1.5.1."
 ```
 
-::::{dropdown} Common settings
-
-xpack.security.authc.realms.saml.saml1
-:   Defines a new `saml` authentication realm named "saml1". See [Realms](/deploy-manage/users-roles/cluster-or-deployment-auth/authentication-realms.md) for more explanation of realms.
+1. Controls realm priority. Assign SSO realms higher order values than password-based realms (native, LDAP). If you're using {{eck}}, set `order` to a value greater than the file realm (default `-100`) and native realm (default `-99`).
```
shainaraskas: shall we specify that lower number == higher prio? it's unclear now

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,134 +23,119 @@ products:
...
+::::
 
-This guide will walk you through the most common settings.
+::::{step} Create a SAML realm in {{es}}
+:anchor: saml-create-realm
 
-Create a realm by adding the following to your [`elasticsearch.yml`](/deploy-manage/stack-settings.md) configuration file. Each configuration value is explained below.
+Add a SAML realm to your [{{es}} configuration](/deploy-manage/stack-settings.md#configure-stack-settings) and restart the cluster for the changes to take effect.
 
-If you're using {{ece}} or {{ech}}, and you're using machine learning or a deployment with hot-warm architecture, you must include this configuration in the user settings section for each node type.
```
shainaraskas, line 128: looks like we have this stale note in a couple other places!

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -568,152 +418,25 @@ kubectl cp elasticsearch-sample-es-default-0:/usr/share/elasticsearch/saml-elast
...
-  saml.saml1:
-    order: 0
-    realm: "saml1"
-  basic.basic1:
-    order: 1
-```
-
-If {{kib}} is configured in this way, users are presented with a choice at the Login Selector UI. They log in with SAML or they provide a username and password and rely on one of the other security realms within {{es}}. Only users who have a username and password for a configured {{es}} authentication realm can log in via {{kib}} login form.
-
-Alternatively, when the `basic` authentication provider is enabled, you can place a reverse proxy in front of {{kib}}, and configure it to send a basic authentication header (`Authorization: Basic ....`) for each request. If this header is present and valid, {{kib}} will not initiate the SAML authentication process.
```
shainaraskas, line 698: do we care about this

shainaraskas (approved):
super solid. probably most of my comments are on the old stuff - language nits you didn't introduce you can ignore, but we could also make AI quickly fix the tense issues and maybe some of the italics, unplain language, and parentheticals.

buuut imo largely shipit. thank you 👏 👏 👏 👏 👏 

stray thoughts:

* would be nice if this page linked up to its conceptual home: https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7484/troubleshoot/elasticsearch/security/trb-security-saml

shainaraskas (approved):
super solid. probably most of my comments are on the old stuff - language nits you didn't introduce you can ignore, but we could also make AI quickly fix the tense issues and maybe some of the italics, unplain language, and parentheticals.

buuut imo largely shipit. thank you 👏 👏 👏 👏 👏 

stray thoughts:

* would be nice if this page linked up to its conceptual home: https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7484/troubleshoot/elasticsearch/security/trb-security-saml

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: Yep, that's something I was concerned about... :)

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: yeah... and I included some... =)

sometimes I feel a parenthetical statement, or an em dash makes sense and makes reading easier. But I agree they are not ideal and we shouldn't abuse.

But on the other hand if we remove all of them by policy some of our sentences or paragraphs could feel worse or more difficult to process.

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: the key is to highlight the Kibana privileges section that is available when a creating role, I guess those are mapped to stack features data and kibana access to certain parts of the UI.

Because SAML is intended for kibana access we want to avoid a reader adding permissions to the data but forgetting granting basic Kibana access.

Anyway I'll try to rephrase a bit.

deploy-manage/users-roles/cluster-or-deployment-auth/saml-without-kibana.md
eedugon: Yeah, my plan with this big initial refactor was to not touch some of the docs that are just moved to individual docs (like SAML without kibana, Signing and encryption + Mapping SAML attributes to Elasticsearch user properties). But we can apply a basic level of improvements.

However, from all these docs, I think the less popular and important one is this one (saml without kibana), so I wouldn't put much effort there.

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: what do you mean? 

Are you talking about this?
> If you're using {{ece}} or {{ech}}, and you're using machine learning or a deployment with hot-warm architecture, you must include this configuration in the user settings section for each node type.

If so that's removed on purpose. It's not applicable anymore, as now in ECH and ECE the settings are applicable to all node types by default. Stef already removed a similar paragraph in a recent PR for another SAML related doc.

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: I'll take a look, I think it should be already covered but it worths checking again.
I think yes, we should care about it.

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
shainaraskas: yeah - I think sometimes adding a second sentence with contextualization is a little heavy, but we should try to take a stab and fall back to parentheses only when it breaks up the narrative

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
shainaraskas: yeah I meant that it's in other files, e.g. oidc. agree it doesn't belong and should be removed. outside of the scope of this pr for sure

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
eedugon: great catch

deploy-manage/users-roles/cluster-or-deployment-auth/saml-signing-encryption.md
eedugon: I've included a small intro sentence before the original intro.
But in general this will need future rework, as the introduction (the original one) says the same 4 times, but let's do that at a later stage.

eedugon: Created https://github.com/elastic/docs-content-internal/issues/1585 to keep us busy :P

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: yeah, it's not super clear. The reason is that they are required and recommended considering `attributes.*` settings, not the realm setting in general.

For the realm setting in general we are ignoring that.

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: yes, although it should be more than obvious, as non-production clusters don't listen for connections coming from the outside :)

anyway added to the prereqs that we have specifically for self-managed. Good catch again!

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: Improved, thanks a lot for this!

deploy-manage/users-roles/cluster-or-deployment-auth/saml-attribute-mapping.md
eedugon: yeah it's very useful but i think the main doc also has good directives on this regard, and if we convert it into a snippet we should change the main doc in a significant way now.

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: I agree! The thing is that we are not following the overview / parent page and sub-pages / tutorials and procedures pattern in this PR. That would be a much bigger change.

Created https://github.com/elastic/docs-content-internal/issues/1586 for following this up.

deploy-manage/users-roles/cluster-or-deployment-auth/saml-without-kibana.md
eedugon: https://github.com/elastic/docs-content-internal/issues/1587

deploy-manage/users-roles/cluster-or-deployment-auth/saml-without-kibana.md
eedugon: https://github.com/elastic/docs-content-internal/issues/1587

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: Good finding, but that content was already covered in `kibana-authentication.md`. I'm adding a link to the other doc for more details, as the second sentence is better explained in the other doc and it's kind of irrelevant for SAML setup per se.

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,524 +23,372 @@ products:
...
-`encryption.keystore.path`
-:   The path to the PKCS#12 or JKS keystore. e.g. `saml/saml-crypt.p12`
+Configure the class reference values using `req_authn_context_class_ref` in your SAML realm. For example:
 
-`encryption.keystore.alias`
-:   The alias of the key within the keystore. e.g. `encryption-key`
+```yaml
+xpack.security.authc.realms.saml.saml1:
+  req_authn_context_class_ref:
+    - "http://schemas.microsoft.com/claims/multipleauthn"  # <1>
```
richard-dennehy, line 367: I'm not an expert on Entra SAML, but I'm curious where this has come from? 

Microsoft's docs suggest that `multipleauthn` may be returned in the authentication response to indicate the user has completed MFA, but the list of valid AuthnContextClassRef values that we can request appear to refer to specific authentication methods (password, Kerberos, etc) rather than being able to request MFA.

https://learn.microsoft.com/en-us/entra/identity-platform/single-sign-on-saml-protocol#requestedauthncontext

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
```diff
@@ -23,524 +23,372 @@ products:
...
 
-The IdP will have been assigned an identifier or *EntityID*, which is most commonly expressed in *Uniform Resource Identifier* (URI) form. Your admin interface might tell you what this is, or you might need to read the metadata document to find it - look for the `entityID` attribute on the `EntityDescriptor` element.
+To configure {{es}}, you will need a standard XML-formatted SAML *metadata* document from your IdP, which defines its capabilities and features. You should be able to download or generate it from your IdP's administration interface. You can pass it to {{es}} as a URL (preferred, so {{es}} reloads it automatically when it changes) or as a local file. For more information, refer to [Create a SAML realm in {{es}}](#saml-create-realm).
```
richard-dennehy, line 63: The current docs say that downloading the file is preferred over using a URL. Is there a reason we're changing our recommendation?

Also, local files get reloaded automatically as well

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: Good catch here @richard-dennehy , thanks!

My goal here was to enrich the original section by adding an example that the reader could find useful (we didn't offer any example in the original section). But it's true that the example is incorrect, and I mixed Entra's `authnmethodsreferences` with supported `RequestedAuthnContext` class references.

Although the AI agent says there are reports that requesting multipleauthn can work in some configurations, Microsoft does not document it as a supported requested value. I'm replacing the example with standard class-reference examples and clarify that the configured values must be supported by the IdP.

Thanks!

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: You have found an interesting inconsistency. Not big but I think it's worthy to solve it. Let me explain the background:

At first, our existing SAML guide doesn't really says file is preferred over using a URL. It just shows the file message with the explanation about URL being supported also.

But it's true that in the Elasticsearch reference documentation, when explaining the setting, it recommends file over a URL.

On the other hand, from my old experience in support, for environments like ECH, ECE, or ECK we have always recommended URLs over files to avoid the need of building, uploading and maintaining a bundle (ECE & ECH), or needing to create a Secret or ConfigMap + a volume and volumeMount in the Elasticsearch resource specification (ECK).

Anyway I didn't intend to reverse that recommendation. So I'll restore the local-file example, remove the incorrect reload rationale, and explain the operational trade-off for hosted and Kubernetes deployments.

Let me know your thoughts or if you prefer a different approach! Thanks!

eedugon: addressed at https://github.com/elastic/docs-content/pull/7484/commits/3e9586c74193bd502b021e0d032a0638e21c86ed

deploy-manage/users-roles/cluster-or-deployment-auth/saml.md
eedugon: Addressed at https://github.com/elastic/docs-content/pull/7484/commits/e377425481e574b489a0094bf06ebb2237859f76

richard-dennehy (approved):

PR #7497, opened 2026-07-21
[Sig Events] Add architecture and operator docs

mdbirnstiehl:
## Summary
This PR closes https://github.com/elastic/nightshift-program/issues/499 and adds Significant Events architecture and operator docs.  These docs are hidden for a private preview, except for the knowledge indicators docs were already public.

- `index.md`: Overview of the Significant Events feature.
- `how-it-works.md`: Pipeline behavior for engineers and adopters covering KI extraction, rule generation, rule execution, and discovery phase (Detection, Discovery, Triage). 
**Note:** I added a mermaid diagram here to try to get the overall message. We can make one that fits the docs style a little better if the content is correct.
- `knowledge-indicators.md`: Updates the existing KI docs.
- `operator-guide.md`: Operational reference covering component placement (what runs on Kibana vs Elasticsearch), system impact (query load, pipeline lag, memory, storage growth), LLM cost drivers by phase, and recovery procedures for common degraded states.
## Generative AI disclosure

- [x] Yes  
- [ ] No  

Tool(s) and model(s) used: Claude code


solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,154 @@
...
+
+| What to query | {{esql}} target |
+|---|---|
+| Knowledge Indicators | `FROM .significant_events-knowledge_indicators` |
+| Alert documents (rule execution results) | `FROM .rule-events` |
+| Detection state per rule | `FROM .significant_events-detections` |
+| Discovery hypotheses | `FROM .significant_events-discoveries` |
+| Significant Events | `FROM .significant_events-events` |
+
+**Traceability**: The `kibana.alert.rule.uuid` in alert documents links back to the rule definition stored in the stream. The `kibana.alert.rule.tags` array always includes the stream name, so you can filter the alerts index by stream without a join. Across the discovery phase, `discovery_slug` is the durable cross-index key linking discoveries to events. To trace a Significant Event back to its origin, follow `discovery_slug` from the events index to the discoveries index, then use `detections[].detection_id` to reach the detection document, and `kibana.alert.rule.uuid` to reach the alerting rule.
```
crespocarlos: This is outdated:
- instead of `discovery_slug` we should use `event_id`
- `detections[].detection_id` -> `signals[].metadata.detection_id` (this is used to connect to `.significant_events-detections`
- to reach the alert index: `signals[].metadata.rule_uuid` 


Should we mention the alert index here?

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,154 @@
...
+## {{esql}} traceability [sig-events-hiw-traceability]
+
+{{esql}} is the exclusive query language across the pipeline. Every rule is expressed as {{esql}}. Every artifact is stored as an {{es}} document. This means any point in the pipeline is queryable without leaving {{esql}}.
+
+| What to query | {{esql}} target |
+|---|---|
+| Knowledge Indicators | `FROM .significant_events-knowledge_indicators` |
+| Alert documents (rule execution results) | `FROM .rule-events` |
+| Detection state per rule | `FROM .significant_events-detections` |
+| Discovery hypotheses | `FROM .significant_events-discoveries` |
```
crespocarlos: We're planning to remove this. Should we mention it in the docs?

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,154 @@
...
+
+Discovery converts raw alert signals into confirmed Significant Events. It runs as three sequenced workflows (detection, discovery, and triage) coordinated by an Orchestrator workflow.
+
+- [Detection Workflow](#sig-events-hiw-detection) — `change_point` analysis per alerting rule, written to the detections index
+- [Discovery Workflow](#sig-events-hiw-discovery-agent) — discovery agent generates hypotheses, written to the discoveries index
+- [Triage Workflow](#sig-events-hiw-triage) — judge agent independently verifies and promotes, written to the events index
+
+### Detection [sig-events-hiw-detection]
+
+The Detection Workflow reads `.rule-events` and runs `change_point` analysis grouped per alerting rule. When a rule's alert pattern enters a genuinely anomalous state, the workflow appends a detection document to the detections index. The index is append-only — `detection`, `quiet`, and `handled` document kinds track anomaly state over time. Current state per rule is the latest document.
```
crespocarlos: detection`, `quiet`, and `handled` document kinds have been removed. detection docs don't have kind

solutions/observability/streams/significant-events/how-it-works.md
crespocarlos: Should we add examples of queries to match `events` with `detections`?

solutions/observability/streams/significant-events/index.md
```diff
@@ -0,0 +1,59 @@
...
+
+Significant Events continuously converts raw log data into a ranked feed of incidents. The pipeline moves from broad to narrow: many alerting rule executions produce fewer statistical detections, which an AI agent correlates into even fewer discoveries, which a judge reviews and surfaces as Significant Events.
+
+For example: you onboard a stream, Significant Events extracts what's in it ("this stream contains a Python gRPC service called `recommendationservice` running on GKE"), generates targeted detection rules ("detect gRPC connection failures"), and runs those rules continuously. Results flow through discovery and surface as correlated Significant Events across your entire environment.
+
+## Before you get started [sig-events-prerequisites]
+
+Significant Events requires an **Enterprise license** or an active Enterprise trial.
+
+To use Significant Events you also need:
```
crespocarlos: There is this feature flag:

```
feature_flags.overrides:
  streams.significantEventsAvailable: true
```

solutions/observability/streams/significant-events/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+# Significant Events [streams-sig-events-overview]
+
+Significant Events continuously converts raw log data into a ranked feed of incidents. The pipeline moves from broad to narrow: many alerting rule executions produce fewer statistical detections, which an AI agent correlates into even fewer discoveries, which a judge reviews and surfaces as Significant Events.
+
+For example: you onboard a stream, Significant Events extracts what's in it ("this stream contains a Python gRPC service called `recommendationservice` running on GKE"), generates targeted detection rules ("detect gRPC connection failures"), and runs those rules continuously. Results flow through discovery and surface as correlated Significant Events across your entire environment.
+
+## Before you get started [sig-events-prerequisites]
+
+Significant Events requires an **Enterprise license** or an active Enterprise trial.
```
nastasha-solomon, line 26: The page-level `applies_to` tags makes me think this feature is available in serverless too, so you prob need to add the [serverless-equivalent](https://www.elastic.co/docs/solutions/observability/observability-serverless-feature-tiers#obs-subscription-features) here.

solutions/observability/streams/significant-events/index.md
```diff
@@ -0,0 +1,60 @@
...
+  - id: cloud-serverless
+  - id: cloud-hosted
+  - id: cloud-enterprise
+  - id: cloud-kubernetes
+  - id: elastic-stack
+---
+
+# Significant Events [streams-sig-events-overview]
+
+Significant Events continuously converts raw log data into a ranked feed of incidents. The pipeline moves from broad to narrow: many alerting rule executions produce fewer statistical detections, which an AI agent correlates into even fewer discoveries, which a judge reviews and surfaces as Significant Events.
```
nastasha-solomon: Good idea to present a simplified version of the flow, then a concrete example to anchor it. My only suggestion would be to remove some of the specialized terms since this is the very first explanation of the feature that readers/users would encounter in the docs, and it should be simple enough for anyone to understand. For example:


```suggestion
Significant Events continuously converts raw log data into a ranked feed of incidents. The pipeline moves from broad to narrow. Many alerts fire, an AI agent narrows these down into findings, and the most important ones are surfaced as Significant Events.
```

solutions/observability/streams/significant-events/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Before you get started [sig-events-prerequisites]
+
+Significant Events requires an **Enterprise license** or an active Enterprise trial.
+
+To use Significant Events you also need:
+
+- The `streams.significantEventsAvailable: true` feature flag enabled in your {{kib}} configuration.
+- The `observability:streamsSigEventsScheduledDiscoveryEnabled` {{kib}} setting enabled (required for the detection, discovery, and triage workflows).
+- A [Generative AI connector](kibana://reference/connectors-kibana/gen-ai-connectors.md), which routes LLM calls and incurs additional costs.
```
nastasha-solomon: Idk if you want to include that last detail here. If you need to, it could be better to add it as a note so it's not part of the main requirement. In it's current state, the line reads as "You specifically need a Generative AI connector that routs LLM calls _and_ costs you extra money". I'm fairly positive we're not  suggesting that 😅 (though I can see Claude recommending it 🤖 )

```suggestion
- A [Generative AI connector](kibana://reference/connectors-kibana/gen-ai-connectors.md), which routes LLM calls.
```

solutions/observability/streams/significant-events/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+## How the pipeline works [sig-events-pipeline]
+
+The pipeline runs in sequential phases that build on the output of the previous phase. KI extraction and rule generation prepare your streams for detection and run on your schedule. Detection and discovery run continuously in the background after you enable Significant Events.
```
nastasha-solomon: Define first instance

```suggestion
The pipeline runs in sequential phases that build on the output of the previous phase. Knowledge Indicator (KI) extraction and rule generation prepare your streams for detection and run on your schedule. Detection and discovery run continuously in the background after you enable Significant Events.
```

solutions/observability/streams/significant-events/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+### 1. Knowledge indicator extraction [sig-events-phase-ki]
+
+An LLM analyzes log samples from your streams and extracts services, infrastructure, dependencies, and failure patterns as Knowledge Indicators (KIs). Deterministic generators run in parallel to produce dataset statistics, log patterns, and error samples.
```
nastasha-solomon, line 36: Quick note about terminology - Earlier on the page, the actor performing the extraction is an AI agent. Here, it's an LLM. While the two are prob used interchangeably, it'd be good to settle on a single entity for the sake of consistency. 

```suggestion
An LLM analyzes log samples from your streams and extracts services, infrastructure, dependencies, and failure patterns as KIs. Deterministic generators run in parallel to produce dataset statistics, log patterns, and error samples.
```

solutions/observability/streams/significant-events/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+### 2. Rule generation [sig-events-phase-rules]
+
+An LLM converts KIs into {{esql}} alerting rules scoped to the services and failure modes present in each stream. Generated rules are saved as unbacked stream assets — you review and promote them before they start running.
+
+### 3. Rule execution and detection [sig-events-phase-detection]
+
+Promoted rules run against stream data on a per-rule schedule and write matching events to the alerts index. A detection workflow runs `change_point` analysis across those results every minute and writes detection records for statistically significant transitions.
```
nastasha-solomon: (Revisiting this comment to apply my understanding from digging through the nightshift program repo)

Change point analysis might be an internal term - it's not used in Elasticsearch's change point aggregation [docs](https://www.elastic.co/docs/reference/aggregations/search-aggregations-change-point-aggregation). It's also be good to replace "detection records" with "detection documents", which is used more often throughout this PR.

Here's the revision I came up with after spending some time in the nightshift program repo:

```suggestion
Promoted rules execute on a specified schedule against stream logs and record a count of matches per time bucket to the `.rule-events` index. A Detection workflow runs change point aggregation against those per-rule counts and writes a detection record whenever the aggregation finds a statistically significant shift.
```

solutions/observability/streams/significant-events/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+### 4. Discovery [sig-events-phase-discovery]
+
+Two agents run in sequence. The Discovery agent reads detection records, correlates signals across rules and streams, and writes discovery records. The Judge agent independently verifies each discovery and promotes, acknowledges, or dismisses it as a Significant Event.
```
nastasha-solomon: This is cool - new agents. If these are built-in agents, the [Elastic Agent docs](https://www.elastic.co/docs/explore-analyze/ai-features/agent-builder/builtin-agents-reference) will prob need to be updated at some point.

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -15,22 +15,24 @@ products:
   - id: elastic-stack
 ---
 
-# Knowledge Indicators [streams-knowledge-indicators]
+# Knowledge Indicators [sig-events-ki]
 
 Knowledge Indicators (KIs) are structured facts that Elastic extracts from your raw log data automatically without requiring schemas, service catalogs, or manual configuration. When you run extraction against a log stream, Elastic analyzes the raw data and returns facts about your environment: which services are running, the underlying infrastructure they rely on, how they depend on each other, and the log schemas they use.
 
-Rather than a static configuration, this knowledge accumulates over time, automatically expires when a service disappears, and feeds directly into downstream capabilities like Rules, topology maps, AI agent investigations, and dashboards.
+Rather than a static configuration, this knowledge accumulates over time, automatically expires when a service disappears, and feeds directly into downstream capabilities like the Significant Events detection pipeline, topology maps, AI agent investigations, and dashboards.
```
nastasha-solomon: This would allow you to the remove the next sentence, which is kinda floating in between the paragraphs.

```suggestion
Rather than a static configuration, this knowledge accumulates over time, automatically expires when a service disappears, and feeds directly into downstream capabilities like the [Significant Events detection pipeline](../significant-events/how-it-works.md), topology maps, AI agent investigations, and dashboards.
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -15,22 +15,24 @@ products:
...
+
+KIs feed the Significant Events pipeline. See [How Significant Events works](../significant-events/how-it-works.md) for how KIs connect to detection, discovery, and triage.
```
nastasha-solomon:
```suggestion
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -15,22 +15,24 @@ products:
...
 
 To access Knowledge Indicators, open **Significant Events** from the Streams main page and select the **Knowledge Indicators** tab.
 
 :::{admonition} Requirements
 To use this feature, you need:
 
 - A [Generative AI connector](kibana://reference/connectors-kibana/gen-ai-connectors.md), which can incur additional costs.
```
nastasha-solomon:
```suggestion
- A [Generative AI connector](kibana://reference/connectors-kibana/gen-ai-connectors.md).
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -40,15 +42,23 @@ On demand
...
+## How KI extraction works [sig-events-ki-extraction]
 
 The extraction pipeline samples a small batch of logs from a stream and processes them through a combination of large language model (LLM) analysis and deterministic code generators. It accumulates its findings across multiple iterations, entirely configuration-free.
 
-The pipeline runs multiple iterations, each time fetching a small sample of documents using a mix of random and entity-filtered documents to ensure full coverage of the system. KIs found in one iteration are fed back as exclusions into the next, so each round focuses on what the previous one missed — ensuring quieter, less-represented services aren't crowded out by noisier ones.
+The pipeline runs up to five iterations. Each iteration fetches a batch of up to 20 documents assembled from three prioritized buckets:
+
+| Bucket | Share | Strategy |
+|---|---|---|
+| Entity-filtered | 40% | Random sample that excludes documents matching already-discovered feature filters, steering toward undiscovered patterns |
```
nastasha-solomon: What's a feature filter? I don't think I've heard of that term before.

nastasha-solomon: Oh, it's defined further down. It'd be good to define it here first, just so users aren't wondering what the term means.


```suggestion
| Entity-filtered | 40% | Random sample that excludes documents matching already-discovered feature filters (features excluded by users), steering toward undiscovered patterns |
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -75,35 +85,35 @@ The pipeline extracts:
...
 | Entity | Distinct system components: services, applications, jobs |
-| Infrastructure | Environment context: Kubernetes, cloud provider, OS |
+| Infrastructure | Environment context: {{k8s}}, cloud provider, OS |
 | Technology | Languages, frameworks, libraries, databases |
 | Dependency | Relationships between components |
-| Schema | Log format conventions: Elastic Common Schema (ECS), OTel, custom |
+| Schema | Log format conventions: {{product.ecs}}, OTel, custom |
 
-Every feature must include stable identifying properties and cite direct evidence from the sampled logs. The LLM assigns a confidence score from 0–100 for each KI. Features intentionally excluded by users (false positives) are also tracked and carried forward to prevent re-identification in future runs.
+Every feature must include stable identifying properties and cite direct evidence from the sampled logs. The LLM assigns a confidence score from 0–100 for each KI. The pipeline also tracks features excluded by users (false positives) and carries them forward to prevent re-identification in future runs.
```
nastasha-solomon: In a future iteration, it'd be useful to have a section explaining the confidence score ranges. For now, it'd be sufficient to explain the 0-100 scale, for example:


```suggestion
Every feature must include stable identifying properties and cite direct evidence from the sampled logs. The LLM assigns a confidence score from 0–100 (where 0 is the lowest confidence and 100 is the highest) for each KI. The pipeline also tracks features excluded by users (false positives) and carries them forward to prevent re-identification in future runs.
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -75,35 +85,35 @@ The pipeline extracts:
...
 
-### Deterministic generators [streams-ki-generators]
+### Deterministic generators [sig-events-ki-generators]
 
 In parallel with LLM analysis, a set of deterministic code-based generators independently analyze the data to produce statistical summaries, log samples, pattern clusters, and error-specific features. Because these are computed rather than inferred, they always receive a confidence score of 100.
 
-### Merging results [streams-ki-merge]
+### Merging results [sig-events-ki-merge]
```
nastasha-solomon: A possible alternative that aligns with the format the other section titles follow.

```suggestion
### Results handling [sig-events-ki-merge]
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -160,24 +170,55 @@ Example query KI:
...
+**Stats queries** aggregate metrics over time — rates, counts, or derived values — useful when the signal is a change in volume rather than the presence of a specific event:
 
-**Rules**: Query KIs automatically generate active rules to surface signals without manual configuration.
+```esql
+FROM logs-mystream,logs-mystream.*
+| STATS error_count = COUNT(*) WHERE level == "ERROR" BY service.name, @timestamp
+```
 
-**Topology graphs**: Dependency KIs construct an infrastructure graph inferred entirely from log data. No distributed tracing or manual configuration is required. During an incident, the graph immediately shows which upstream services are affected when a specific database or service goes down.
+### Downstream path: from query KIs to alerting rules [sig-events-ki-downstream]
```
nastasha-solomon:
```suggestion
### Downstream path: From query KIs to alerting rules [sig-events-ki-downstream]
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+  - id: cloud-enterprise
+  - id: cloud-kubernetes
+  - id: elastic-stack
+---
+
+# Operator guide [sig-events-operator]
+
+Use this guide to understand how Significant Events affects your cluster and how to manage it.
+
+- [What runs where](#sig-events-op-components): which components run on {{kib}}, {{es}}, and Workflows
```
nastasha-solomon:
```suggestion
- [What runs where](#sig-events-op-components): Which components run on {{kib}}, {{es}}, and Workflows
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+- [System impact](#sig-events-op-impact): query load, pipeline lag, memory, and storage growth
```
nastasha-solomon:
```suggestion
- [System impact](#sig-events-op-impact): Query load, pipeline lag, memory, and storage growth
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+- [Cost drivers](#sig-events-op-costs): LLM call sites and token usage by phase
+- [Disable and re-enable](#sig-events-op-disable): how to stop the pipeline or pause KI refresh
```
nastasha-solomon:
```suggestion
- [Disable and re-enable](#sig-events-op-disable): How to stop the pipeline or pause Knowledge indicator (KI) refresh
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+- [Recovery procedures](#sig-events-op-recovery): symptoms and actions for common degraded states
```
nastasha-solomon:
```suggestion
- [Recovery procedures](#sig-events-op-recovery): Symptoms and actions for common degraded states
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+
+## What runs where [sig-events-op-components]
+
+The following table shows each pipeline component, where it runs, what triggers it, and what it reads and writes:
+
+| Component | Runs on | Trigger | Reads | Writes |
+|---|---|---|---|---|
+| Knowledge indicator (KI) feature identification | {{kib}} (Task Manager + workflow) | On-demand / continuous extraction | Stream logs | `.significant_events-knowledge_indicators` |
```
nastasha-solomon: I'm not sure how well "On-demand / continuous extraction" fits in the Trigger column, which describes what triggers the pipeline component.

```suggestion
| KI feature identification  | Task Manager + Workflow | On-demand / continuous extraction | Stream logs | `.significant_events-knowledge_indicators` |
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+| KI query generation | {{kib}} (workflow) | On-demand | Features + existing queries | Query KI assets |
```
nastasha-solomon: Same comment as above - I'm not sure "On-demand" works in the Trigger column.

```suggestion
| KI query generation | Workflow | On-demand | Features + existing queries | Query KI assets |
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+- [System impact](#sig-events-op-impact): query load, pipeline lag, memory, and storage growth
+- [Cost drivers](#sig-events-op-costs): LLM call sites and token usage by phase
+- [Disable and re-enable](#sig-events-op-disable): how to stop the pipeline or pause KI refresh
+- [Recovery procedures](#sig-events-op-recovery): symptoms and actions for common degraded states
+
+## What runs where [sig-events-op-components]
+
+The following table shows each pipeline component, where it runs, what triggers it, and what it reads and writes:
+
+| Component | Runs on | Trigger | Reads | Writes |
```
nastasha-solomon: Since the second column contains a mix of Solutions and features, a more general term like "Uses" might be better than "Runs on". I'd also recommend sticking with either solutions or features, not both. The following suggestions use features.   

```suggestion
| Component | Uses | Trigger | Reads | Writes |
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+|---|---|---|---|---|
+| Knowledge indicator (KI) feature identification | {{kib}} (Task Manager + workflow) | On-demand / continuous extraction | Stream logs | `.significant_events-knowledge_indicators` |
+| KI query generation | {{kib}} (workflow) | On-demand | Features + existing queries | Query KI assets |
+| Alerting rule execution | {{kib}} alerting → {{es}} | Per-rule schedule | Stream data via {{esql}} | `.rule-events` |
+| Detection workflow | {{kib}} Workflows | Cron 10m | `.rule-events` | `.significant_events-detections` |
```
nastasha-solomon:
```suggestion
| Detection Workflow | Workflow | Cron 10m | `.rule-events` | `.significant_events-detections` |
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+| Discovery workflow | {{kib}} Workflows + Agent Builder | Cron 10m | `.significant_events-detections` + KIs | `.significant_events-discoveries` |
```
nastasha-solomon:
```suggestion
| Discovery Workflow | Workflows + Agent Builder | Cron 10m | `.significant_events-detections` + KIs | `.significant_events-discoveries` |
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
+---
+navigation_title: How it works
+description: Pipeline internals for engineers and adopters — KI extraction, rule generation, detection, discovery, storage model, and ES|QL traceability.
+applies_to:
+  serverless: beta
```
nastasha-solomon: I could be wrong, but I'm fairly certain that we shouldn't use `beta` anymore when describing a lifecycle state. Here are the resources I'm referencing for guidance:
- [Versioning and availability](https://www.elastic.co/docs/get-started/versioning-availability#lifecycle-states)
- [Feature Lifecycle and Support Model Matrix (Experimental, Preview, GA)](https://docs.google.com/document/d/1pwkifJHcYKGxf8J2Swd5DcWPZ0OhwPgwAubwbEz7-Mw/edit?usp=sharing)


@bmorelli25 if we shouldn't be using `beta`, we should (or I can) update the examples in our `applies_to` [syntax guidance](https://elastic.github.io/docs-builder/syntax/applies/).

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+This page describes the pipeline behavior for engineers and adopters who need to understand how Significant Events processes data, what runs where, and how to trace results across the system.
+
+The pipeline has four sequential phases, each building on the outputs of the previous:
+
+- [Phase 1: KI extraction](#sig-events-hiw-ki): LLM and programmatic analysis of stream logs
+- [Phase 2: Rule generation](#sig-events-hiw-rules): LLM generates detection rules using KIs as input
+- [Phase 3: Rule execution](#sig-events-hiw-execution): Alerting framework runs promoted rules continuously
+- [Phase 4: Discovery](#sig-events-hiw-discovery): Detection, discovery, and triage workflows convert alert signals into confirmed Significant Events
+
+```mermaid
```
nastasha-solomon: This visual is nice, but the distance between the first and second phases is comically long (like Littlefoot's-neck-in-The-Land-Before-Time-long). It's also not easy to scroll through the flowchart after expanding it. Last thiing - there's something a lil funny going on with one of the cells 🫣 I've not worked with mermaid diagrams before so idk how to fix it:

Image

If you can shorten the distance between the first and second phase, that'd be optimal. If that's not possible, it'd be good to revisit this flowchart down the line or use a different tool to generate it.

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+  - id: elastic-stack
+---
+
+# How Significant Events works [sig-events-how-it-works]
+
+This page describes the pipeline behavior for engineers and adopters who need to understand how Significant Events processes data, what runs where, and how to trace results across the system.
+
+The pipeline has four sequential phases, each building on the outputs of the previous:
+
+- [Phase 1: KI extraction](#sig-events-hiw-ki): LLM and programmatic analysis of stream logs
```
nastasha-solomon: First instance

```suggestion
- [Phase 1: Knowledge Indicators (KIs) extraction](#sig-events-hiw-ki): LLM and programmatic analysis of stream logs
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+
+    classDef llm fill:#5c3fa3,color:#fff,stroke:none
+    classDef det fill:#1d6b47,color:#fff,stroke:none
+    class KI,RG,DA,JA llm
+    class RE,DW det
+```
+
+## Phase 1: KI extraction [sig-events-hiw-ki]
+
+Knowledge Indicators (KIs) are stable, evidence-backed facts about the services and infrastructure present in a stream's log data. The LLM that generates detection operates entirely on KIs. This keeps rule generation deterministic and decouples detection logic from data format specifics.
```
nastasha-solomon:
```suggestion
KIs are stable, evidence-backed facts about the services and infrastructure present in a stream's log data. The LLM that generates detection operates entirely on KIs. This keeps rule generation deterministic and decouples detection logic from data format specifics.
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+
+The pipeline runs up to five iterations, sampling up to 20 documents per round from entity-filtered, diverse, and random buckets. KIs found in one iteration are excluded from the next, steering each round toward undiscovered patterns. LLM analysis and deterministic code generators run in parallel; deterministic KIs always score 100 confidence. See [KI extraction](./knowledge-indicators.md#sig-events-ki-extraction) for the full extraction flow.
+
+### KI types
+
+KIs cover five types: Entity, Infrastructure, Technology, Dependency, and Schema. Entity KIs have the highest extraction priority — services and databases are identified before individual pods or hosts. See [KI types](./knowledge-indicators.md#sig-events-ki-types) for the full data model.
+
+### KI lifecycle
+
+KIs are written to `.significant_events-knowledge_indicators`, deduplicated on `(type, subtype, properties)`, and expire after 30 days if not re-observed. See [KI lifecycles](./knowledge-indicators.md#sig-events-ki-lifecycle) for lifecycle and continuous extraction details.
```
nastasha-solomon: Some minor adjustments for clarity. As usual, fix if wrong.

```suggestion
KIs are written to `.significant_events-knowledge_indicators` and deduplicated on `type`, `subtype`, and `properties`. KIs expire after 30 days if they aren't re-observed. See [KI lifecycles](./knowledge-indicators.md#sig-events-ki-lifecycle) for lifecycle and continuous extraction details.
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Phase 2: Rule generation [sig-events-hiw-rules]
+
+Once KIs exist for a stream, the LLM generates detection rules using them as input. Rules are expressed as native {{esql}} expressions targeting the stream's index pattern:
```
nastasha-solomon: Made some adjustments because some of the language (e.g., "rules are expressed as") reads as implementation details. Hopefully this is more direct and still accurate: 

```suggestion
Once a stream has KIs, the LLM generates a detection rule scoped to that stream. The generated rule is an {{esql}} query with conditions derived from the stream's KIs:
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+## Phase 3: Rule execution [sig-events-hiw-execution]
+
+Each promoted rule runs as a {{kib}} alerting rule on a per-rule schedule. On each execution cycle, the rule:
+
+1. Builds an {{esql}} query from the stored rule parameters.
+2. Queries the stream with a lookback window of 2× the rule interval to ensure no events fall through the gap between runs.
+3. Excludes document IDs seen in previous executions.
+4. Writes matching rows to `.rule-events`.
+
+**Execution cap**: Each rule execution writes at most 1,000 alert documents. Matching events beyond this limit are dropped.
```
nastasha-solomon: Potential revision that adds a bit more detail to improve clarity:

```suggestion
**Write limits**: Each time a rule executes, it can write up to 1000 documents. Once this limit is met, documents aren't written to `.rule-events` until the next time the rule executes. 
```

solutions/observability/streams/significant-events/index.md
nastasha-solomon: Nvm, it's already doc'd [here](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7497/solutions/observability/streams/significant-events/how-it-works#sig-events-hiw-discovery-agent)! This would be a good place to link that section.

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Phase 4: Discovery [sig-events-hiw-discovery]
+
+Discovery converts raw alert signals into confirmed Significant Events. It runs as three sequenced workflows (detection, discovery, and triage) coordinated by an Orchestrator workflow.
+
+- [Detection Workflow](#sig-events-hiw-detection) — `change_point` analysis per alerting rule, written to the detections index
```
nastasha-solomon:
```suggestion
- [Detection Workflow](#sig-events-hiw-detection) — Change point aggregation per alerting rule, written to the detections index
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+- [Discovery Workflow](#sig-events-hiw-discovery-agent) — discovery agent generates hypotheses, written to the discoveries index
+- [Triage Workflow](#sig-events-hiw-triage) — judge agent independently verifies and promotes, written to the events index
+
+### Detection [sig-events-hiw-detection]
+
+The Detection Workflow reads `.rule-events` and runs `change_point` analysis grouped per alerting rule. When a rule's alert pattern enters a genuinely anomalous state, the workflow appends a detection document to the detections index. Current state per rule is the latest document.
```
nastasha-solomon: This took me some time to understand. I queried the nightshift program repo to answer my questions and came up with the following revision:

```suggestion
A Detection workflow reads `.rule-events`, buckets each rule's alerts by time, and runs the change point aggregation function against those counts to detect statistically significant shifts. When a shift is found, a document is appended to the detections index. The most recent detection document for a rule reflects its current state, for example, whether it's still producing an unusual rate of alerts or has returned to normal.
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -160,24 +170,55 @@ Example query KI:
...
 
-**AI agent investigations**: Instead of reconstructing basic facts during every incident, an AI agent begins with your system's actual topology and known failure modes. It identifies the relevant streams, runs the applicable queries, and formulates a specific hypothesis using the available KI context.
+When you promote a query KI, it becomes a {{kib}} alerting rule. Each promoted rule runs its {{esql}} query on a per-rule schedule and writes results to `.rule-events`. The number of promoted query KIs is the main driver of alerting query load on your cluster.
 
-**Dashboards**: KIs drive AI-generated dashboard suggestions for your streams.
+The Significant Events pipeline picks up from there: a detection workflow runs the `change_point` aggregation over alert firing patterns and writes significant transitions to `.significant_events-detections`. The discovery workflow then processes those detections with an AI agent.
```
nastasha-solomon:
```suggestion
The Significant Events pipeline picks up from there. A Detection workflow runs change point aggregation over alert generation patterns and writes documents to `.significant_events-detections` when statistically significant shifts are found. A Discovery workflow then uses an LLM to process those documents.
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+
+Streams must have recent log data for KI extraction to produce useful results. Empty streams or streams with very sparse data produce weak KIs, which produce weak query KIs, which produce fewer promoted rules.
+
+**Storage growth**
+
+Significant Events writes to the following data streams:
+
+| Data stream | Written by | Growth driver |
+|---|---|---|
+| `.significant_events-detections` | Detection workflow | Append-only; one document per observed state transition per rule |
```
nastasha-solomon:
```suggestion
| `.significant_events-detections` | Detection Workflow | Append-only; one document per observed state transition per rule |
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+
+Change point detection is per-rule: a stream can have many independent rules, and one rule recovering does not collapse the signal from others still anomalous on the same stream.
+
+### Discovery agent [sig-events-hiw-discovery-agent]
+
+The discovery workflow reads unhandled detection documents and calls the discovery agent. The discovery agent is a hypothesis agent. It reads detection signals and produces structured discovery documents describing what is happening. Its scope is observation only, describing the anomaly, not prescribing remediation.
```
nastasha-solomon:
```suggestion
A Discovery workflow reads unhandled detection documents and calls the Discovery agent. The Discovery agent is a hypothesis agent. It reads detection signals and produces structured discovery documents describing what is happening. Its scope is observation only, describing the anomaly, not prescribing remediation.
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+
+- [Detection Workflow](#sig-events-hiw-detection) — `change_point` analysis per alerting rule, written to the detections index
+- [Discovery Workflow](#sig-events-hiw-discovery-agent) — discovery agent generates hypotheses, written to the discoveries index
+- [Triage Workflow](#sig-events-hiw-triage) — judge agent independently verifies and promotes, written to the events index
+
+### Detection [sig-events-hiw-detection]
+
+The Detection Workflow reads `.rule-events` and runs `change_point` analysis grouped per alerting rule. When a rule's alert pattern enters a genuinely anomalous state, the workflow appends a detection document to the detections index. Current state per rule is the latest document.
+
+Change point detection is per-rule: a stream can have many independent rules, and one rule recovering does not collapse the signal from others still anomalous on the same stream.
```
nastasha-solomon:
```suggestion
Change point detection is per-rule. A stream can have many independent rules, and one rule recovering does not collapse the signal from others still anomalous on the same stream.
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -160,24 +170,55 @@ Example query KI:
...
+```esql
+FROM logs-mystream,logs-mystream.*
+| STATS error_count = COUNT(*) WHERE level == "ERROR" BY service.name, @timestamp
+```
 
-**Topology graphs**: Dependency KIs construct an infrastructure graph inferred entirely from log data. No distributed tracing or manual configuration is required. During an incident, the graph immediately shows which upstream services are affected when a specific database or service goes down.
+### Downstream path: from query KIs to alerting rules [sig-events-ki-downstream]
 
-**AI agent investigations**: Instead of reconstructing basic facts during every incident, an AI agent begins with your system's actual topology and known failure modes. It identifies the relevant streams, runs the applicable queries, and formulates a specific hypothesis using the available KI context.
+When you promote a query KI, it becomes a {{kib}} alerting rule. Each promoted rule runs its {{esql}} query on a per-rule schedule and writes results to `.rule-events`. The number of promoted query KIs is the main driver of alerting query load on your cluster.
```
nastasha-solomon: Making some tweaks based on info from the program repo. 

```suggestion
When you promote a query KI, it becomes an alerting rule that runs its {{esql}} query on a schedule and writes a match count to `.rule-events` for each time bucket. The number of promoted query KIs, and how often each one executes, is the main driver of alerting query load on your cluster.
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -160,24 +170,55 @@ Example query KI:
...
 
-**Dashboards**: KIs drive AI-generated dashboard suggestions for your streams.
+The Significant Events pipeline picks up from there: a detection workflow runs the `change_point` aggregation over alert firing patterns and writes significant transitions to `.significant_events-detections`. The discovery workflow then processes those detections with an AI agent.
 
-**Grok patterns**: KIs inform Grok pattern generation when you introduce new streams.
+## Continuous extraction [sig-events-ki-continuous]
 
-## KI lifecycle and maintenance [streams-ki-lifecycle]
+When continuous extraction is enabled, a {{kib}} Workflow runs every 35 minutes and processes up to five eligible streams per run. A stream is eligible if:
```
nastasha-solomon:
```suggestion
When continuous extraction is enabled, a workflow runs every 35 minutes and processes up to five eligible streams per run. A stream is eligible if:
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -160,24 +170,55 @@ Example query KI:
...
+
+### Continuous extraction settings
+
+| Setting | Default | Description |
+|---|---|---|
+| `observability:streamsContinuousKiExtractionEnabled` | `false` | Enables or disables continuous extraction |
+| `observability:streamsContinuousKiExtractionIntervalHours` | `12` | Minimum hours between extraction runs for a single stream |
+| `observability:streamsContinuousKiExtractionExcludedStreamPatterns` | `""` | Comma-separated glob patterns for streams to skip |
+
+These settings are only modifiable through the Significant Events Settings page, not through {{kib}} Advanced Settings.
```
nastasha-solomon: Statement stands without the last part.

```suggestion
These settings are only modifiable through the Significant Events Settings page.
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Discovery agent [sig-events-hiw-discovery-agent]
+
+The discovery workflow reads unhandled detection documents and calls the discovery agent. The discovery agent is a hypothesis agent. It reads detection signals and produces structured discovery documents describing what is happening. Its scope is observation only, describing the anomaly, not prescribing remediation.
+
+### Triage and the judge agent [sig-events-hiw-triage]
```
nastasha-solomon:
```suggestion
### Triage and the Judge agent [sig-events-hiw-triage]
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+
+The triage workflow reads unassessed discovery documents and calls the judge agent. The judge agent independently verifies the discovery agent's hypothesis and promotes, acknowledges, demotes, or resolves it. The judge agent runs in a separate invocation with no shared state. It receives the discovery agent's output as input but is explicitly instructed to treat it as a hypothesis to challenge, not a conclusion to ratify. Status transitions are written as new documents to the events index.
```
nastasha-solomon:
```suggestion
The Triage workflow reads unassessed discovery documents and calls the Judge agent. The Judge agent independently verifies the Discovery agent's hypothesis and promotes, acknowledges, demotes, or resolves it. The Judge agent runs in a separate invocation with no shared state. It receives the Discovery agent's output as input but is explicitly instructed to treat it as a hypothesis to challenge, not a conclusion to ratify. Status transitions are written as new documents to the events index.
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+| Phase | Typical lag |
+|---|---|
+| First detection | ~10 min |
+| Discovery | ~10 min after detection |
+| New Significant Event | Sync with discovery |
+| Stale re-review | Up to ~10 min |
+
+**{{kib}} memory**
+
+Agent runs add memory pressure. Observe {{kib}} memory under realistic load. Exact memory requirements depend on your workload. Measure under realistic load before committing to a configuration.
```
nastasha-solomon, line 62: Flagging this because I'm not sure if "Agent" refers to the Judge and Discovery agents, AI agents, or LLMs.

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+| Detection workflow | {{kib}} Workflows | Cron 10m | `.rule-events` | `.significant_events-detections` |
+| Discovery workflow | {{kib}} Workflows + Agent Builder | Cron 10m | `.significant_events-detections` + KIs | `.significant_events-discoveries` |
+
+## System impact [sig-events-op-impact]
+
+The following sections describe the query load, pipeline lag, memory, and storage growth you can expect when running Significant Events. These are observable signals, not hard guarantees.
+
+**Alerting rule query load**
+
+One {{esql}} alerting rule runs per promoted query KI. Each rule fires on its own schedule. A `change_point` aggregation runs across all active rules' alert results on each detection cycle. Alert volume in `.rule-events` grows proportionally to the number of promoted rules and the rate at which they fire.
```
nastasha-solomon:
```suggestion
One {{esql}} alerting rule runs for each promoted query KI. Each rule executes on its own schedule, creating an alert when its query matches. On each detection cycle, change point aggregation runs separately for each active rule, against that rule's per-time-bucket alert counts. Document volume in `.rule-events` grows with the number of promoted rules and how often they execute.
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -160,24 +170,55 @@ Example query KI:
...
+- Enough time has elapsed since its last extraction (controlled by the **Extraction interval** setting, default 12 hours)
 
-Users can mark individual Feature KIs as false positives. The system carries those exclusions forward into future runs to prevent re-identification.
+Streams that have never been processed are always prioritized. Among remaining candidates, streams with the oldest last-completed extraction run are processed first.
 
-Because KI extraction focuses on a specific classification task, analyzing around 20 log samples to identify services, infrastructure, and dependencies, it does not require a large frontier model. A fast, cost-effective model handles this classification without multi-step reasoning.
+The continuous extraction workflow has a 34-minute timeout (one minute shorter than the 35-minute schedule, to prevent overlapping runs). If multiple runs would overlap, excess triggers are silently dropped.
+
+Toggling continuous extraction off cancels any in-flight extraction tasks and force-deletes the workflow. Re-enabling it creates a fresh workflow from the latest definition. The workflow state is not preserved between enable/disable cycles.
```
nastasha-solomon:
```suggestion
Toggling continuous extraction off cancels any in-flight extraction tasks and force-deletes the workflow. Re-enabling it creates a fresh workflow from the latest definition. The workflow state is not preserved between enable or disable cycles.
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+4. Writes matching rows to `.rule-events`.
+
+**Execution cap**: Each rule execution writes at most 1,000 alert documents. Matching events beyond this limit are dropped.
+
+## Phase 4: Discovery [sig-events-hiw-discovery]
+
+Discovery converts raw alert signals into confirmed Significant Events. It runs as three sequenced workflows (detection, discovery, and triage) coordinated by an Orchestrator workflow.
+
+- [Detection Workflow](#sig-events-hiw-detection) — `change_point` analysis per alerting rule, written to the detections index
+- [Discovery Workflow](#sig-events-hiw-discovery-agent) — discovery agent generates hypotheses, written to the discoveries index
```
nastasha-solomon:
```suggestion
- [Discovery workflow](#sig-events-hiw-discovery-agent) — Discovery agent generates hypotheses, written to the discoveries index
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+- [Triage Workflow](#sig-events-hiw-triage) — judge agent independently verifies and promotes, written to the events index
```
nastasha-solomon:
```suggestion
- [Triage workflow](#sig-events-hiw-triage) — Judge agent independently verifies and promotes, written to the events index
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+
+```
+FROM {stream},{stream}.* METADATA _id | WHERE <esql_expression>
+```
+
+The LLM operates in a reasoning loop: it calls `get_stream_features` to retrieve KIs for the stream, analyzes what entities and technologies are present, then calls `add_queries` to submit a batch of rules. This means rules are scoped to the specific services and failure modes present in each stream.
+
+## Phase 3: Rule execution [sig-events-hiw-execution]
+
+Each promoted rule runs as a {{kib}} alerting rule on a per-rule schedule. On each execution cycle, the rule:
```
nastasha-solomon:
```suggestion
Each promoted KI runs as a {{kib}} alerting rule, on its own schedule. On each execution cycle, the rule:
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+
+## What runs where [sig-events-op-components]
+
+The following table shows each pipeline component, where it runs, what triggers it, and what it reads and writes:
+
+| Component | Runs on | Trigger | Reads | Writes |
+|---|---|---|---|---|
+| Knowledge indicator (KI) feature identification | {{kib}} (Task Manager + workflow) | On-demand / continuous extraction | Stream logs | `.significant_events-knowledge_indicators` |
+| KI query generation | {{kib}} (workflow) | On-demand | Features + existing queries | Query KI assets |
+| Alerting rule execution | {{kib}} alerting → {{es}} | Per-rule schedule | Stream data via {{esql}} | `.rule-events` |
```
nastasha-solomon:
```suggestion
| Alerting rule execution | {{kib}} alerting → {{es}} | Per-rule schedule | Stream data using {{esql}} | `.rule-events` |
```

nastasha-solomon (approved):
Great job with this! Lot's of educational material in here.

A good amount of my comments are flagging inconsistencies with things like casing and terminology. I also left some some suggestions for items that didn't quite make sense to me. Feel free to take them or leave em. : )

solutions/observability/streams/significant-events/how-it-works.md
tonyghiani: This diagram is probably a bit complex and doesn't render well in the narrow page, since it has to verticalize a lot as show in the picture.

Also, it exposes a lot of information like the target indices that, IMO, are not necessary for the high-level diagram and you specify the storage detail in each sub-section later, which is good enough. Removing them for starting would potentially simplify this already.

Image

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+
+1. Builds an {{esql}} query from the stored rule parameters.
+2. Queries the stream with a lookback window of 2× the rule interval to ensure no events fall through the gap between runs.
+3. Excludes document IDs seen in previous executions.
+4. Writes matching rows to `.rule-events`.
```
tonyghiani: This will be obsolete as soon as https://github.com/elastic/kibana/pull/280339 gets merged in a few days, I left what's changed in the PR description so you can use it to update this section please

solutions/observability/streams/significant-events/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+For example: you onboard a stream, Significant Events extracts what's in it ("this stream contains a Python gRPC service called `recommendationservice` running on GKE"), generates targeted detection rules ("detect gRPC connection failures"), and runs those rules continuously. Results flow through discovery and surface as correlated Significant Events across your entire environment.
+
+## Before you get started [sig-events-prerequisites]
+
+Significant Events requires an **Enterprise license** or an active Enterprise trial.
+
+To use Significant Events you also need:
+
+- The `streams.significantEventsAvailable: true` feature flag enabled in your {{kib}} configuration.
```
tonyghiani: This is not fully true, we control that setting for a controlled rollout, so users won't know about this.

solutions/observability/streams/significant-events/index.md
```diff
@@ -0,0 +1,60 @@
...
+- The `observability:streamsSigEventsScheduledDiscoveryEnabled` {{kib}} setting enabled (required for the detection, discovery, and triage workflows).
```
tonyghiani: I wouldn't expose this directly, we programmatically enable/disable that setting, and it is transparent to the user since they cannot update it from the advanced settings page.

solutions/observability/streams/significant-events/index.md
tonyghiani: issue (blocking): "every minute" contradicts the code and the other two pages in this PR.

The default detection workflow schedule is **10 minutes**, not one minute (`DEFAULT_SIG_EVENTS_SCHEDULED_DETECTION_INTERVAL_MINUTES = 10` in `common/constants.ts`). One minute is the *change-point histogram bucket interval* (`DEFAULT_..._BUCKET_INTERVAL_MINUTES = 1`) and the *minimum configurable* schedule, but not the cadence. `how-it-works.md` and `operator-guide.md` both correctly say "cron 10m", so this line is also internally inconsistent.

```suggestion
Promoted rules run against stream data on a per-rule schedule and write matching events to the alerts index. A detection workflow runs `change_point` analysis across those results about every 10 minutes and writes detection records for statistically significant transitions.
```

solutions/observability/streams/significant-events/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+The pipeline runs in sequential phases that build on the output of the previous phase. KI extraction and rule generation prepare your streams for detection and run on your schedule. Detection and discovery run continuously in the background after you enable Significant Events.
+
+### 1. Knowledge indicator extraction [sig-events-phase-ki]
+
+An LLM analyzes log samples from your streams and extracts services, infrastructure, dependencies, and failure patterns as Knowledge Indicators (KIs). Deterministic generators run in parallel to produce dataset statistics, log patterns, and error samples.
+
+### 2. Rule generation [sig-events-phase-rules]
+
+An LLM converts KIs into {{esql}} alerting rules scoped to the services and failure modes present in each stream. Generated rules are saved as unbacked stream assets — you review and promote them before they start running.
```
tonyghiani: suggestion: this isn't the whole story, and the exception matters to operators.

High-severity MATCH queries auto-promote. In `persist_queries.ts`, `isRuleEligible` returns true when `severity_score >= HIGH_SEVERITY_THRESHOLD` (60), and those queries get rule-backed at generation time via `replaceStreamQueries` — they start running without any manual review. Only lower-severity queries are saved unbacked and wait for promotion. Since the whole point of this sentence is "nothing runs until you say so," the auto-run path is worth calling out.

```suggestion
An LLM converts KIs into {{esql}} alerting rules scoped to the services and failure modes present in each stream. Generated rules are saved as stream assets. Lower-severity rules stay unbacked until you review and promote them; high-severity detections are promoted automatically and begin running right away.
```

solutions/observability/streams/significant-events/index.md
tonyghiani: issue (blocking): "promotes, acknowledges, or dismisses" — those aren't the actual statuses.

The judge's status enum is `['open', 'closed', 'dismissed']` (`kbn-significant-events-schema/.../events/index.ts:13`, and the triage workflow's structured-output enum matches). "Promote" and "acknowledge" don't exist as judge actions. (There's a stale one-liner in `judge.ts`'s agent description that still says "promote, acknowledge, or demote" — I suspect that's where this came from, but it doesn't match the schema or the judge instructions.)

```suggestion
Two agents run in sequence. The Discovery agent reads detection records, correlates signals across rules and streams, and writes discovery records. The Judge agent independently verifies each discovery and sets its status to open, closed, or dismissed.
```

solutions/observability/streams/significant-events/how-it-works.md
tonyghiani: issue (blocking): same status-verb problem as `index.md`.

"promotes, acknowledges, demotes, or resolves it" — none of those four are real. The statuses are `open` / `closed` / `dismissed`. Suggest:

```suggestion
The triage workflow reads unassessed discovery documents and calls the judge agent. The judge agent independently verifies the discovery agent's hypothesis and sets the event status to open (page on-call), closed (confirmed settled), or dismissed (low severity and low confidence). The judge agent runs in a separate invocation with no shared state. It receives the discovery agent's output as input but is explicitly instructed to treat it as a hypothesis to challenge, not a conclusion to ratify. Status transitions are written as new documents to the events index.
```

solutions/observability/streams/significant-events/how-it-works.md
```diff
@@ -0,0 +1,153 @@
...
+
+Knowledge Indicators (KIs) are stable, evidence-backed facts about the services and infrastructure present in a stream's log data. The LLM that generates detection operates entirely on KIs. This keeps rule generation deterministic and decouples detection logic from data format specifics.
+
+### How KIs are extracted
+
+The pipeline runs up to five iterations, sampling up to 20 documents per round from entity-filtered, diverse, and random buckets. KIs found in one iteration are excluded from the next, steering each round toward undiscovered patterns. LLM analysis and deterministic code generators run in parallel; deterministic KIs always score 100 confidence. See [KI extraction](./knowledge-indicators.md#sig-events-ki-extraction) for the full extraction flow.
+
+### KI types
+
+KIs cover five types: Entity, Infrastructure, Technology, Dependency, and Schema. Entity KIs have the highest extraction priority — services and databases are identified before individual pods or hosts. See [KI types](./knowledge-indicators.md#sig-events-ki-types) for the full data model.
```
tonyghiani: thought (non-blocking): "highest extraction priority" overstates what's in the code.

The system prompt does mark Entity as the "Primary Focus" and tells the LLM not to create entities for individual pods/hosts, so the spirit is right. But there's no code-level priority ordering that runs entity extraction first; the round-robin over already-identified features is actually sorted alphabetically (`identify_inferred_features.ts`). If you want to keep this, I'd frame it as prompt guidance ("the model is instructed to prioritize services and databases over individual instances") rather than an extraction-priority guarantee.

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -75,35 +85,35 @@ The pipeline extracts:
...
-### Deterministic generators [streams-ki-generators]
+### Deterministic generators [sig-events-ki-generators]
 
 In parallel with LLM analysis, a set of deterministic code-based generators independently analyze the data to produce statistical summaries, log samples, pattern clusters, and error-specific features. Because these are computed rather than inferred, they always receive a confidence score of 100.
 
-### Merging results [streams-ki-merge]
+### Merging results [sig-events-ki-merge]
 
-LLM results and computed features are merged and deduplicated. Known KIs reuse their existing UUIDs, new discoveries get fresh ones, and user-excluded features are dropped. Surviving KIs are saved with an active status and an expiration date set seven days out.
+LLM results and computed features are merged and deduplicated. Known KIs reuse their existing UUIDs, new discoveries get fresh ones, and the pipeline drops user-excluded features. Surviving KIs are saved with an active status and an expiration date set seven days out.
```
tonyghiani: issue (blocking): expiration is 30 days, not seven — and this line contradicts your own lifecycle section.

`feature_ttl_days: 30` (`tuning_config.ts:56`), and the keep-alive rolls it forward by another 30. Your "KI lifecycle and maintenance" section correctly says "auto-expire after 30 days," and `how-it-works.md` says 30 too. This one sentence is the outlier.

```suggestion
LLM results and computed features are merged and deduplicated. Known KIs reuse their existing UUIDs, new discoveries get fresh ones, and the pipeline drops user-excluded features. Surviving KIs are saved with an active status and an expiration date set 30 days out.
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -160,24 +170,55 @@ Example query KI:
...
 
-**Grok patterns**: KIs inform Grok pattern generation when you introduce new streams.
+## Continuous extraction [sig-events-ki-continuous]
 
-## KI lifecycle and maintenance [streams-ki-lifecycle]
+When continuous extraction is enabled, a {{kib}} Workflow runs every 35 minutes and processes up to five eligible streams per run. A stream is eligible if:
 
-KIs auto-expire after 7 days if not observed in subsequent extraction runs. KIs for decommissioned services are automatically removed without manual cleanup. If a service comes back online, its KIs are re-extracted automatically.
+- It is a wired or classic stream
+- It does not match any configured exclusion glob pattern
```
tonyghiani: issue (blocking): "does not match any configured exclusion glob pattern" describes a setting that was removed.

There is no exclusion-glob mechanism. Stream scoping is done by *inclusion* via `observability:streamsSigEventsIndexPatterns`, and the old `excludedStreamPatterns` workflow input was deleted — there's even a test asserting it's gone (`continuous_onboarding_workflow.test.ts`: "does not declare the removed excludedStreamPatterns input"). Two other nuances: query streams are also eligible (not just wired/classic), and non-query streams must *match* the configured index patterns. Suggest:

```suggestion
- It is a wired, classic, or query stream that matches the configured Significant Events index patterns
- No feature identification is currently running for it
```

solutions/observability/streams/significant-events/knowledge-indicators.md
```diff
@@ -160,24 +170,55 @@ Example query KI:
...
+
+Toggling continuous extraction off cancels any in-flight extraction tasks and force-deletes the workflow. Re-enabling it creates a fresh workflow from the latest definition. The workflow state is not preserved between enable/disable cycles.
+
+### Continuous extraction settings
+
+| Setting | Default | Description |
+|---|---|---|
+| `observability:streamsContinuousKiExtractionEnabled` | `false` | Enables or disables continuous extraction |
+| `observability:streamsContinuousKiExtractionIntervalHours` | `12` | Minimum hours between extraction runs for a single stream |
+| `observability:streamsContinuousKiExtractionExcludedStreamPatterns` | `""` | Comma-separated glob patterns for streams to skip |
```
tonyghiani: issue (blocking): `observability:streamsContinuousKiExtractionExcludedStreamPatterns` doesn't exist.

I grepped the whole repo — no setting ID, no registration, nothing. The other two rows (`...Enabled` default false, `...IntervalHours` default 12) are correct. Drop the excluded-patterns row. If you want to document stream scoping, point at `observability:streamsSigEventsIndexPatterns` (the inclusion patterns on the Settings page) instead.

solutions/observability/streams/significant-events/knowledge-indicators.md
tonyghiani: suggestion: the disable/re-enable description doesn't match the current managed workflow.

For the managed continuous workflow, disabling *disables* it (sets `enabled: false`) and cancels in-flight onboarding; it does not force-delete it. Only the *legacy* pre-migration workflow gets force-deleted. Re-enabling flips `enabled: true` on the same workflow — it's `enablement: 'restorable'`, so it's not recreated from scratch and there's no "fresh workflow." Suggest:

```suggestion
Toggling continuous extraction off cancels any in-flight extraction tasks and disables the workflow. Re-enabling turns it back on; already-extracted KIs are preserved throughout.
```

solutions/observability/streams/significant-events/knowledge-indicators.md
tonyghiani: nitpick (non-blocking): true only for MATCH queries.

STATS queries can't be rule-backed yet (`canQueryBeRuleBacked` returns false for STATS; installing one throws, gated on #265778). So a STATS query KI won't become an alerting rule today. Might be worth a half-sentence so nobody's surprised when their STATS KI never fires.

solutions/observability/streams/significant-events/operator-guide.md
tonyghiani: suggestion: the table lists Detection and Discovery workflows but not Triage.

The triage/judge workflow (`system-significant-events-triage`, agent `significant_events.discovery-judge`) is a real third workflow and it's the thing that actually writes Significant Events. Add a row:

```
| Triage workflow | {{kib}} Workflows + Agent Builder | Cron 10m | `.significant_events-discoveries` | `.significant_events-events` |
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+Streams must have recent log data for KI extraction to produce useful results. Empty streams or streams with very sparse data produce weak KIs, which produce weak query KIs, which produce fewer promoted rules.
+
+**Storage growth**
+
+Significant Events writes to the following data streams:
+
+| Data stream | Written by | Growth driver |
+|---|---|---|
+| `.significant_events-detections` | Detection workflow | Append-only; one document per observed state transition per rule |
+| `.significant_events-discoveries` | Discovery agent + Judge agent | Append-only; one document per discovery state change |
```
tonyghiani: nitpick (non-blocking): the judge doesn't write to the discoveries stream.

`.significant_events-discoveries` is written by the discovery agent (`discovery_write`) and by the triage workflow's deterministic `handled` markers — but the judge agent itself only writes to `.significant_events-events` (`events_write`). "Discovery agent + Judge agent" for the discoveries row is slightly off; the judge belongs on the events row (which you already have right).

```suggestion
| `.significant_events-discoveries` | Discovery agent | Append-only; one document per discovery state change |
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+| `.significant_events-events` | Judge | Append-only; one document per Significant Event state change |
+
+The `.significant_events-*` data streams use Data Stream Lifecycle (DSL) with a default 90-day retention. You can override retention per stream using the DSL API. See [{{esql}} traceability](./how-it-works.md#sig-events-hiw-traceability) for the full index layout and traceability guidance.
```
tonyghiani: suggestion: "the `.significant_events-*` data streams use DSL with a default 90-day retention" isn't uniform.

Detections and discoveries do set `lifecycle: { data_retention: '90d' }`, but the events data stream template has *no* lifecycle block at all (`events/data_stream.ts`), so it doesn't get the 90-day default. Given events are the highest-value artifact, this is a real gap worth being precise about:

```suggestion
The `.significant_events-detections` and `.significant_events-discoveries` data streams use DSL with a default 90-day retention. (The events data stream currently has no default retention configured.) You can override retention per stream using the DSL API. See [{{esql}} traceability](./how-it-works.md#sig-events-hiw-traceability) for the full index layout and traceability guidance.
```

solutions/observability/streams/significant-events/operator-guide.md
```diff
@@ -0,0 +1,137 @@
...
+Set `observability:streamsSigEventsScheduledDiscoveryEnabled` to `false` in {{kib}} settings. This stops the detection, discovery, and triage workflows.
+
+### Stop background KI refresh only
+
+To stop continuous extraction without disabling Significant Events:
+
+1. Select **Significant Events** → **Settings**.
+2. Under **Continuous KI extraction**, turn off **Enable continuous KI extraction**.
+
+Disabling continuous extraction cancels all in-flight feature identification tasks and force-deletes the continuous extraction workflow. Already-extracted KIs are not deleted. Manually-triggered extractions continue to work.
```
tonyghiani: todo: same force-delete wording as `knowledge-indicators.md`. The managed workflow is disabled, not force-deleted.

```suggestion
Disabling continuous extraction cancels all in-flight feature identification tasks and disables the continuous extraction workflow. Already-extracted KIs are not deleted. Manually-triggered extractions continue to work.
```

explore-analyze/ai-features/agent-builder/tools/builtin-tools-reference.md
```diff
@@ -230,7 +230,7 @@ The following streams tools were available in 9.4 and were removed in 9.5:
 stack: preview 9.5
 ```
 
-Significant events tools search, create, and manage [significant events](/solutions/observability/streams/knowledge-indicators.md) and Knowledge Indicators for [Streams](/solutions/observability/streams/streams.md).
+Significant events tools search, create, and manage significant events and [Knowledge Indicators](/solutions/observability/streams/significant-events/knowledge-indicators.md) for [Streams](/solutions/observability/streams/streams.md).
 
 **Prerequisites:** Significant events require an [Enterprise license](https://www.elastic.co/subscriptions) and must be enabled for the deployment through the `observability:streamsEnableSignificantEvents` [advanced setting](kibana://reference/advanced-settings.md#kibana-general-settings).
```
tonyghiani: suggestion: while you're in this file, the setting name here is out of date.

It says Significant Events "must be enabled ... through the `observability:streamsEnableSignificantEvents` advanced setting." That setting was removed (stack 9.6 per `advanced-settings-space.yml`; there's an explicit "this flag replaces the removed ...streamsEnableSignificantEvents" comment in `common/feature_flags.ts`). Availability is now gated by the `streams.significantEventsAvailable` feature flag + Enterprise license + pricing tier. It's technically outside your primary diff, but you're editing the line right above it and it directly contradicts the new KI page.

tonyghiani (changes requested):
Code-accuracy review against the Significant Events plugin. Most of the technical detail is solid, but there are several paragraphs that contradict the code (and sometimes contradict other pages in this same PR).

solutions/observability/streams/significant-events/how-it-works.md
mdbirnstiehl: I've removed this for now, but I'll recreate outside of mermaid. formatting was still off.

tonyghiani (approved):

lcawl (approved):
I reviewed only the builtin-tools-reference.md but LGTM

PR #7517, opened 2026-07-22
[Security][Workflows][9.5 & Serverless] Document workflow graph view

nastasha-solomon:
## Summary

Adds documentation for the experimental read-only workflow graph view, which renders a workflow as a visual graph from its YAML so users can review triggers, steps, and branches without editing the graph.

Fixes https://github.com/elastic/docs-content-internal/issues/1454.

## Previews

* [Visualize a workflow](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7517/explore-analyze/workflows/authoring-techniques/visualize-workflows): New page on the read-only graph view — how to turn it on, switch between YAML and Graph, inspect a step or trigger, jump to it in the YAML editor, run a step, and what the graph shows (triggers, steps, implicit branches, parallel paths, and loops).
* [Use the YAML editor](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7517/explore-analyze/workflows/authoring-techniques/use-yaml-editor): Added a link to the new page.
* [Workflow authoring techniques](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/7517/explore-analyze/workflows/authoring-techniques): Added the new page to the list.

## Generative AI disclosure

- [x] Yes - Cursor + Claude
- [ ] No


explore-analyze/workflows/authoring-techniques/visualize-workflows.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Enable the diagram view [workflows-visualize-enable]
+
+The diagram view is off by default. Turn on the **Elastic Workflows: Experimental Features** advanced setting (`workflows:experimentalFeatures`) to use it.
+
+1. Go to the **Advanced Settings** page using the navigation menu or the [global search field](/explore-analyze/find-and-organize/find-apps-and-objects.md).
+2. Search for `workflows:experimentalFeatures` (or **Elastic Workflows: Experimental Features**).
+3. Toggle the setting on, then save your changes and reload the page.
+
+Reload the YAML editor. The **YAML ↔ Graph** toggle will appear in the editor's bottom bar.
```
bmorelli25: do you have to reload twice?

bmorelli25 (approved):
L(mostly)GTM

explore-analyze/workflows/authoring-techniques/visualize-workflows.md
nastasha-solomon: Do you? No. Can you? Yes.

```suggestion
3. Toggle the setting on, then save your changes and reload the page.

The **YAML ↔ Graph** toggle will appear in the editor's bottom bar.
```

kosabogi (approved):

tinnytintin10 (changes requested):
@nastasha-solomon one naming thing before this merges: the page currently uses three terms for the same thing. the title says "visualize a workflow", the body calls it "the diagram view", and the instructions say select graph, which is the actual ui label on the toggle.

i'd suggest standardizing on graph view in the body text since it matches what users see on the button, and it's the established term for this kind of view elsewhere (airflow's graph view, github actions' workflow graph). the "visualize a workflow" page title is fine as is, it's task oriented and searchable.

i'd avoid "visual editor" for now even though that's the internal code name. the view is read only in this release, so "editor" overpromises. we can revisit when editing from the canvas ships.

if we'd rather the user facing word be "diagram", the right fix is changing the ui toggle label in kibana while the feature is still experimental (happy to own that call), but docs and ui should land on the same word either way. wdyt @talboren @dej611?

talboren:
> @nastasha-solomon one naming thing before this merges: the page currently uses three terms for the same thing. the title says "visualize a workflow", the body calls it "the diagram view", and the instructions say select graph, which is the actual ui label on the toggle.
> 
> i'd suggest standardizing on graph view in the body text since it matches what users see on the button, and it's the established term for this kind of view elsewhere (airflow's graph view, github actions' workflow graph). the "visualize a workflow" page title is fine as is, it's task oriented and searchable.
> 
> i'd avoid "visual editor" for now even though that's the internal code name. the view is read only in this release, so "editor" overpromises. we can revisit when editing from the canvas ships.
> 
> if we'd rather the user facing word be "diagram", the right fix is changing the ui toggle label in kibana while the feature is still experimental (happy to own that call), but docs and ui should land on the same word either way. wdyt @talboren @dej611?

i think we should align on the industry standard, so i would stay with graph view.

PR #7529, opened 2026-07-23
[CCR] Clarify that version ordering includes the patch version

julius-ban:
The version requirement and the compatibility matrix are currently one sentence, so it reads like a "Yes" in the matrix is all you need. In practice they're two separate checks, and the version-ordering one goes down to the patch level, the matrix can't show that since it groups versions into ranges.

Split these into two explicit conditions and added a note for the error you actually hit when the local cluster is behind.

[WARN ][o.e.s.RestoreService] [elastic03] [[default/_ccr_cluster_b]:_latest_] failed to restore snapshot org.elasticsearch.snapshots.SnapshotException: [default:_ccr_cluster_b:_latest_/_latest_] the snapshot was created with version [9.4.3] which is higher than the version of this node [9.4.1]
... Both clusters were 9.4.x and fall in the same matrix cell (8.0–9.0 → 8.0–9.0), so the matrix gave no hint this would fail.


deploy-manage/tools/cross-cluster-replication.md
```diff
@@ -25,7 +25,14 @@ You configure {{ccr}} clusters in a uni-directional or bi-directional setup:
 * In a uni-directional configuration, one cluster contains only leader indices, and the other cluster contains only follower indices.
 * In a bi-directional configuration, each cluster contains both leader and follower indices.
 
-In a uni-directional configuration, the cluster containing follower indices must be running **the same or newer** version of {{es}} as the remote cluster. If newer, the versions must also be compatible as outlined in the following matrix.
+In a uni-directional configuration, the cluster containing follower indices (the *local* cluster) must be compatible with the cluster containing leader indices (the *remote* cluster). Both of the following conditions must be met:
```
eedugon, line 30:
```suggestion
## Version compatibility

In a uni-directional configuration, the cluster containing follower indices (the *local* cluster) must be compatible with the cluster containing leader indices (the *remote* cluster). Both of the following conditions must be met:
```

Considering the compatibility statements are growing, I'd suggest to create its own section.

@yetanothertw , what do you think of this PR?

Currently the preview would render like (showing the entire introduction):

image

I think this PR adds value but probably we need to put the new content under its own H2 section.

deploy-manage/tools/cross-cluster-replication.md
```diff
@@ -25,7 +25,14 @@ You configure {{ccr}} clusters in a uni-directional or bi-directional setup:
...
+
+* The local cluster must be running **the same or newer** version of {{es}} as the remote cluster. This comparison includes the patch version: if the remote cluster is running `X.Y.3`, the local cluster must be running `X.Y.3` or higher. `X.Y.1` is not sufficient, even though both clusters are on the same minor version.
+* If the local cluster is running a newer version, the versions must also be compatible as outlined in the following matrix.
```
eedugon:
```suggestion
* If the local cluster is running a later version than the remote cluster, the versions must also be compatible as outlined in the following matrix.
```

deploy-manage/tools/cross-cluster-replication.md
```diff
@@ -25,7 +25,14 @@ You configure {{ccr}} clusters in a uni-directional or bi-directional setup:
...
+
+::::{important}
+The following matrix groups versions into ranges, and expresses only the second condition. It does not relax the first: within a single cell of the matrix, the local cluster must still be running the same or a higher patch version than the remote cluster.
+::::
 
 :::::{dropdown} Version compatibility matrix
 :name: ccr-version-compatibility
```
eedugon, line 40: Not in the scope of this PR, but @yetanothertw , don't you think the big table is outdated? It doesn't show 9.x versions. Maybe we should create a separate issue to address that (if it's confirmed)

deploy-manage/tools/cross-cluster-replication.md
```diff
@@ -25,7 +25,14 @@ You configure {{ccr}} clusters in a uni-directional or bi-directional setup:
 * In a uni-directional configuration, one cluster contains only leader indices, and the other cluster contains only follower indices.
 * In a bi-directional configuration, each cluster contains both leader and follower indices.
 
-In a uni-directional configuration, the cluster containing follower indices must be running **the same or newer** version of {{es}} as the remote cluster. If newer, the versions must also be compatible as outlined in the following matrix.
+In a uni-directional configuration, the cluster containing follower indices (the *local* cluster) must be compatible with the cluster containing leader indices (the *remote* cluster). Both of the following conditions must be met:
+
+* The local cluster must be running **the same or newer** version of {{es}} as the remote cluster. This comparison includes the patch version: if the remote cluster is running `X.Y.3`, the local cluster must be running `X.Y.3` or higher. `X.Y.1` is not sufficient, even though both clusters are on the same minor version.
```
eedugon:
```suggestion
* The local cluster must be running **the same or newer** version of {{es}} as the remote cluster. This comparison includes the patch version: for example, if the remote cluster is running `9.4.3`, the local cluster must be running `9.4.3` or higher. `9.4.1` is not sufficient, even though both clusters are on the same minor version.
```

eedugon (commented):
Adding a few comments. @yetanothertw : I'd love to get your thoughts and review on this too.

I think the PR adds value, but maybe the original single-paragraph + table is considered enough.

If we go for this change then I'd suggest to have the versioning information in its own H2 section.

deploy-manage/tools/cross-cluster-replication.md
yetanothertw: That makes sense to me, @eedugon

deploy-manage/tools/cross-cluster-replication.md
yetanothertw:
```suggestion
* The local cluster must be running **the same or newer (later)** version of {{es}} as the remote cluster, including the patch version. For example, a remote cluster running on `9.4.3`, requires the local cluster to run on `9.4.3` or later — `9.4.1` is not sufficient.
```

deploy-manage/tools/cross-cluster-replication.md
julius-ban: I agree with that.

julius-ban:
@eedugon @yetanothertw Applied all the suggestions, please re-review. :)

deploy-manage/tools/cross-cluster-replication.md
```diff
@@ -25,7 +25,16 @@ You configure {{ccr}} clusters in a uni-directional or bi-directional setup:
...
 * In a bi-directional configuration, each cluster contains both leader and follower indices.
 
-In a uni-directional configuration, the cluster containing follower indices must be running **the same or newer** version of {{es}} as the remote cluster. If newer, the versions must also be compatible as outlined in the following matrix.
+## Version compatibility
+
+In a uni-directional configuration, the cluster containing follower indices (the *local* cluster) must be compatible with the cluster containing leader indices (the *remote* cluster). Both of the following conditions must be met:
+
+* The local cluster must be running **the same or newer** version of {{es}} as the remote cluster. This comparison includes the patch version: for example, if the remote cluster is running `9.4.3`, the local cluster must be running `9.4.3` or higher. `9.4.1` is not sufficient, even though both clusters are on the same minor version.
+* If the local cluster is running a newer version than the remote cluster, the versions must also be compatible as outlined in the following matrix.
```
yetanothertw:
```suggestion
* If the local cluster is running a later version than the remote cluster, the versions must also be compatible as outlined in the following matrix.
```

deploy-manage/tools/cross-cluster-replication.md
```diff
@@ -25,7 +25,16 @@ You configure {{ccr}} clusters in a uni-directional or bi-directional setup:
 * In a uni-directional configuration, one cluster contains only leader indices, and the other cluster contains only follower indices.
 * In a bi-directional configuration, each cluster contains both leader and follower indices.
 
-In a uni-directional configuration, the cluster containing follower indices must be running **the same or newer** version of {{es}} as the remote cluster. If newer, the versions must also be compatible as outlined in the following matrix.
+## Version compatibility
+
+In a uni-directional configuration, the cluster containing follower indices (the *local* cluster) must be compatible with the cluster containing leader indices (the *remote* cluster). Both of the following conditions must be met:
+
+* The local cluster must be running **the same or newer** version of {{es}} as the remote cluster. This comparison includes the patch version: for example, if the remote cluster is running `9.4.3`, the local cluster must be running `9.4.3` or higher. `9.4.1` is not sufficient, even though both clusters are on the same minor version.
```
yetanothertw:
```suggestion
* The local cluster must be running **the same or newer (later)** version of {{es}} as the remote cluster, including the patch version. For example, a remote cluster running on `9.4.3`, requires the local cluster to run on `9.4.3` or later — `9.4.1` is not sufficient.
```

yetanothertw:
> @eedugon @yetanothertw Applied all the suggestions, please re-review. :)

Hi @julius-ban, thanks so much for your contribution, it's really helpful. Please give me a couple of minutes to have a look and offer a few tiny suggestions for improvements.

deploy-manage/tools/cross-cluster-replication.md
```diff
@@ -25,7 +25,16 @@ You configure {{ccr}} clusters in a uni-directional or bi-directional setup:
...
+* If the local cluster is running a newer version than the remote cluster, the versions must also be compatible as outlined in the following matrix.
+
+::::{important}
+The following matrix groups versions into ranges, and expresses only the second condition. It does not relax the first: within a single cell of the matrix, the local cluster must still be running the same or a higher patch version than the remote cluster.
```
yetanothertw: I recommend rephrasing to clarify the conditions we're referring to (rather than just saying the second condition or the first). 

```suggestion
The following matrix describes the version compatibility between local and remote clusters when the local cluster is running a later version than the remote. The patch version requirement still applies: within any cell, the local cluster must still match or exceed the remote cluster's exact patch version.
```

yetanothertw (changes requested):

deploy-manage/tools/cross-cluster-replication.md
yetanothertw: Sure, Edu, let's create another issue to track that.

julius-ban:
Done. Thanks for the guidance. @yetanothertw

yetanothertw (approved):
Looks good, @julius-ban -- thank you for improving our docs! 🚀

PR #1260, opened 2025-04-25
Add OTel data stream routing page

ericywl:
This PR adds a new page under APM OTel to describe data stream routing.

Closes https://github.com/elastic/apm-server/issues/16204.

PR #7564, opened 2026-07-24
Add docs for the new workload credentials permission

yetanothertw:
## Summary

Closes [#1260](https://github.com/elastic/docs-content-internal/issues/1260) Adds details about a new role that allows a member to generate workload credentials to the [Elastic Cloud API keys](https://www.elastic.co/docs/deploy-manage/api-keys/elastic-cloud-api-keys) page.

Reviewer SMEs: @/nathalie-tran and @/s-nel + @/alxchalkias 

## Generative AI disclosure

- [x] Yes  
- [ ] No


deploy-manage/users-roles/cloud-organization/manage-users.md
```diff
@@ -55,7 +55,7 @@ To invite users to your organization:
 
 5. If desired, assign roles to the users so that they automatically get the appropriate permissions when they accept the invitation and sign in to your organization.
 
-    You can grant access to {{ech}} deployments, {{serverless-full}} projects, or connected clusters, either to all resources or scoped to specific ones. For more details, refer to [User roles and privileges](/deploy-manage/users-roles/cloud-organization/user-roles.md).
+    You can grant access to {{ech}} deployments, {{serverless-full}} projects, or connected clusters, either to all resources or scoped to specific ones. You can also allow the member you invite to manage workload credentials by assigning the [**Workload credentials owner**](/deploy-manage/users-roles/cloud-organization/user-roles.md#workload-credentials-owner) role, so they can create and manage their own [{{ecloud}} API keys](/deploy-manage/api-keys/elastic-cloud-api-keys.md).
```
alxchalkias: There's a small. nuance here that I'd like to discuss:
The reason we call the role "Workload credentials owner" and not "API Key owner/manager" is that we foresee we will be generalizing the capability once we introduce other credential types, e.g. Service accounts. So calling out "can create and manage **API Keys**" here is correct but not necessarily evergreen.

deploy-manage/users-roles/cloud-organization/user-roles.md
```diff
@@ -51,10 +51,14 @@ There are three categories of roles you can assign to users:
...
 
 * **Billing admin**: Can manage an organization’s billing details such as credit card information, subscription and invoice history. Cannot manage other organization or deployment details and properties.
 
+* $$$workload-credentials-owner$$$**Workload credentials owner**: Allows the member to manage workload credentials, such as creating, listing, and revoking their own [{{ecloud}} API keys](/deploy-manage/api-keys/elastic-cloud-api-keys.md).
+
+    This role only adds the ability to manage credentials. It does not give the member access to any deployments or projects. To create useful API keys, the member also needs [cloud resource access roles](#ec_instance_access_roles) for the deployments or projects the keys they manage should cover. An API key created with this role can only include roles that the member already holds on those resources. Organization owners can still view and revoke all API keys in the organization.
```
alxchalkias: Similar to my previous comment, we could say:
"To create useful credentials, such as API Keys,[...]"

deploy-manage/users-roles/cloud-organization/manage-users.md
yetanothertw: OK, should we call it out in the docs as well? 

If we still want to call out the API keys link, we could say something like:

```suggestion
    You can grant access to {{ech}} deployments, {{serverless-full}} projects, or connected clusters, either to all resources or scoped to specific ones. You can also allow the member you invite to manage workload credentials by assigning the [**Workload credentials owner**](/deploy-manage/users-roles/cloud-organization/user-roles.md#workload-credentials-owner) role. Today, that means they can create and manage their own [{{ecloud}} API keys](/deploy-manage/api-keys/elastic-cloud-api-keys.md).
```

Otherwise, I'm happy to drop the `can create and manage API Keys` bit of this sentence altogether.

yetanothertw: I've updated the sentence now to be slightly more generic. Wdyt?

yetanothertw:
Thanks @alxchalkias, I've updated the language to be slightly more generic wrt to workload credentials. It's ready for your review again.

deploy-manage/users-roles/cloud-organization/manage-users.md
alxchalkias: > should we call it out in the docs as well?

No, I don't think we ought to talk about future capabilities in the docs.

Your edit LGTM.

alxchalkias (approved):

deploy-manage/api-keys/elastic-cloud-api-keys.md
```diff
@@ -18,7 +18,12 @@ products:
...
 
-Only **Organization owners** can create and manage API keys. An API key is not tied to the user who created it. When creating a key, you assign it specific roles to control its access to organizational resources, including hosted deployments and serverless projects. If a user leaves the organization, the API keys they have created will still function until they expire.
+An {{ecloud}} API key belongs to the organization and is not tied to the user who created it. When a key is created, it is assigned specific roles that control its access to organizational resources. If the creator leaves the organization, the API keys they created will still function until they expire or are revoked.
+
+**Organization owners** can create and manage API keys for the organization. They can view and revoke all API keys in the organization. 
+
+Members with the **Workload credentials owner** role can create, list, and revoke only the API keys they created. For more information about the **Workload credentials owner** role, refer to [User roles and privileges](/deploy-manage/users-roles/cloud-organization/user-roles.md#workload-credentials-owner).
```
s-nel: "role can create ... only API keys they created" sounds a bit odd

```suggestion
Members with the **Workload credentials owner** role can create API keys and can list and revoke only the API keys they created. For more information about the **Workload credentials owner** role, refer to [User roles and privileges](/deploy-manage/users-roles/cloud-organization/user-roles.md#workload-credentials-owner).
```

s-nel (approved):

deploy-manage/api-keys/elastic-cloud-api-keys.md
```diff
@@ -18,7 +18,12 @@ products:
...
 
-Only **Organization owners** can create and manage API keys. An API key is not tied to the user who created it. When creating a key, you assign it specific roles to control its access to organizational resources, including hosted deployments and serverless projects. If a user leaves the organization, the API keys they have created will still function until they expire.
+An {{ecloud}} API key belongs to the organization and is not tied to the user who created it. When a key is created, it is assigned specific roles that control its access to organizational resources. If the creator leaves the organization, the API keys they created will still function until they expire or are revoked.
+
+**Organization owners** can create and manage API keys for the organization. They can view and revoke all API keys in the organization. 
+
+Members with the **Workload credentials owner** role can create API keys and can list and revoke only the API keys they created. For more information about the **Workload credentials owner** role, refer to [User roles and privileges](/deploy-manage/users-roles/cloud-organization/user-roles.md#workload-credentials-owner).
```
shainaraskas: wonder if this would read more clearly if it was in some sort of requirements/permissions section

also note that this is referring to actors as "members" but also keys with these roles can do the same activities programmatically :) I also think we prefer `users` over `members`

deploy-manage/api-keys/elastic-cloud-api-keys.md
```diff
@@ -112,6 +119,11 @@ Roles grant an API key specific privileges for your {{ecloud}} organization and
 
 You can grant a cloud API key [the same types of roles that you assign to users](/deploy-manage/users-roles/cloud-organization/user-roles.md#types-of-roles): organization-level roles, cloud resource access roles, and connected cluster access roles.
 
+The roles you can assign depend on your privileges:
+
+* **Organization owners** can assign any organization-level roles, cloud resource access roles, and connected cluster access roles.
```
shainaraskas: consider: "any" without specifying, to avoid maintaining the roles in multiple places. 

also we are missing the "manage workload credentials" option (it is not a role, first thing that is not part of a role)

deploy-manage/api-keys/elastic-cloud-api-keys.md
```diff
@@ -18,7 +18,12 @@ products:
...
+An {{ecloud}} API key belongs to the organization and is not tied to the user who created it. When a key is created, it is assigned specific roles that control its access to organizational resources. If the creator leaves the organization, the API keys they created will still function until they expire or are revoked.
+
+**Organization owners** can create and manage API keys for the organization. They can view and revoke all API keys in the organization. 
+
+Members with the **Workload credentials owner** role can create API keys and can list and revoke only the API keys they created. For more information about the **Workload credentials owner** role, refer to [User roles and privileges](/deploy-manage/users-roles/cloud-organization/user-roles.md#workload-credentials-owner).
+
 
 You can have multiple API keys for different purposes, and you can revoke them when you no longer need them. Each organization can have up to 500 active API keys.
```
shainaraskas, line 34: create procedure here (API one) should prob have a grant for workload credentials in the payload

deploy-manage/users-roles/cloud-organization/user-roles.md
```diff
@@ -51,10 +51,14 @@ There are three categories of roles you can assign to users:
...
+    Organization owners have full access to {{ecloud}} for the organization. They can manage {{ech}} deployments, {{serverless-full}} projects, and clusters linked through [Cloud Connect](/deploy-manage/cloud-connect.md). They can also manage members, organization settings, billing, and subscription details. Organization owners can create and manage all [{{ecloud}} API keys](/deploy-manage/api-keys/elastic-cloud-api-keys.md) in the organization.
 
 * **Billing admin**: Can manage an organization’s billing details such as credit card information, subscription and invoice history. Cannot manage other organization or deployment details and properties.
 
+* $$$workload-credentials-owner$$$**Workload credentials owner**: Allows the member to manage workload credentials, such as creating, listing, and revoking their own [{{ecloud}} API keys](/deploy-manage/api-keys/elastic-cloud-api-keys.md).
```
shainaraskas: is this a role? It doesn't look like it in the ui: https://www.figma.com/design/3bjglWkjCHd0qSRRsXkGP7/API-key-self-management-for-non-org-owners?node-id=8037-9225&m=dev

I also don't think this "role name" is the copy we use anywhere else.

deploy-manage/users-roles/cloud-organization/user-roles.md
```diff
@@ -51,10 +51,14 @@ There are three categories of roles you can assign to users:
...
+
+    This role only adds the ability to manage credentials. It does not give the member access to any deployments or projects. To create useful credentials, such as API keys, the member also needs [cloud resource access roles](#ec_instance_access_roles) for the deployments or projects those credentials should cover. Credentials created with this role can only include roles that the member already holds on those resources. Organization owners can still view and revoke all workload credentials in the organization.
```
shainaraskas: IMO I don't think we need to repeat this

wonder if we should front-load "limited by user's roles" (prefer user over member here), as the "this doesn't grant permissions" is secondary, especially if we consider this not exactly a role. 

one gotcha is that this is not about the roles the user _already_ holds, but instead the user's roles _at key creation time_. "already" implies "at workload credentials owner" grant time.

```suggestion
    This role only adds the ability to manage credentials. It does not give the member access to any deployments or projects. To create useful credentials, such as API keys, the member also needs [cloud resource access roles](#ec_instance_access_roles) for the deployments or projects those credentials should cover. Credentials created with this role can only include roles that the member already holds on those resources.
```

shainaraskas (commented):
some thoughts on copy and positioning for you

deploy-manage/api-keys/elastic-cloud-api-keys.md
yetanothertw: Got it, I'll move these into a required permissions section in a new commit. Thank you

deploy-manage/api-keys/elastic-cloud-api-keys.md
yetanothertw: OK, so Ive been referring to it as a role, which is not correct, so I'll update that now.

deploy-manage/users-roles/cloud-organization/user-roles.md
yetanothertw: I think i got the name from a functional design doc, but will update the name to match the UI/design instead.

yetanothertw: Also made edits to avoid referring to it as a role

deploy-manage/users-roles/cloud-organization/user-roles.md
yetanothertw: Ok, made some tweaks, hopefully it's more accurate now.

deploy-manage/api-keys/elastic-cloud-api-keys.md
yetanothertw, line 70: Hi @s-nel, could you please advise on what would be the correct way to modify this example request (or we can also create a separate one) to include this new **manage workload credentials** permission? 

Maybe there's a PR to update the [API spec](https://www.elastic.co/docs/api/doc/cloud/operation/operation-create-api-key) I could look at?

deploy-manage/api-keys/elastic-cloud-api-keys.md
yetanothertw: I've made some edits that moved content around and this comment now seems out of sync with the content. I'm resolving it to reduce confusion (not because I won't work on it 😄), and will track this suggestion here https://github.com/elastic/docs-content/pull/7564/changes#r3720207172

yetanothertw:
> some thoughts on copy and positioning for you

Thank you, @shainaraskas -- much appreciated since all these concepts are so new to me. Please have another go at it when you get the chance.

deploy-manage/api-keys/elastic-cloud-api-keys.md
s-nel: The API for creating API keys is unchanged. I don't really think we should touch this example. We could have an example for creating an API key _that can create more API keys_, but that feels unnecessary. We already had that capability with `organization-admin` role, but we didn't feel the need to call out that particular use case here

shainaraskas, line 58: here, we need to say the person can allow the key to manage workload credentials (it's not a role) :) 

(I think this is a thing)

deploy-manage/users-roles/cloud-organization/user-roles.md
```diff
@@ -51,10 +51,14 @@ There are three categories of roles you can assign to users:
...
 
 * **Billing admin**: Can manage an organization’s billing details such as credit card information, subscription and invoice history. Cannot manage other organization or deployment details and properties.
 
+* $$$manage-workload-credentials$$$**Manage workload credentials**: Allows the user to manage workload credentials, such as creating, listing, and revoking their own [{{ecloud}} API keys](/deploy-manage/api-keys/elastic-cloud-api-keys.md).
+
+    These permissions only add the ability to manage credentials and is limited by the roles that the user already holds on any resource. It does not give the user access to any deployments or projects. To create useful credentials, such as API keys, the user also needs [cloud resource access roles](#ec_instance_access_roles) for the deployments or projects those credentials should cover.
```
shainaraskas, line 60: not a role - needs to get its own heading and category in the preceding bulleted list

shainaraskas, line 60: > is limited by the roles that the user already holds on any resource. 

this is a little misleading - it is limited by the roles the user has at key creation time. reads as "limited by the roles the user holds when they are granted this new capability"

deploy-manage/users-roles/cloud-organization/user-roles.md
```diff
@@ -155,6 +159,7 @@ This list describes the scope of the different roles:
 * **Billing admin**: This role does not refer to any deployment, project, or connected cluster.
 * **Cloud resource access roles**, including **Admin**: These roles can be scoped to either all deployments or projects, or specific deployments, project types, or projects.
 * **Connected cluster access roles**: These roles can be scoped to either all connected clusters or selected clusters.
+* **Manage workload credentials**: These permissions do not refer to any deployment, project, or connected cluster. They allow the user to manage workload credentials scoped to their existing role assignments.
```
shainaraskas, line 162: this is not a role - it is a capability. it shouldn't be in this section

shainaraskas (commented):
I think we are still relying too much on the existing structure of these pages - manage workload credentials is consistently slotted into role-related sections

deploy-manage/api-keys/elastic-cloud-api-keys.md
yetanothertw: It's difficult to tell from the current designs (which cover the Invite members page/path), but I'd imagine a similar solution like toggling on an option to allow users to manage workload credentials would apply in this case.

I'll need to check this after feature freeze.

deploy-manage/users-roles/cloud-organization/user-roles.md
yetanothertw: Ok, got it now! I added a category of `Other permissions` as an umbrella term, in case more types of such permissions are added in the future.

yetanothertw: updated, thanks

deploy-manage/users-roles/cloud-organization/user-roles.md
yetanothertw: removed, thanks (it now has its own section)

deploy-manage/api-keys/elastic-cloud-api-keys.md
shainaraskas: fwiw I got asked for copy for this pathway

https://www.figma.com/design/3bjglWkjCHd0qSRRsXkGP7/API-key-self-management-for-non-org-owners?node-id=8037-9225&m=dev#1853182091

shainaraskas (approved):
lgtm, thanks <3

deploy-manage/api-keys/elastic-cloud-api-keys.md
yetanothertw: Oh I totally misread that comment. Added ✅  Thank you!

PR #7588, opened 2026-07-27
[ECE] docs: clarify route server cert behavior on proxy cert expiry - Part 1

kunisen:
## Context

When the proxy certificate expires in ECE, renewing it does not automatically renew the internal route server certificate. This can cause the administration console and Kibana to become unavailable. This is a known issue - https://github.com/elastic/ecp-traffic-team/issues/880) that is not yet fixed.

This has caused repeated support cases. The workaround was confirmed by the dev team in [this comment](https://github.com/elastic/sdh-control-plane/issues/13214#issuecomment-5089148585).

Reference ticket for tracking: https://github.com/elastic/support-tech-lead/issues/1876

## Changes

- `docs/release-notes/cloud-enterprise/known-issues.md`: Updated the existing route server cert bullet to more clearly describe the actual scenario that causes the issue.
- `deploy-manage/security/secure-your-elastic-cloud-enterprise-installation/manage-security-certificates.md`: Added a warning block at the end of the proxy certificate section, so users see it at the point where they are actively managing proxy cert renewal.

## Target doc PR tickets
Please note that there are 2 different doc PRs:
* One in proxy cert doc: This one. https://github.com/elastic/docs-content/pull/7588.
* Another in ECE known issue doc: https://github.com/elastic/cloud/pull/157422.

## Generative AI disclosure

- [x] Yes  
- [ ] No  

Claude

---

Hi @nikfot thanks for your kind help in the internal ticket: https://github.com/elastic/sdh-control-plane/issues/13214. 
May I trouble you to review this doc PR please?

Please note that there are 2 different doc PRs:
* One in proxy cert doc: This one.
* Another in ECE known issue doc: https://github.com/elastic/cloud/pull/157422.

---

Thank you!


deploy-manage/security/secure-your-elastic-cloud-enterprise-installation/manage-security-certificates.md
```diff
@@ -309,6 +309,10 @@ To add a proxy certificate from the command line:
     openssl s_client -CAfile CA_CERTIFICATE_FILENAME -showcerts -connect HOSTNAME_OR_IP:9343 < /dev/zero
     ```
 
+::::{warning}
+When the proxy certificate expires, renewing it does not automatically renew the internal route server certificate, which may cause the administration console to become unavailable. You need to rotate the route server certificate manually. For details, check [our KB article](https://ela.st/ece-route-server-cert-rotation).
```
nikfot:
```suggestion
After the proxy certificate expires, renewing it does not automatically renew the internal route server certificate, which may cause the administration console to become unavailable. You need to rotate the route server certificate manually. For details, check [our KB article](https://ela.st/ece-route-server-cert-rotation).
```

nikfot (approved):
LGTM. Added a suggestion in case you want to keep it.

deploy-manage/security/secure-your-elastic-cloud-enterprise-installation/manage-security-certificates.md
```diff
@@ -309,6 +309,10 @@ To add a proxy certificate from the command line:
     openssl s_client -CAfile CA_CERTIFICATE_FILENAME -showcerts -connect HOSTNAME_OR_IP:9343 < /dev/zero
     ```
 
+::::{warning}
+After the proxy certificate expires, renewing it does not automatically renew the internal route server certificate, which may cause the administration console to become unavailable. You need to rotate the route server certificate manually. For details, check [our KB article](https://ela.st/ece-route-server-cert-rotation).
```
eedugon: @kunisen , @nikfot : If the user renews / updates the proxy certificate BEFORE expiration. Does the issue occur?

If the answer is no, then I'd suggest to rephrase the sentence to tell users directly to renew the certs before expiration, with something like

```suggestion
Renew the proxy certificate before it expires. After the proxy certificate expires, renewing it does not automatically renew the internal route server certificate, which may cause the administration console to become unavailable. In such case, you need to rotate the route server certificate manually. For details, check [our KB article](https://ela.st/ece-route-server-cert-rotation).
```

eedugon (approved):
Looks good! Just a suggestion in case Renewing the proxy cert before expiration really makes a difference here.

In case there isn't please ignore my comment.

deploy-manage/security/secure-your-elastic-cloud-enterprise-installation/manage-security-certificates.md
nikfot: I would suggest that  the sentence implies this since it says that after it expires renewal does not happen automatically.
However, makes sense to be more explicit.

Given the discussion and taking a look at the full doc, I see that there is no other refference to certificate renewal.

Maybe we should consider adding those warnings under a dedicated `Renew Certificates` sections, adding more content to make everything explicit and clear when it gets searched.

Because right now renewal is only mentioned as a note.

kunisen: Thanks @eedugon and @nikfot 

I think I am also up for making the "hey dear user you must renew cert before expiration" in higher level instead of this proxy section, because it's not only the proxy cert :) (OTOH if we only add this to proxy one, then user may confuse that do they also need to renew other certs before expiration?, etc).

nikfot: It depends on what real life has shown. IIUC getting this issue for route-server is common. So it makes sense to address this specifically so we can easily link it to user requests.

Creating a more centralized directory depends on whether this is frequent for other certificates as well. At the end of the day, it should be apparent that certificates should be renewed before they expire, so if for the rest of the certs this is flowing nicely, lets not complicate things.

eedugon: I'm ok on adding a sub-section in the manage certificates doc saying the "obvious": that we recommend renewing certificates before they expire, and warning about the known issue at proxy layer if they let the proxy certs to expire.

That would be simple enough and would fit the scope of the "manage security certs" doc.

Thoughts?

kunisen: Thanks both again!
I will accept the suggestion per Nik's comment - _However, makes sense to be more explicit._

FWIW, I am more than happy to add a more explicit section to highlight the action to renew the cert, as well as the repercussions after cert expiration. However, I don't find very solid internal resource to describe the impact after cloud UI cert expiration. (The `adminconsole` cert is described in this external KB - https://support.elastic.co/knowledge/0071a95d). So IMHO, it would be best to have a section to say the "obvious" for all 3 certs after we get to know more about the cloud UI cert expiration. (I can imagine the traffic warning logs due to cert expiration, but we don't have external KB for this)

Hope that makes sense and thanks for your help as always!

Issue #7057, opened 2026-06-22
Kibana 9.5 - Dashboards and library APIs GA


florent-leborgne:
In Kibana 9.5, the Dashboards and _some_ library APIs are going GA. This issue tracks the work required on the documentation side to ensure these API docs are GA quality. It does not track previous work on these APIs.

The APIs involved:
- Dashboards
- Visualizations
- Markdowns
- Links
- ... TBC

Documentation for these APIs must live together. It is not yet clear that they'll be rehomed from the current location https://dashboardsapispec.kibana.dev/index.html

Task list:

**In spec repo** (shortly before the release)
- [x] Update latest spec https://github.com/elastic/dashboards-api-spec/pull/11
- [x] Update site to filter in markdowns, links, Tags and any other relevant APIs when they become available. If serverless availability differs from stack availability, make sure to mention it in the intro (can override it on the dashboards-api-spec repo side). https://github.com/elastic/dashboards-api-spec/pull/11
- [x] Link to release notes https://github.com/elastic/dashboards-api-spec/pull/11
- [x] Add 9.5 info https://github.com/elastic/dashboards-api-spec/pull/10

**In Kibana code** (along with code changes)
- [x] Confirm list of library APIs
- [x] Ensure overlay content is up to date and all APIs have a good intro description. Only the biggest ones need a detailed getting started (dashboards and visualizations)
   - [x] Dashboards https://github.com/elastic/kibana/pull/274208
   - [x] Visualizations
   - [x] Markdowns (9.6 but sooner in serverless - date TBD) -> no overlay needed
   - [x] Links (9.6 but sooner in serverless - date TBD) -> no overlay needed
   - [x] Tags -> no overlay needed
- [x] Ensure specs for all of these APIs have correct descriptions
   - [x] Dashboards
   - [x] Visualizations
   - [x] Markdowns https://github.com/elastic/kibana/pull/267785 https://github.com/elastic/dashboards-api-spec/pull/11
   - [x] Links https://github.com/elastic/kibana/pull/268783 https://github.com/elastic/dashboards-api-spec/pull/11
   - [x] Tags
- [x] Ensure specs for all of these APIs have correct examples
   - [x] Dashboards
   - [x] Visualizations
   - [x] Markdowns https://github.com/elastic/kibana/pull/267785
   - [x] Links https://github.com/elastic/kibana/pull/268783
   - [x] Tags
- [x] Confirm that smaller library APIs don't have to be trimmed down from the Bump.sh output
[...]

PR #7593, opened 2026-07-27
Fix gauge chart API example units and note API versioning

florent-leborgne:
## Summary

Follow-up to the Dashboards and Visualizations API GA work (#7057).

- **`gauge-charts.md`**: the duration format units in the API example used the old preview codes (`microseconds`/`asMilliseconds`). Updated to the GA schema unit codes (`us`/`ms`) so the request validates.
- Added a shared note (`_snippets/api-payload-version-note.md`) before every API payload example, pointing readers to the Kibana and Elastic Cloud Serverless breaking changes release notes so they can track field changes across versions.

The area chart example keeps its inline `data_view_spec` for the query annotation. On 9.5 that currently returns a 500 (elastic/kibana#280977), but this is likely a bug that will be resolved, so the example reflects the intended behavior.


vishaangelova (approved):
LGTM!

PR #7599, opened 2026-07-28
ECH Azure MP: document BYOK configuration after deployment creation

eedugon:
## Summary

- Added an admonition to the Azure Native Service page explaining that some advanced settings (deployment size, customer-managed encryption keys) are not available when creating a deployment from the Azure portal, and how to access them from the Elastic Cloud console after creation.
- Added a new sub-page (`azure-native-isv-service-byok.md`) documenting the BYOK workflow specifically for Azure Native Service users, including context on default encryption and a link to the main BYOK doc.
- Registered the new page in `toc.yml` under the Azure Native Service section.

## Test plan

- [ ] Check admonition renders correctly in the Azure Native Service page
- [ ] Check new BYOK sub-page renders and appears in the TOC under Azure Native Service
- [ ] Verify links to the main BYOK doc resolve correctly


deploy-manage/deploy/elastic-cloud/azure-native-isv-service-byok.md
```diff
@@ -0,0 +1,26 @@
+---
+applies_to:
+  deployment:
+    ess: preview
+products:
+  - id: cloud-hosted
+navigation_title: Customer-managed encryption keys
+---
+
+# Customer-managed encryption keys for Azure Native Service [azure-native-isv-service-byok]
```
shainaraskas, line 10: imo this page is not adding a lot of value, so I'd just leave the note on the other page but have it link to the BYOK config page instead (deep link to reincryption)

shainaraskas (approved):
one note about the PR structure. imo this is a lot of words for "you can encrypt after you create the deployment. go look here" :)

nothing wrong with the content as it stands so ✅ if you really prefer this strategy

eedugon:
> one note about the PR structure. imo this is a lot of words for "you can encrypt after you create the deployment. go look here" :)

I agree, my initial implementation didn't include the extra child-doc, and probably I'd go for it.

@demers-eric , @alxchalkias : what do you think? Do you think the new doc really adds value or should be enough by just improving the note in the create deployment section?

check @shainaraskas's comment at https://github.com/elastic/docs-content/pull/7599#discussion_r3668551418 too

alxchalkias (approved):

alxchalkias:
> Do you think the new doc really adds value

I'd let @demers-eric have the final say, but I don't mind the new doc under the Azure portal docs hierarchy.

demers-eric:
In this day and age, I'm not sure where I stand anymore on the spectrum of verbosity <-> conciseness. However, I do like the idea that the information architecture for our MP integration would include an item for the features that we *could* have, and therefore that someone could expect that we have, like BYOK.

LGTM! Thanks

demers-eric (approved):

PR #7600, opened 2026-07-28
Change title of ECH configure doc

eedugon:
Updating title from "Configure". I think this should be better for SEO and also the title is too generic IMO.

@shainaraskas , please share your thoughts.

## Generative AI disclosure

- [ ] Yes  
- [x] No

### DIFF
--- a/deploy-manage/deploy/elastic-cloud/configure.md
+++ b/deploy-manage/deploy/elastic-cloud/configure.md
@@ -10,7 +10,7 @@ products:
   - id: cloud-hosted
 ---
 
-# Configure
+# Configure ECH deployments
 
 You might want to change the configuration of your deployment to:
 
--- a/deploy-manage/deploy/elastic-cloud/configure.md
+++ b/deploy-manage/deploy/elastic-cloud/configure.md
@@ -6,6 +6,7 @@ mapped_pages:
 applies_to:
   deployment:
     ess: ga
+navigation_title: Configure
 products:
   - id: cloud-hosted
 ---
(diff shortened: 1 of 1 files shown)


deploy-manage/deploy/elastic-cloud/configure.md
```diff
@@ -10,7 +10,7 @@ products:
   - id: cloud-hosted
 ---
 
-# Configure
+# Configure ECH deployments
```
yetanothertw: Maybe still keep it as **Configure** in the nav title? It's easier to read/scan through the ToC

eedugon: definitely, yeah. I'll look into it

eedugon: done

yetanothertw (approved):
Sorry, Edu, I thought I'd approved it already! 🙏

deploy-manage/deploy/elastic-cloud/configure.md
```diff
@@ -6,11 +6,12 @@ mapped_pages:
 applies_to:
   deployment:
     ess: ga
+navigation_title: Configure
 products:
   - id: cloud-hosted
 ---
 
-# Configure
+# Configure ECH deployments
```
shainaraskas: consider something like this ... when I read it I was worried that people would view this as "deployments of ECH" not "deployments on ECH"


```suggestion
# Configure {{stack}} deployments on ECH
```

shainaraskas (approved):
eek, naked "Configure" is not ideal! thanks for the fix. one suggestion but I don't feel too strongly. if we don't want {{stack}} in the title maybe it can go in the first line

PR #7604, opened 2026-07-28
[Workflows] Queue concurrency strategy

natasha-moore-elastic:
## Summary
- Documents the new `queue` concurrency strategy for Elastic Workflows in 9.5, including `queue-size` and `queue-ttl`, the **Queued** execution status, and a strategy comparison.
- Adds the `queued` execution lifecycle state in the workflow anatomy reference.

Resolves elastic/docs-content-internal#1451


explore-analyze/workflows/authoring-techniques/settings.md
```diff
@@ -87,32 +87,67 @@ Use `settings.concurrency` to prevent overlapping executions from colliding. Thi
 - Workflows that mutate shared state (create or update a case, write a counter).
 - Scheduled workflows whose runtime can exceed the schedule interval.
 
+Concurrency is scoped per group within a single workflow. There is no cross-workflow queue, and queued runs start in the order they arrived (no priority).
```
h88: not precisely, concurrency is scoped by the evaluated key (group) within a Kibana space & different workflows that resolve to the same key eventually share the same group (queue). so if I understood this claim correctly the queue is actually cross-workflows...

queued runs promote FIFO with no priority - this part is correct 👍

explore-analyze/workflows/authoring-techniques/settings.md
```diff
@@ -87,32 +87,67 @@ Use `settings.concurrency` to prevent overlapping executions from colliding. Thi
...
 ```
 
-When two alerts fire on the same host within seconds, the first workflow runs; the second goes straight to the `skipped` terminal state.
+When two alerts fire on the same host within seconds, the first workflow runs. The second enters the `queued` state and appears in the execution view with a **Queued** status. When the first run reaches a terminal state, the oldest queued execution is promoted to `pending` and starts automatically. If the backlog already has `queue-size` runs, or a run waits longer than `queue-ttl`, that run ends as `skipped`.
+
+`queue` also works with `max` greater than 1: as soon as a slot frees, the next queued execution in that group starts.
 
 :::{warning}
 Concurrency keys that use `event.*` are alert-specific. `event.alerts[0].host.name` is only populated when an [alert trigger](/explore-analyze/workflows/triggers/alert-triggers.md) fires. If the same workflow is ever invoked by a manual or scheduled trigger, the key evaluates to the empty string and all executions share one group. For workflows that mix trigger types, prefer a key derived from an input (`{{ inputs.host_name }}`) or a static key (`"my-workflow"`).
```
h88: not part of the changes but caught my eyes ... empty key === no concurrency enforcement

h88 (changes requested):

explore-analyze/workflows/authoring-techniques/settings.md
natasha-moore-elastic: Fixed in [daa66cc](https://github.com/elastic/docs-content/pull/7604/commits/daa66cc210b13769a546d834b0ce770bfbf05c40).

explore-analyze/workflows/authoring-techniques/settings.md
natasha-moore-elastic: Fixed in [daa66cc](https://github.com/elastic/docs-content/pull/7604/commits/daa66cc210b13769a546d834b0ce770bfbf05c40).

h88 (approved):

explore-analyze/workflows/authoring-techniques/settings.md
```diff
@@ -87,32 +87,67 @@ Use `settings.concurrency` to prevent overlapping executions from colliding. Thi
...
 
 ```yaml
 settings:
   concurrency:
     key: "{{ event.alerts[0].host.name }}"
-    strategy: drop
+    strategy: queue
     max: 1
+    # queue-size: 100   # optional; default 100
+    # queue-ttl: 24h    # optional; default 24h
```
florent-leborgne, line 126: these render as code callouts, I didn't know!

explore-analyze/workflows/authoring-techniques/anatomy.md
florent-leborgne, line 1: this page has a few applies-switch that list older versions first, they should be inverted in a follow-up

explore-analyze/workflows/authoring-techniques/settings.md
florent-leborgne, line 1: Nit: we should probably avoid letting too many `;` and em dashes even if it's not really a big deal, these 2 AI signals in particular are more and more disliked by readers 😅

florent-leborgne (approved):
LGTM!

explore-analyze/workflows/authoring-techniques/anatomy.md
natasha-moore-elastic: Fixed in [43cf05a](https://github.com/elastic/docs-content/pull/7604/commits/43cf05ac0067afd7a3c6e0133d51fb10a27441ca).

explore-analyze/workflows/authoring-techniques/settings.md
natasha-moore-elastic: Fair! I didn't find too many on this page, but I rephrased to avoid the one this PR introduced 👍

PR #7615, opened 2026-07-29
Docs: Clarify index downgrade limitations in snapshot compatibility matrix

jiwan67:
## Summary
This PR resolves a logical contradiction in the Snapshot Version Compatibility matrix in the `snapshot-and-restore.md` documentation.

## Issue Details
The text below the compatibility table explicitly states:
> "You can’t restore an index to an earlier version of Elasticsearch. For example, you can’t restore an index created in 8.18.0 to a cluster running 8.15.0."

However, because minor versions are grouped together in the table (e.g., `8.0-8.19`), the `✅` in the matrix technically implies that an 8.19 index could be restored to an 8.0 cluster, which contradicts the rule. 

## Changes
- Added a new footnote (`^3^`) to the `8.0-8.19` row.
- Provided a footnote description explicitly stating that restoring to an earlier minor version within the same major version is not supported.

This ensures the table remains compact while preventing user confusion regarding minor version downgrades.


eedugon:
@jiwan67 , the diff of the PR looks like a full file rewrite because the line endings were converted to CRLF. Could you normalize the file back to LF (as on main), then commit and push again?

Then we can review the footnote update. I agree with the contradiction you've shared but I'd suggest something different.

jiwan67:
The change has been completed.

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -187,6 +187,8 @@ The following table shows whether an index can be restored to a given cluster ve
 
 ^2^ $$$footnote-2$$$ Supported with [searchable snapshots](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md).
 
+^3^ $$$footnote-3$$$ Restoring to a cluster within the same major version is only supported if the target cluster's minor version is greater than or equal to the index creation minor version.
```
eedugon: Isn't that covered in the main statement we have previously? `You can’t restore a snapshot to an earlier version of {{es}}. For example, you can’t restore a snapshot taken in 7.6.0 to a cluster running 7.5.0.`

Also the next sentence says: `You can’t restore an index to an earlier version of {{es}}. For example, you can’t restore an index created in 8.18.0 to a cluster running 8.15.0.`

I think the footnote here would feel repetitive, and shouldn't be associated with the 8.0-8.19 line only, but with the entire table. So I don't think a footnote is what we need.

What we need is something to ensure the reader understand that the table is not contradictory to the standard snapshot compatibility statement that already says that you can't restore anything to an earlier version.

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -187,6 +187,8 @@ The following table shows whether an index can be restored to a given cluster ve
...
+
 You can’t restore an index to an earlier version of {{es}}. For example, you can’t restore an index created in 8.18.0 to a cluster running 8.15.0.
```
eedugon: Not sure if this is ideal, but just a small improvement suggestion. @yetanothertw , feel free to share your thoughts or an alternative.

```suggestion
In the table above, a ✅ within a version range does not mean you can restore an index to any cluster version in that range, because a snapshot can be restored only to the same or a newer {{es}} version. For example, you can’t restore an index created in 8.18.0 to a cluster running 8.15.0.
```

deploy-manage/tools/snapshot-and-restore.md
eedugon:
```suggestion
```

Suggesting removing this and replacing it by a different clarification.

deploy-manage/tools/snapshot-and-restore.md
```diff
@@ -178,7 +178,7 @@ The following table shows whether an index can be restored to a given cluster ve
 | 6.8                   | ✅ ^1^                                   | ✅ ^1^   | ❌      | ✅       | ❌      | ✅  |
 | 7.0–7.1               | ✅ ^1, 2^                                | ✅       | ✅      | ✅       | ✅      | ❌  |
 | 7.2–7.17              | ✅ ^1, 2^                                | ✅       | ✅      | ✅       | ❌      | ❌  |
-| 8.0–8.19              | ✅                                       | ✅       | ✅      | ❌       | ❌      | ❌  |
+| 8.0–8.19              | ✅                                       | ✅ ^3^   | ✅ ^3^  | ❌       | ❌      | ❌  |
```
eedugon:
```suggestion
| 8.0–8.19              | ✅                                       | ✅       | ✅       | ❌       | ❌      | ❌  |
```

Suggeting to removing the footnote

eedugon (commented):
I'm suggesting to remove the footnote and changing it by a different clarification right after the table.

deploy-manage/tools/snapshot-and-restore.md
jiwan67: Thanks for the clarification. I completely agree that adding a footnote would be repetitive and miss the broader scope of the rule.

To ensure readers don't misinterpret the ✅ marks without being repetitive, I've updated the PR by adding a clear disclaimer right before the table. This sets the correct context before the user even looks at the compatibility matrix.

Let me know if the phrasing looks good to you

jiwan67:
The changes have been completed.

PR #7679, opened 2026-07-31
docs: note TS_INFO/METRICS_INFO exclusion from grouped results in Discover (#7625)

### LINKED ISSUE #7625: Expanding grouped results now works for ES|QL time series (TS) queries in Discover
## Summary

In Discover&#x27;s grouped results layout (a STATS BY query with a single grouping field), expanding a group to see its underlying documents previously failed when the query used the TS (time series) source command, because the generated subquery reused TS with functions it doesn&#x27;t support outside STATS. The subquery now converts TS to FROM so expansion works. Additionally, queries containing TS_INFO or METRICS_INFO no longer offer expandable groups, since those synthetic metric-metadata rows have no real documents to drill into.

**Why this needs docs:** Why this needs docs: the &#x27;View grouped results from a STATS query&#x27; section documents exactly which queries activate the grouped layout, and it currently says nothing about TS-based queries or the TS_INFO/METRICS_INFO exclusion, so readers can&#x27;t tell whether time series queries support group expansion.

## Resources

- PR [#280580](https://github.com/elastic/kibana/pull/280580) — [Discover] [ES|QL] Fix cascade drilldown for TS queries by converting to FROM
- Also applies to: ES|QL in Kibana

## Availability

| Channel | Details |
|---------|---------|
| **Stack** | v9.5.0, v9.5.1, v9.6.0 |
| **Serverless** | Jul 27–Jul 31 |
| **Feature status** | preview |
| **Feature flag** | None — active by default |

---
*Created with [Docs Quest Scanner](https://github.com/florent-leborgne/docs-quest-scanner) by @florent-leborgne*

## Suggested edits

[Using ES|QL](https://www.elastic.co/docs/explore-analyze/discover/try-esql) > View grouped results from a STATS query
- **What the docs say:** The grouped layout activates when the BY clause contains a single field reference or a single CATEGORIZE call. Other grouping functions like BUCKET or TBUCKET, and queries that group by more than one field (for example, BY clientip, extension), keep the standard flat results table.
- **What to add:** The activation rules paragraph could clarify time series behavior: queries starting with the TS source command support the grouped layout and group expansion (Discover converts TS to FROM internally when fetching the underlying documents), while queries using TS_INFO or METRICS_INFO keep the flat table because their synthetic rows have no underlying documents to expand. A tagged sentence or list item in the existing paragraph is enough; no new section needed. Applies from 9.5.0 (technical preview) and in serverless.

florent-leborgne:
## Summary

This PR addresses #7625 with the following change:

- **`explore-analyze/discover/try-esql.md`**: In the "View grouped results from a STATS query" section, extend the activation-rules paragraph to note that queries using `TS_INFO` or `METRICS_INFO` keep the flat results table, because those commands return synthetic metric-metadata rows with no underlying documents to expand.

### Scope and tagging decisions

The linked implementation PR ([elastic/kibana#280580](https://github.com/elastic/kibana/pull/280580)) is a `release_note:fix` with two effects:

1. **`TS` group expansion fix** — before 9.5, the grouped layout already appeared for `TS` queries; only *expanding* a group failed. This PR fixes expansion, after which a `TS … | STATS … BY <field>` query behaves like any other single-field `STATS BY` query already documented here. No lasting user-facing distinction, so this stays in release notes only — **not documented**.
2. **`TS_INFO` / `METRICS_INFO` exclusion** — documented, but **not tagged** `applies_to`. This is inherent behavior: those commands produce metadata rows, not documents, so there is nothing to drill into. The pre-9.5 state (a synthetic-row query showing an expandable layout) was the bug; we don't document buggy states, and bugfixes don't warrant a version badge. It's the same class of statement as the existing untagged exclusions (`BUCKET`/`TBUCKET`, multi-field `BY`), so it folds into that paragraph rather than standing alone.

Verified against the implementation at HEAD: `hasTimeseriesInfoCommand` (matches both `TS_INFO` and `METRICS_INFO`) makes `getESQLStatsQueryMeta` return no cascade metadata; the detection function at the PR's base commit confirms `TS` queries already activated the layout pre-fix. Both linked reference pages (`ts-info`, `metrics-info`) are live in the ES|QL commands reference.

## Resolves

Closes #7625

---

> **AI-generated draft** — created with Claude Opus 4.8.
> Review all generated content for factual accuracy before merging.


mdbirnstiehl (approved):
LGTM!

PR #7685, opened 2026-07-31
Adds near-real-time knn parameter and vectordb_document index mode to conceptual docs

kosabogi:
## Summary
This PR adds documentation for the `vectordb_document` index mode and `near-real-time` knn parameter that will be introduced in the 9.5 release.

## Generative AI disclosure

- [x] Yes  
- [ ] No  

Cursor agents


solutions/search/vector/knn.md
```diff
@@ -179,6 +179,34 @@ PUT image-index
 }
 ```
 
+### Near-real-time kNN [near-real-time-knn]
+```{applies_to}
+stack: ga 9.5
+```
+
+Use the `near_real_time` parameter in the top-level `knn` object to control whether approximate kNN search includes vectors from data that was indexed recently but not yet fully prepared for search.
```
leemthompo: an intro sentence that explains the default behavior might be good for setting the table here :)

"but not yet fully prepared for search" doesn't read super precise to me, can we be more explicit about why the vectors aren't ready for search?

or at least be consistent everywhere and just say "not yet optimized for search"

solutions/search/vector/knn.md
```diff
@@ -179,6 +179,34 @@ PUT image-index
...
+
+When `near_real_time` is `true`, kNN search can include vectors from newly indexed data, even if {{es}} has not finished optimizing them yet. Results may include documents you indexed seconds ago, but search can be slower while optimization is still in progress.
```
leemthompo: "has not finished optimizing them yet"

same comment as first one, let's just explain once, clearly at the outset why vectors aren't available immediately for search and briefly what these optimizations are

if that explanation isn't useful or available, fine to omit, but using same term everywhere simplifies reading

solutions/search/vector/knn.md
```diff
@@ -179,6 +179,34 @@ PUT image-index
...
+
+When `near_real_time` is `false`, kNN search skips vectors that are not yet fully optimized. Indexing runs faster on large vector workloads, but documents you just indexed may not appear in kNN results until {{es}} finishes preparing them in the background.
```
leemthompo: again here we talk about optimizing and preparing in the background, but a single clean explanation at the start would avoid referring to the same process in 3 different ways :)

solutions/search/vector/knn.md
```diff
@@ -179,6 +179,34 @@ PUT image-index
...
+
+The default value depends on the index mode. For indices that use the `standard` mode, `near_real_time` defaults to `true`. For indices that use `vectordb_document` mode, it defaults to `false`.
```
leemthompo: " For indices that use `vectordb_document` mode"
Consider linking to the dense-vector index modes reference or the new vector index mode section?

leemthompo (commented):
This section duplicates the parameter reference (true/false behavior, defaults, code example). Consider focusing on when and why in the introduction — the reference page already covers the _what_.

nit: I noticed that the same background process is described three different ways across this section: "prepared for search" , "optimizing" , and "preparing" . I'd suggest naming it once up front and reusing it consistently. I'd lean towards "optimize" as the consistent term.

For example:

```
Use the `near_real_time` parameter in the top-level `knn` object to control whether approximate kNN search includes vectors that have not yet been optimized for search.
```

We need to ensure cross-links with the index mode reference page too.

solutions/search/vector/knn.md
```diff
@@ -179,6 +179,34 @@ PUT image-index
 }
 ```
 
+### Near-real-time kNN [near-real-time-knn]
+```{applies_to}
+stack: ga 9.5
```
leemthompo, line 184: does it merit an explanation why this isn't tagged for serverless?

kosabogi: As far as I know, our convention is to call out deployment types and version availability only when they differ from what is stated at the top of the main page.

Since Serverless is already listed as supported on this page, I don’t think we need to mention it again here. Otherwise, we would probably need to add the same callout in other sections as well.

For example, [this section](https://www.elastic.co/docs/solutions/search/vector/knn#the-on_disk_rescore-option) is specifically tagged as Serverless: unavailable, so my understanding is that only exceptions like this need a Serverless tag.

kosabogi:
> This section duplicates the parameter reference (true/false behavior, defaults, code example). Consider focusing on when and why in the introduction — the reference page already covers the _what_.
> 
> nit: I noticed that the same background process is described three different ways across this section: "prepared for search" , "optimizing" , and "preparing" . I'd suggest naming it once up front and reusing it consistently. I'd lean towards "optimize" as the consistent term.
> 
> For example:
> 
> ```
> Use the `near_real_time` parameter in the top-level `knn` object to control whether approximate kNN search includes vectors that have not yet been optimized for search.
> ```
> 
> We need to ensure cross-links with the index mode reference page too.

Thanks for your review, @leemthompo !
I've updated that section:

- Used "optimized" consistently for this process
- Kept the true/false explanations, but focused them more on the why
- Added cross-links

solutions/search/vector/knn.md
leemthompo: true! and I've been following this myself, but for some reason my brain thought otherwise when reviewing 😄 

not for this PR, but I don't think all docs have been taking the same approach 😢  (I think this is because the reader experience is slightly confusing about the cascading nature from page-level)

leemthompo (approved):
👍

PR #7699, opened 2026-08-01
Recommend synthetic _source before disabling _source entirely

ash9146:
## Summary

The `mapping-source-field.md` reference page lists **synthetic `_source`**
as the first option to consider for reducing `_source` disk usage,
before disabling `_source` completely:

> If disk usage is important to you, then consider the following options:
> 1. Using synthetic `_source`...
> 2. Disabling the `_source` field completely.

This production-guidance page (`disk-usage.md`) only covered the second,
more disruptive option under "Disable `_source`" - reindex, update, and
highlighting stop working once `_source` is disabled. Synthetic
`_source` avoids that tradeoff for most common field types by
reconstructing `_source` at retrieval time from `doc_values` instead of
storing it, so reindex and update continue to work normally.

## What's added

A new "Reduce `_source` storage overhead" section, placed right before
the existing "Disable `_source`" section, that:
- Recommends synthetic `_source` as the option to consider first
- Notes it requires an [Enterprise subscription](https://www.elastic.co/pricing)
  (per elastic.co/subscriptions, synthetic `_source` has been an
  Enterprise feature since 8.17)
- Notes the reconstructed `_source` can differ slightly from the
  original document (field order, array handling), linking to the
  relevant reference section for details

This isn't presented as a strictly better replacement for the existing
"Disable `_source`" section - both stay, since synthetic `_source`
requires a subscription and has its own caveats, so fully disabling
`_source` remains a valid option for some readers.

## Testing
- Docs build passes
- New section renders correctly between "Watch your shard size" and
  "Disable `_source`"


deploy-manage/production-guidance/optimize-performance/disk-usage.md
shainaraskas, line 81: I like this change, but I think it would be nice if the options were presented as one narrative. you're still disabling `_source`, but have an escape hatch of keeping access to source data using synthetic `_source`.

We also don't need to state subscription levels changing between 8.17 and later because these docs only apply to 9.x. we also defer specific subscription requirements to our subscription docs generally ([example](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/mapping-source-field#synthetic-source)). 

let me know what you think.


```suggestion
## Reduce `_source` storage overhead [disable-source]

The [`_source`](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md) field stores the original JSON body of each document. If disk usage is a concern, you have two options to reduce or eliminate this overhead.

* **Use [synthetic `_source`](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#synthetic-source)**: Synthetic `_source` reconstructs the field at retrieval time instead of storing it on disk, with no additional storage overhead for most common field types. Choose this option if you need to update documents in place, reindex data, or highlight search results.

  Synthetic `_source` requires a [subscription](https://www.elastic.co/pricing). The reconstructed document might differ slightly from the original. Refer to [Synthetic `_source` modifications](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#synthetic-source-modifications) for details.

* **Disable `_source` entirely**: Choose this option if you don't need access to the original document body at all. This eliminates storage overhead completely, but APIs that depend on `_source`, including update, reindex, and highlight, stop working.

```

ash9146: Thanks for the suggestion, @shainaraskas! That makes the narrative much cleaner and aligns well with the docs standard. Applied!

deploy-manage/production-guidance/optimize-performance/disk-usage.md
```diff
@@ -69,9 +69,24 @@ Keep in mind that large shard sizes come with drawbacks, such as long full recov
...
+
+Synthetic `_source` requires an [Enterprise subscription](https://www.elastic.co/pricing) (from {{es}} 8.17 onward), and the reconstructed `_source` may differ slightly from the original document (field order, array handling) — see [synthetic `_source` modifications](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#synthetic-source-modifications) for details.
+
 ## Disable `_source` [disable-source]
 
 The [`_source`](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md) field stores the original JSON body of the document. If you don't need access to it you can disable it. However, APIs that needs access to `_source` such as update, highlight and reindex won't work.
```
shainaraskas: looks like some duplicate content crept in
```suggestion
```

shainaraskas (approved):

deploy-manage/production-guidance/optimize-performance/disk-usage.md
```diff
@@ -69,9 +69,15 @@ Keep in mind that large shard sizes come with drawbacks, such as long full recov
...
+
+* **Use [synthetic `_source`](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#synthetic-source)**: Synthetic `_source` reconstructs the field at retrieval time instead of storing it on disk, with no additional storage overhead for most common field types. Choose this option if you need to update documents in place, reindex data, or highlight search results.
+
+  Synthetic `_source` requires a [subscription](https://www.elastic.co/pricing). The reconstructed document might differ slightly from the original. Refer to [Synthetic `_source` modifications](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#synthetic-source-modifications) for details.
+
+* **Disable `_source` entirely**: Choose this option if you don't need access to the original document body at all. This eliminates storage overhead completely, but APIs that depend on `_source`, including update, reindex, and highlight, stop working.
```
shainaraskas:
```suggestion
* **[Disable `_source` entirely](elasticsearch://reference/elasticsearch/mapping-reference/mapping-source-field.md#disable-source-field)**: Choose this option if you don't need access to the original document body at all. This eliminates storage overhead completely, but APIs that depend on `_source`, including update, reindex, and highlight, stop working.
```

shainaraskas:
thanks for your contribution, @ash9146 - hope you don't mind me merging these :)

ash9146:
Thank you so much for the thorough reviews and support! Happy to contribute! 😊

PR #7790, opened 2026-08-05
Note the storage tradeoff of copy_to and how to avoid it

ash9146:
## Summary

The "Search as few fields as possible" section shows `copy_to` purely
as a search-speed optimization, with no mention that the source fields
(`name`, `plot`) remain separately indexed alongside the new combined
field. As written, the example adds roughly a third field's worth of
index storage rather than saving any - this tradeoff wasn't documented
anywhere.

## Context

`size-shards.md` already frames `copy_to` as a way to *reduce*
per-field mapping overhead:

> If a collection of fields are typically used together, consider
> using `copy_to` to consolidate them at index time.

That framing only holds if indexing is disabled on the source fields
once they're only needed via the combined field (`"index": false`).
This connection wasn't made from `search-speed.md`, so a reader
following this guide as-is would take on the storage overhead without
knowing there's a way to avoid it.

## What's added

A short note right after the `copy_to` example that:
- Clarifies that `name` and `plot` are still indexed individually in
  this example, in addition to `name_and_plot`
- Explains how to avoid that overhead (`"index": false` on the source
  fields, if they don't need to be searched individually)
- Cross-links to `size-shards.md` for more detail on `copy_to` as a
  mapping-overhead reduction technique

## Testing
- Docs build passes
- New note renders correctly between the `copy_to` example and
  "Pre-index data"


deploy-manage/production-guidance/optimize-performance/search-speed.md
```diff
@@ -117,6 +117,8 @@ PUT movies
 }
 ```
 
+Note that in the example above, both `name` and `plot` are still indexed individually in addition to `name_and_plot`, so this adds roughly a third field's worth of index storage. If you only need `name` and `plot` to be searchable through the combined field (for example, you never query them individually), you can avoid this overhead by disabling indexing on the source fields with `"index": false`, keeping only `name_and_plot` indexed for search. See [Size your shards](/deploy-manage/production-guidance/optimize-performance/size-shards.md) for more on `copy_to` as a way to reduce per-field mapping overhead.
```
georgewallace:
```suggestion
:::note
In the previous example, `name` and `plot` are still indexed individually alongside `name_and_plot`, which adds storage overhead for each source field. If you don't need to search those fields individually, you can avoid this by setting `"index": false` on them. See [Size your shards](/deploy-manage/production-guidance/optimize-performance/size-shards.md) for more on using `copy_to` to reduce per-field mapping overhead.
:::
```

georgewallace (changes requested):
We actually have a directive for notes so they render as such on the page, also made some suggestions to the verbiage.

deploy-manage/production-guidance/optimize-performance/search-speed.md
ash9146: Thanks for the review! I've updated the note to use the :::note directive and applied the suggested wording changes.

georgewallace (approved):
LGTM, thanks for making the improvements

ash9146:
Thank you @georgewallace for the review and approval!

PR #7851, opened 2026-08-07
Document emptyDir impact on ECK Desired Nodes and ResourcesAwareManagement

barkbay:
## Summary
Documents an additional implication of using `emptyDir` (or another non-PVC volume) for Elasticsearch `elasticsearch-data` on ECK: besides data loss risk, ECK cannot compute storage for the Desired Nodes API, so `ResourcesAwareManagement` becomes `False` and Desired Nodes is cleared when Elasticsearch is reachable.

Operators can misread `ResourcesAwareManagement=False` as ECK stopping resource reconciliation. The current emptyDir section only warns about data loss.

Related: elastic/cloud-on-k8s#9669 (condition docs), and the follow-up discussion on deriving Desired Nodes storage from `ephemeral-storage` / `emptyDir.sizeLimit`.

## Generative AI disclosure

- [x] Yes  
- [ ] No


pkoutsovasilis (approved):

deploy-manage/deploy/cloud-on-k8s/volume-claim-templates.md
```diff
@@ -86,3 +86,7 @@ spec:
         - name: elasticsearch-data
           emptyDir: {}
 ```
+
+::::{note}
+Besides the risk of data loss, using `emptyDir` (or another non-PVC volume) for `elasticsearch-data` means ECK cannot determine storage capacity for the {{es}} [Desired Nodes]({{es-apis}}group/endpoint-cluster) API. The Elasticsearch resource then reports `status.conditions` type `ResourcesAwareManagement` as `False`, and ECK attempts to clear Desired Nodes when {{es}} is reachable so an incomplete topology is not kept. That is not specific to mixed PVC and `emptyDir` NodeSets: if any NodeSet lacks a computable storage value from a `volumeClaimTemplates` PVC, Desired Nodes cannot be published for the cluster. ECK still reconciles the Elasticsearch resource (StatefulSets, Pods, and CPU/memory settings) as usual.
```
barkbay: Removed link to the Elasticsearch Desired Nodes API: that page is no longer available under current/9.x docs (redirects to a generic Cluster API group).

deploy-manage/deploy/cloud-on-k8s/volume-claim-templates.md
```diff
@@ -86,3 +86,9 @@ spec:
         - name: elasticsearch-data
           emptyDir: {}
 ```
+
+::::{note}
+Besides the risk of data loss, using `emptyDir` (or another non-PVC volume) for `elasticsearch-data` means ECK cannot determine storage capacity for every node. The Elasticsearch resource then reports `status.conditions` type `ResourcesAwareManagement` as `False`.
```
pebrc: Two accuracy points in this sentence pair.

**Version scope.** The condition is only ever reported on {{es}} 8.3 and later — `driver/stateful/nodes.go:96` wraps the whole desired-nodes step in `if esClient.IsDesiredNodesSupported() {`, and `client/desired_nodes.go:14` sets `desiredNodesMinVersion = version.MinFor(8, 3, 0)`. On anything older the condition is absent, not `False`. Since the page is `applies_to: deployment: eck: all` and ECK still technically supports 7.x, this needs a qualifier.

**"for every node"** reads ambiguously — it could mean "ECK fails for each node individually" or "ECK can't cover all nodes." And `status.conditions` type X as `False` is a field path used as a noun; naming the condition directly is easier to read.

```suggestion
Besides the risk of data loss, `emptyDir` (or any volume that isn't a `volumeClaimTemplates` PVC) prevents ECK from determining the storage capacity of the affected nodes. On {{es}} 8.3 and later, the `Elasticsearch` resource then reports the `ResourcesAwareManagement` condition as `False`.
```

deploy-manage/deploy/cloud-on-k8s/volume-claim-templates.md
```diff
@@ -86,3 +86,9 @@ spec:
...
+
+`ResourcesAwareManagement` indicates whether ECK could compute CPU, memory, and storage for all expected nodes so it can tell {{es}} the intended cluster topology. When the condition is `False`, that topology information is not published. Without it, {{es}} may make less optimal topology-aware decisions. This is not specific to mixed PVC and `emptyDir` NodeSets: if any NodeSet lacks a computable storage value from a `volumeClaimTemplates` PVC, the condition is `False` for the cluster. ECK still reconciles the Elasticsearch resource (StatefulSets, Pods, and CPU/memory settings) as usual.
```
pebrc: This paragraph has one substantive gap and two smaller issues.

**Storage isn't the only input, and the omitted one is the more common cause.** ECK needs three things to compute a node's resources, and it can fail on any of them:

- a CPU request or limit — `nodespec/desired_nodes.go:100`, `addReason("no CPU request or limit set")`
- a memory limit, equal to the request if a request is set — `:111`
- storage derivable from a `volumeClaimTemplates` PVC — the case this note describes

And ECK sets no CPU request or limit by default: `nodespec/defaults.go:34`, `// No CPU requirement is set by default.` So `ResourcesAwareManagement` is `False` out of the box on a cluster that uses PVCs for every nodeSet. As written, this paragraph tells the reader storage is the general cause; someone who removes `emptyDir` to fix the condition will still see `False`.

**"that topology information is not published" understates the effect.** `driver/stateful/desired_nodes.go:59` calls `esClient.DeleteDesiredNodes(ctx)` when {{es}} is reachable, so topology ECK published earlier is actively removed. That matters for day-two: converting an existing PVC-backed cluster to `emptyDir` doesn't just stop future updates.

**"This is not specific to mixed PVC and `emptyDir` NodeSets"** negates a scenario the page never introduces, so it reads as an answer to a review thread rather than to a reader's question. A positive statement of the cluster-wide scope carries the same information.

```suggestion
`ResourcesAwareManagement` indicates whether ECK could compute CPU, memory, and storage for all expected nodes so it can tell {{es}} the intended cluster topology. When the condition is `False`, ECK stops publishing that topology and removes any topology it published earlier, so {{es}} can no longer take the intended topology into account. The condition is cluster-wide: a single nodeSet without a data PVC sets it to `False` for the whole cluster. Storage isn't the only input — ECK also needs a CPU request or limit and a memory limit on the {{es}} container, and it sets no CPU request or limit by default, so the condition can be `False` even when every nodeSet uses a PVC. ECK still reconciles the `Elasticsearch` resource (StatefulSets, Pods, and CPU and memory settings) as usual.
```

deploy-manage/deploy/cloud-on-k8s/volume-claim-templates.md
```diff
@@ -86,3 +86,9 @@ spec:
...
+::::
```
pebrc: *Admonition weight.* Around 90 words over two paragraphs inside a `::::{note}` is heavy, and the repo guidance is to keep content in prose where it reads naturally. The explanatory paragraph reads better as body prose after the code block, with at most a one-sentence note left for the genuinely surprising part — that the condition can be `False` even with PVCs everywhere, because no CPU request is set by default. Rough shape:

> *(prose)* Besides the risk of data loss, `emptyDir` … prevents ECK from determining the storage capacity of the affected nodes. …
>
> *(prose)* On {{es}} 8.3 and later, the `Elasticsearch` resource reports this as the `ResourcesAwareManagement` condition: *(kubectl snippet)* …
>
> *(note)* Storage isn't the only input. ECK also needs a CPU request or limit and a memory limit on the {{es}} container. Because ECK sets no CPU request or limit by default, `ResourcesAwareManagement` can be `False` even when every nodeSet uses a PVC.

If you'd rather keep the single-note structure, the two suggestions above are self-contained and fix the accuracy issues on their own.

deploy-manage/deploy/cloud-on-k8s/volume-claim-templates.md
pebrc, line 91: The condition is only reported on {{es}} 8.3 and later — `driver/stateful/nodes.go:96` wraps the desired-nodes step in `if esClient.IsDesiredNodesSupported() {`, and `client/desired_nodes.go:14` sets `desiredNodesMinVersion = version.MinFor(8, 3, 0)`. On older clusters the condition is absent rather than `False`, so the unqualified "then reports" isn't accurate for every version this page covers.

Also, "for every node" reads two ways ("fails for each node" vs. "can't cover all nodes"), and naming the condition directly is easier to read than `status.conditions` type X.

```suggestion
Besides the risk of data loss, `emptyDir` (or any volume that isn't a `volumeClaimTemplates` PVC) prevents ECK from determining the storage capacity of the affected nodes. On {{es}} 8.3 and later, the `Elasticsearch` resource then reports the `ResourcesAwareManagement` condition as `False`.
```

deploy-manage/deploy/cloud-on-k8s/volume-claim-templates.md
pebrc, line 93: Two changes, both net-shorter — I don't think this note should grow.

**Drop the "not specific to" clause.** It negates a scenario the page never introduces, so it reads as an answer to a review thread rather than to a reader's question. More importantly, the general rule it states is incomplete: ECK also needs a CPU request or limit (`nodespec/desired_nodes.go:100`) and a memory limit (`:111`), and it sets no CPU request or limit by default (`nodespec/defaults.go:34`), so the condition is `False` out of the box even when every nodeSet uses a PVC. Rather than list all three inputs here — that's reference material for elastic/cloud-on-k8s#9669 — I'd keep the cluster-wide fact but scope it to `emptyDir`, which is what a reader on this page needs.

**"that topology information is not published" → also removed.** `driver/stateful/desired_nodes.go:59` calls `esClient.DeleteDesiredNodes(ctx)` when {{es}} is reachable, so topology published earlier is deleted, not merely left stale. Worth the three extra words for anyone converting an existing cluster.

```suggestion
`ResourcesAwareManagement` indicates whether ECK could compute CPU, memory, and storage for all expected nodes so it can tell {{es}} the intended cluster topology. When the condition is `False`, ECK stops publishing that topology and removes any topology it published earlier. Without it, {{es}} may make less optimal topology-aware decisions. The condition is cluster-wide, so a single nodeSet using `emptyDir` sets it to `False` for the whole cluster. ECK still reconciles the `Elasticsearch` resource (StatefulSets, Pods, and CPU and memory settings) as usual.
```

pebrc (approved):
LGTM, my 🤖 has a few edit suggestions that overall seem sensible to me. See them as non-blocking.

barkbay:
@elastic/admin-docs, 👋 , approved by ECK team, enabling auto-merge on this one.

shainaraskas (approved):

PR #7891, opened 2026-08-10
docs: add _routing (custom routing) not supported in Serverless

pete-naylor:
## Summary

Adds a missing row to the Elasticsearch comparison table noting that `_routing` (custom routing) is not available in Serverless. Since Elastic fully manages sharding in Serverless projects, users cannot specify custom routing values at index or query time.

This gap applies to all Serverless project types (Elasticsearch, Observability, Security) but is documented once in the Elasticsearch table as `_routing` is a core Elasticsearch capability.

## Changes

- Added `_routing` (custom routing) row to the Elasticsearch feature comparison table, between Cross-cluster search and Data lifecycle management

## Notes for reviewer

Row follows existing table conventions — feature name linked to the `_routing` mapping field reference doc.


shainaraskas (commented):
hey Pete! I am thinking that this is too granular for this page. Ultimately, because there is no node/shard visibility for serverless, it follows that a bunch of stuff that targets those concepts doesn't work (e.g. shard allocation filtering, search `preference`, _shrink ...). I think that `_routing` is a given in that respect (where we need to expose things like ILM here because we need to direct users to an alternative).

pete-naylor:
Thanks @shainaraskas - I see your reasoning here. For more context, we have identified a bug that was allowing customers to use _routing only in the _bulk API. One customer repeatedly ran into operational problems and poor experience as a result. It seems that they will now move to ECH where they can use _routing because they feel there is a performance benefit for this. Please also see a related request to remove mention of _routing in various API entries for the Serverless API reference https://github.com/elastic/docs-content-internal/issues/1598

We want to be very upfront about this difference in the Serverless product offering, hence the mention here.

deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings.md
shainaraskas, line 45:
```suggestion
* High availability and disaster recovery strategies

Because Elastic fully manages sharding, several shard-related operations, including custom routing, are not supported. You can't set a `routing` value on document or search requests, and you can't create indexes that require routing with the [`_routing` field](elasticsearch://reference/elasticsearch/mapping-reference/mapping-routing-field.md). This applies to all {{serverless-short}} project types.
```

deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings.md
```diff
@@ -92,6 +92,7 @@ This table compares Elasticsearch capabilities between {{ech}} deployments and S
 | [**Bulk indexing**](/deploy-manage/production-guidance/optimize-performance/indexing-speed.md#_use_bulk_requests) |  ✅ | ✅ | The baseline write latency in {{serverless-short}} is 200ms [^1^](#footnote-1) |
 | [**Cross-cluster replication**](/deploy-manage/tools/cross-cluster-replication.md) | ✅ | **Planned** | Anticipated in a future release |
 | [**Cross-cluster search**](/explore-analyze/cross-cluster-search.md) | ✅ | **Tech preview** | As [cross-project search](/deploy-manage/cross-project-search-config.md) |
+| [**`_routing` (custom routing)**](elasticsearch://reference/elasticsearch/mapping-reference/mapping-routing-field.md) | ✅ | ❌ | Not available in Serverless |
```
shainaraskas:
```suggestion
| **Custom routing ([`_routing`](elasticsearch://reference/elasticsearch/mapping-reference/mapping-routing-field.md))** | ✅ | ❌ | Elastic fully manages sharding in {{serverless-short}}, so you can't set custom `routing` values or require routing on an index. |
```

shainaraskas (approved):
fair enough. let's be more specific, and include a callout where we talk about fully managed activities for completeness :) 

these edits assume that the underlying bug will be fixed shortly. if it will be a while, we might want to put a warning about what misuse can lead to

pete-naylor:
Thanks @shainaraskas - great suggestions.

deploy-manage/deploy/elastic-cloud/differences-from-other-elasticsearch-offerings.md
```diff
@@ -44,6 +44,8 @@ In Serverless, Elastic automatically manages:
 * Resource utilization and monitoring
 * High availability and disaster recovery strategies
 
+Because Elastic fully manages sharding, several shard-related operations, including custom routing, are not supported. You can't set a `routing` value on document or search requests, and you can't create indexes that require routing with the [`_routing` field](elasticsearch://reference/elasticsearch/mapping-reference/mapping-routing-field.md). This applies to all {{serverless-short}} project types.
```
shainaraskas:
```suggestion
Because Elastic fully manages sharding, several shard-related operations, including custom routing, are not supported. You can't set a `routing` value on document or search requests, and you can't create indices that require routing with the [`_routing` field](elasticsearch://reference/elasticsearch/mapping-reference/mapping-routing-field.md). This applies to all {{serverless-short}} project types.
```

shainaraskas (approved):
thanks, looks great

PR #7912, opened 2026-08-11
Clarify expectations around clock sync

### LINKED ISSUE #147398: Node monitor stats can remain stale after wall-clock rollback because refresh checks use System.currentTimeMillis()
### Elasticsearch Version
7.17.x (based on the affected code path in the 7.17-derived branch we investigated; please adjust if maintainers need the exact released tag)

### Installed Plugins
[]

### Java Version
Bundled JDK / affected by JVM wall-clock drift on the node

### OS Version
Linux

### Problem Description

We observed that `_nodes/<node>/stats/fs` can keep reporting stale filesystem usage after the node clock jumps forward and is later moved back to the correct time.

Expected behavior:
- Node monitor stats should continue refreshing according to `monitor.*.refresh_interval` even if the wall clock is corrected.

Actual behavior:
- If a refresh happens while the wall clock is ahead, and the clock is then rolled back, the refresh condition can remain false for a long time.
- As a result, `_nodes/stats` may return stale data until the wall clock catches up again or Elasticsearch is restarted.

In our case:
- `_nodes/<node>/stats/fs` reported about `73.4%` disk usage.
- The actual host filesystem usage was about `88%`.
- Downstream disk-watermark-based logic did not trigger as expected because Elasticsearch kept exposing stale fs stats.
- Restarting Elasticsearch immediately made the reported stats become correct again.

From code inspection, this looks like a wall-clock-based cache expiry problem rather than an `fs` math problem:

- `SingleObjectCache` uses `System.currentTimeMillis()` both to set `lastRefreshTimestamp` and to decide whether the cache needs refresh:
  - `server/src/main/java/org/elasticsearch/common/util/SingleObjectCache.java`
- `FsService` uses `SingleObjectCache` for filesystem stats:
  - `server/src/main/java/org/elasticsearch/monitor/fs/FsService.java`
- `OsService` and `ProcessService` also use `SingleObjectCache`:
  - `server/src/main/java/org/elasticsearch/monitor/os/OsService.java`
  - `server/src/main/java/org/elasticsearch/monitor/process/ProcessService.java`
- `JvmService` has the same wall-clock refresh pattern inline:
  - `server/src/main/java/org/elasticsearch/monitor/jvm/JvmService.java`
- `FsProbe` itself reads `FileStore` values directly on refresh, so the stale result seems to come from refresh suppression, not from wrong `free`/`available` calculation:
  - `server/src/main/java/org/elasticsearch/monitor/fs/FsProbe.java`

This means the impact likely is not limited to `fs` stats. `jvm`, `os`, and `process` stats may have the same stale-refresh problem after clock rollback.

### Steps to Reproduce

1. Start a single-node Elasticsearch instance with default monitor refresh intervals.
2. Call `GET /_nodes/stats/fs,jvm,os,process` at least once.
3. Move the host clock significantly forward (for example, `+1 day`).
4. Call `GET /_nodes/stats/fs,jvm,os,process` again so the monitor cache refreshes while the clock is in the future.
5. Move the host clock back to the correct current time.
6. Wait longer than `monitor.fs.refresh_interval` and change actual disk usage on the host.
7. Call `GET /_nodes/stats/fs,jvm,os,process` repeatedly.

Observed result:
- `fs/jvm/os/process` stats can remain stale for a long time.
- `fs` values do not reflect the current host filesystem state.
- Restarting Elasticsearch makes the stats immediately correct again.

Expected result:
- Refresh scheduling/expiry should use a monotonic time source for cache invalidation.
- Wall-clock corrections should not freeze monitor stats refresh.

### Logs (if relevant)

There are no obvious error logs.
The problem is silent and is visible by comparing:
- `_nodes/<node>/stats/fs`
- host-side `df -h`
- behavior before and after Elasticsearch restart

### Additional Notes

The field-level symptom can look like `free_in_bytes` / `available_in_bytes` or usage percentage being inconsistent with the actual filesystem state, but the deeper issue appears to be that node monitor stats are not refreshed after wall-clock rollback.

DaveCTurner:
Today we don't spell out anywhere the need for system clocks to be
reasonably well synchronized. This commit adds the missing guidance.

Closes elastic/elasticsearch#147398

### DIFF
--- a/deploy-manage/deploy/self-managed/system-config-clocks.md
+++ b/deploy-manage/deploy/self-managed/system-config-clocks.md
@@ -22,7 +22,7 @@ Keep the clocks on all nodes in the cluster synchronized with real time, and avo
 
 A properly-configured time synchronization service typically keeps clocks synchronized to within at most a few milliseconds. A few seconds of clock skew between nodes can cause some confusing effects, but {{es}} otherwise operates normally under these conditions. A skew or discontinuity of over ten seconds can cause unexpected or inefficient behavior.
 
-{{es}} does not rely on clock synchronization for any safety guarantees. For instance, if {{es}} acknowledges a write operation then this operation is guaranteed to be durable regardless of any clock skew or discontinuity.
+{{es}} does not rely on clock synchronization for any safety guarantees. For instance, if {{es}} acknowledges a write operation, then this operation is guaranteed to be durable regardless of any clock skew or discontinuity.
 
 ## Related pages
 
--- a/deploy-manage/deploy/self-managed/system-config-clocks.md
+++ b/deploy-manage/deploy/self-managed/system-config-clocks.md
@@ -18,11 +18,11 @@ products:
 * Timestamped log messages
 * Certain cache invalidation events
 
-Keep the clocks on all nodes in the cluster synchronized with real time, and avoid large discontinuities in these clocks. Run a time synchronization service such as [NTP](http://www.ntp.org/) or Chrony on every host.
+Keep the clocks on all nodes in the cluster synchronized with real time, and avoid discontinuities in these clocks in which the clock time jumps forwards or backwards by a large amount. Run a time synchronization service such as [NTP](http://www.ntp.org/) or Chrony on every host.
 
-A properly-configured time synchronization service will typically keep clocks synchronized to within at most a few milliseconds. A few seconds of clock skew between nodes can cause some confusing effects, but {{es}} will otherwise operate normally in such an environment. A skew or discontinuity of over ten seconds can cause unexpected or inefficient behavior.
+A properly-configured time synchronization service typically keeps clocks synchronized to within at most a few milliseconds. A few seconds of clock skew between nodes can cause some confusing effects, but {{es}} otherwise operates normally under these conditions. A skew or discontinuity of over ten seconds can cause unexpected or inefficient behavior.
 
-{{es}} does not rely on clock synchronization for any safety guarantees. For instance, if {{es}} acknowledges a write operation then this operation is guaranteed to be durable regardless of any clock skew between nodes.
+{{es}} does not rely on clock synchronization for any safety guarantees. For instance, if {{es}} acknowledges a write operation then this operation is guaranteed to be durable regardless of any clock skew or discontinuity.
 
 ## Related pages
 
--- a/deploy-manage/deploy/self-managed/system-config-clocks.md
+++ b/deploy-manage/deploy/self-managed/system-config-clocks.md
@@ -20,7 +20,7 @@ products:
 
 Keep the clocks on all nodes in the cluster synchronized with real time, and avoid large discontinuities in these clocks. Run a time synchronization service such as [NTP](http://www.ntp.org/) or Chrony on every host.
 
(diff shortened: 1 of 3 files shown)


deploy-manage/deploy/self-managed/system-config-clocks.md
```diff
@@ -0,0 +1,30 @@
...
+{{es}} relies on the system clock of each node for many time-dependent operations. This includes:
+
+* Timestamps recorded during ingest
+* Searches that use `now` (see [date math](elasticsearch://reference/elasticsearch/rest-apis/common-options.md#date-math))
+* Scheduled activities
+* Security features such as API key and token expiry, JWT and SAML time validation, and TLS certificate validity
+* Timestamped log messages
+* Certain cache invalidation events
+
+Keep the clocks on all nodes in the cluster synchronized with real time, and avoid large discontinuities in these clocks. Run a time synchronization service such as [NTP](http://www.ntp.org/) or Chrony on every host.
```
shainaraskas: what does discontinuities mean in this context? clocks not agreeing with each other? time jumping forward or backward? we should be specific because discontinuities is a little unclear (at least to my eyes)

deploy-manage/deploy/self-managed/system-config-clocks.md
```diff
@@ -0,0 +1,30 @@
...
+
+A properly-configured time synchronization service will typically keep clocks synchronized to within at most a few milliseconds. A few seconds of clock skew between nodes can cause some confusing effects, but {{es}} will otherwise operate normally in such an environment. A skew or discontinuity of over ten seconds can cause unexpected or inefficient behavior.
```
shainaraskas:
```suggestion
A properly-configured time synchronization service typically keeps clocks synchronized to within at most a few milliseconds. A few seconds of clock skew between nodes can cause some confusing effects, but {{es}} otherwise operates normally under these conditions. A skew or discontinuity of over ten seconds can cause unexpected or inefficient behavior.
```

deploy-manage/deploy/self-managed/system-config-clocks.md
```diff
@@ -0,0 +1,30 @@
...
+
+{{es}} does not rely on clock synchronization for any safety guarantees. For instance, if {{es}} acknowledges a write operation then this operation is guaranteed to be durable regardless of any clock skew between nodes.
```
shainaraskas:
```suggestion
{{es}} does not rely on clock synchronization for any safety guarantees. For instance, if {{es}} acknowledges a write operation, then this operation is guaranteed to be durable regardless of any clock skew between nodes.
```

shainaraskas (approved):
thanks, this looks great 🙏  tiny polishes for smooth reading and present tense

deploy-manage/deploy/self-managed/system-config-clocks.md
DaveCTurner: "Discontinuity" is a reasonably standard term of art here - added `in which the clock time jumps forwards or backwards by a large amount` to clarify.

deploy-manage/deploy/self-managed/system-config-clocks.md
DaveCTurner: ++ thanks

deploy-manage/deploy/self-managed/system-config-clocks.md
DaveCTurner: ++ thanks

PR #7916, opened 2026-08-11
Document the {{risonValue}} URL template variable

ch-bas:
Documents the new `{{risonValue}}` URL field-format template variable added in elastic/kibana#282587.

`{{risonValue}}` escapes a value for rison and then URL-encodes it, so it can be embedded in the rison state of a {{kib}} app URL (for example `query:'{{risonValue}}'`) where plain `{{value}}` would break the URL.

Added alongside the existing `{{rawValue}}` documentation in the URL field formatter sections.


explore-analyze/find-and-organize/data-views.md
```diff
@@ -545,6 +545,8 @@ When the formatter type is **Audio**, the `{{value}}` template string specifies
 
 To pass unescaped values directly to the URL, use the `{{rawValue}}` string.
 
+To insert a value into the rison-encoded state of a {{kib}} app URL, for example `query:'{{risonValue}}'`, use the `{{risonValue}}` string. It escapes the value for rison and then URL-encodes it, so characters such as `'` don't break the target URL.
```
florent-leborgne:
```suggestion
{applies_to}`stack: ga 9.6+` {applies_to}`serverless: ga` To insert a value into the [Rison](https://github.com/Nanonid/rison)-encoded state of a {{kib}} app URL, use the `{{risonValue}}` string in the URL template. Place it inside single quotes, for example `query:'{{risonValue}}'`. It escapes the value for Rison, then URL-encodes it, so that values containing `'` or `!` don't break the app state.
```

What do you think of this edit?
The applies_to are docs specific tags to indicate in which version of kibana the change landed/is landing.

If these changes are fine with you, we can apply that to the 3 occurrences.

Thanks for the contribution, this is much appreciated!

florent-leborgne (commented):
LGTM, thank you for the contribution. I'm not explicitly approving yet so it doesn't get merged and deployed ahead of the Kibana PR being deployed as well, which should be in a couple of days or early next week given it's just been merged. Thank you very much!

explore-analyze/find-and-organize/data-views.md
ch-bas: Looks great, thanks — applied your wording to all 3 occurrences in 1424875.

florent-leborgne:
Thank you again! I'll take care of merging this docs update when the Kibana change goes live

florent-leborgne (approved):

PR #7918, opened 2026-08-12
Document transport TLS certificate expiry behavior

ebarlas:
Manual transport TLS setup recommends using the same PKCS#12 file for both keystore and truststore without explaining that expired peer certificates may still be accepted. This gap confused users and was raised in elastic/security#1966.

Clarify that elastic-certificates.p12 acts as a cluster- membership key and note the expiry implications when the same file is used for keystore and truststore. Add a brief auto- configuration note in security-files.md covering transport.p12, enrollment tokens, and a cross-link to the manual setup guide.


deploy-manage/security/set-up-basic-security.md
```diff
@@ -123,6 +123,10 @@ Complete the following steps **for each node in your cluster**. To join the same
         xpack.security.transport.ssl.truststore.path: elastic-certificates.p12
         ```
 
+        ::::{note}
+        When the same PKCS#12 file is configured as both the keystore and truststore, peer node certificates in that file are trusted directly. In this configuration, {{es}} does not reject connections from nodes presenting an expired certificate from the same file, even when `verification_mode` is set to `certificate`. Internode communication remains encrypted, and the cluster can continue to operate normally, including restarts and adding new nodes that use the same file.
```
jfreden: >When the same PKCS#12 file is configured as both the keystore and truststore, peer node certificates in that file are trusted directly. 

There is an implicit fallback here where only configuring the keystore will result in the keystore turning into the truststore, so a user that tries to "fix" the issue by removing the truststore config would end up in the same situation. 

The note also reads as PKCS#12 specific, the same thing would happen with PEM certs `certificate_authorities: [node.crt, ca.crt]` (or any CA list containing a leaf that peers present).

 > even when `verification_mode` is set to `certificate`

This isn't really relevant, since the same thing would happen with any verification mode I think? 

Should this also mention that even a CA-only truststore doesn't check the CA's own expiry? I guess that's kind of implied.

deploy-manage/deploy/self-managed/_snippets/security-files.md
```diff
@@ -21,4 +21,6 @@ Use the following command to retrieve the password for `transport.p12`:
 
 ```sh
 bin/elasticsearch-keystore show xpack.security.transport.ssl.keystore.secure_password
-```
\ No newline at end of file
+```
+
+Security auto-configuration sets `certs/transport.p12` as both the transport keystore and truststore on each node. Auto-generated transport certificates have a long validity period and are unlikely to expire under normal operation. To add nodes, use an [enrollment token](/deploy-manage/maintenance/add-and-remove-elasticsearch-nodes.md#_enroll_nodes_in_an_existing_cluster_5) rather than copying `transport.p12` between nodes. For manual transport TLS setup and certificate expiry behavior when the same PKCS#12 file is shared across all nodes, refer to [Set up transport TLS](/deploy-manage/security/set-up-basic-security.md).
```
jfreden: Instead of "Long validity period" we can say 99 years and maybe mention that http certs have a 3 year validity period. WDYT?

jfreden (commented):
Thanks for adding this. I think just a docs change for closing the security issue is the right approach. I left a couple of suggestions and also have some high level ideas: 
- Would it make sense to add add some information to `auto-security-config` about what is actually written? Something like: 
> Autoconfig sets `transport.p12` as both keystore and truststore with verification_mode: certificate, the file functions as a cluster-membership key. Transport CA and node certs are valid for 99 years while the HTTP CA is 3 years and the HTTP certificate 2 years.
- Might be out of scope for this PR, but should `add-and-remove-elasticsearch-nodes.md` and `_snippets/enroll-nodes.md` be updated with some more information about  what happens actually happens to security material during enrollement?

deploy-manage/deploy/self-managed/_snippets/security-files.md
ebarlas: That's where I started as well, since those are helpful figures. But I decided against that since (1) it could invite scrutiny and (2) it could also get stale.

I'm open to either.

deploy-manage/security/set-up-basic-security.md
ebarlas: Good callouts. Updated in commit https://github.com/elastic/docs-content/pull/7918/changes/8ac384c0f61a129286c939455b48084eb8a58f3e

deploy-manage/deploy/self-managed/_snippets/security-files.md
jfreden: Yes, I was on the fence too when suggesting. Let's keep it as is.

jfreden (approved):
LGTM!

ebarlas:
@shainaraskas , would you review when you get a moment? Thanks!

ebarlas:
@elastic/admin-docs , would you review when you get a moment? Thanks!

deploy-manage/security/set-up-basic-security.md
```diff
@@ -123,6 +123,10 @@ Complete the following steps **for each node in your cluster**. To join the same
         xpack.security.transport.ssl.truststore.path: elastic-certificates.p12
         ```
 
+        ::::{note}
+        When the same file is configured as both the keystore and truststore, or when only `keystore.path` is set and {{es}} uses that file as the truststore, peer node certificates in the file are trusted directly. The same behavior applies if a peer's leaf certificate is listed in `xpack.security.transport.ssl.certificate_authorities`. In these configurations, {{es}} does not reject connections from nodes presenting an expired certificate that is directly trusted, regardless of the `verification_mode` setting. Internode communication remains encrypted, and the cluster can continue to operate normally, including restarts and adding new nodes that use the same certificate material.
+        ::::
+
```
shainaraskas, line 131: we shouldn't separate a code block and its annotation (the `1.` below)

```suggestion
```

deploy-manage/security/set-up-basic-security.md
```diff
@@ -123,6 +123,10 @@ Complete the following steps **for each node in your cluster**. To join the same
...
         1. If you want to use hostname verification, set the verification mode to `full`. You should generate a different certificate for each host that matches the DNS or IP address. See the `xpack.security.transport.ssl.verification_mode` parameter in [TLS settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#transport-tls-ssl-settings).
```
shainaraskas, line 126: because this is a procedure, we should keep it to the context of this config, and include these other details only where they're relevant 

```suggestion
        1. If you want to use hostname verification, set the verification mode to `full`. You should generate a different certificate for each host that matches the DNS or IP address. See the `xpack.security.transport.ssl.verification_mode` parameter in [TLS settings](elasticsearch://reference/elasticsearch/configuration-reference/security-settings.md#transport-tls-ssl-settings).
        
        ::::{note}
        When the same file is configured as both the keystore and truststore, peer node certificates in the file are trusted directly. {{es}} does not reject connections from nodes presenting an expired certificate that is directly trusted, regardless of the `verification_mode` setting. Internode communication remains encrypted, and the cluster can continue to operate normally, including restarts and adding new nodes that use the same certificate material.
        ::::
```


> or when only `keystore.path` is set and {{es}} uses that file as the truststore, peer node certificates in the file are trusted directly. The same behavior applies if a peer's leaf certificate is listed in `xpack.security.transport.ssl.certificate_authorities`.

this info likely needs to go on the relevant settings in https://www.elastic.co/docs/reference/elasticsearch/configuration-reference/security-settings#transport-tls-ssl-settings

... the core message of this note should probably also go in the reference on the relevant settings

deploy-manage/deploy/self-managed/_snippets/security-files.md
shainaraskas: I think we can move these details up here

```suggestion
:   Keystore that contains the key and certificate for the transport layer for all the nodes in your cluster. {{es}} also uses this file as the transport truststore. Auto-generated transport certificates have a long validity period and are unlikely to expire under normal operation.
```

deploy-manage/deploy/self-managed/_snippets/security-files.md
shainaraskas: this snippet is only used in the context of automatic security setup right now, so I don't think we need to reference the manual steps here

furthermore, this is auto-setup, not cluster maintenance, and we already advise on the correct procedure to enroll new nodes elsewhere (where we don't advise to copy files)

suggested moving the rest of the transport.p12 note up, see comment above


```suggestion
```

shainaraskas (commented):
thanks for this addition!

this PR injects a lot of info that is secondary to the topic at hand on these specific pages. I've suggested paring things down, and perhaps adding some more detail to the relevant settings in the setting reference as a backup. let me know what you think.

deploy-manage/security/set-up-basic-security.md
ebarlas: Are you referring to the trust store path setting `xpack.security.transport.ssl.truststore.path` in the "Transport TLS/SSL key and trusted certificate settings" section?

The setting descriptions are all quite short on the page, so emphasizing the impact of sharing keystore/truststore might be imbalanced.

shainaraskas: They are short, but if we have context that will help people use the settings better and there is no more ideal place, there is nothing wrong with adding more detail. some of our settings in the reference have a couple of paragraphs. I expect this will only end up adding maybe two sentences to each spot because we only need the details specific to that setting.

shainaraskas (approved):
looks great, thanks (and thanks for your patience!)

PR #7995, opened 2026-08-19
ECH upgrade: document workaround for out-of-order release upgrade block

kunisen:
## Summary

To address an internal ticket - https://github.com/elastic/support-tech-lead/issues/1919

The [out-of-order releases section](https://www.elastic.co/docs/deploy-manage/upgrade/deployment-or-cluster#out-of-order-releases) explains the release-date-based upgrade check but gives no guidance for customers already stuck on an out-of-order release. This PR adds an admonition documenting the workaround: create a new deployment at the target version and restore from a snapshot.

Confirmed via [sdh-elasticsearch#10199](https://github.com/elastic/sdh-elasticsearch/issues/10199) that the release date check does not apply to snapshot restore.

## Generative AI disclosure

- [x] Yes
- [ ] No

Claude (Anthropic) via Cowork

### DIFF
--- a/deploy-manage/upgrade/deployment-or-cluster.md
+++ b/deploy-manage/upgrade/deployment-or-cluster.md
@@ -31,7 +31,7 @@ Before proceeding with the upgrade, review the [Plan your upgrade](/deploy-manag
 
 Additionally, review the [Restrictions and known problems](../deploy/elastic-cloud/restrictions-known-problems.md) section to gain a comprehensive understanding of the known issues in {{ech}}. 
 
-If you’re still running {{stack}} version 7.17 or earlier, refer to the [Upgrade from 7.17 guide](/deploy-manage/upgrade/deployment-or-cluster/upgrade-717.md) for detailed guidance on planning and executing the upgrade to the latest {{version.stack}} release.
+If you're still running {{stack}} version 7.17 or earlier, refer to the [Upgrade from 7.17 guide](/deploy-manage/upgrade/deployment-or-cluster/upgrade-717.md) for detailed guidance on planning and executing the upgrade to the latest {{version.stack}} release.
 
 ## Out-of-order releases [out-of-order-releases]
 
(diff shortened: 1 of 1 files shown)


R0ky (approved):
LGTM

DaveCTurner (changes requested):
Hmm I don't think we should write this in the docs. I'm almost certain we don't test this path, and I'm pretty sure it isn't always going to work. The older version might not know how to handle data written by the newer version.

Instead, could we tighten up the wording around snapshot compatibility to spell out that it's not supported to restore a snapshot into a version which is either numerically older or chronologically older?

kunisen:
Thank you @DaveCTurner 
We got this wording per the context in https://github.com/elastic/sdh-elasticsearch/issues/10199, especially in https://github.com/elastic/sdh-elasticsearch/issues/10199#issuecomment-5338865843. 

If this is not true, then may I trouble you to suggest a safe wording from there please?

### Additional note

> Hmm I don't think we should write this in the docs. I'm almost certain we don't test this path, and I'm pretty sure it isn't always going to work. The older version might not know how to handle data written by the newer version.

Is there any blocker we document this fact? 
I understand we may not test this, but in the internal ticket (https://github.com/elastic/sdh-elasticsearch/issues/10199), then we will lose the ability to unblock the customer. 
=> The goal of documenting the unblock method is to let customer can self-service solve the issue. (If we disagree this, then customer / user will have no way to unblock logically. Sorry if I mis-understood your comment.)

---

> Instead, could we tighten up the wording around snapshot compatibility to spell out that it's not supported to restore a snapshot into a version which is either numerically older or chronologically older?

Yes, we could do it later. However, before we tighten up the wording, may I have your opinion about the above, so that we could unblock users who run into this situation please?

R0ky:
I'm looking at this: https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore#index-compatibility 

Unfortunately if it's true, that if we now need to look at release dates, we need to completely rework this document, as here it's quite clear that we only say lower version.

kunisen:
I will close this PR to prevent it from being merged since this path is not a tested and officially supported one.

PR #7998, opened 2026-08-19
[Security] Add missing 9.4.5 release notes fix

natasha-moore-elastic:
Adds the missing entity store extraction crash fix to the 9.4.5 Security release notes and marks the related known issue as resolved.

### DIFF
--- a/release-notes/elastic-security/known-issues.md
+++ b/release-notes/elastic-security/known-issues.md
@@ -31,7 +31,7 @@ On {{stack}} 9.4.4, {{elastic-sec}} deployments with a large number of indices c
 
 
 **Workaround**<br>
-To stabilize {{kib}} until we resolve the issue, we recommend temporarily disabling the Entity store feature:
+To stabilize {{kib}}, we recommend temporarily disabling the Entity store feature:
 
 1. Block the entity store tasks from running by applying this setting:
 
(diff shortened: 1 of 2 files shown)


chennn1990 (approved):

release-notes/elastic-security/known-issues.md
```diff
@@ -31,7 +31,7 @@ On {{stack}} 9.4.4, {{elastic-sec}} deployments with a large number of indices c
 
 
 **Workaround**<br>
-To stabilize {{kib}} until we resolve the issue, we recommend temporarily disabling the Entity store feature:
+To stabilize {{kib}}, we recommend temporarily disabling the Entity store feature:
```
bmorelli25, line 34: I think all of this can be deleted and just updated to:

```
**Workaround**<br>
Upgrade to {{stack}} 9.4.5.
```

That seems to be how we've done it in the past

bmorelli25 (approved):

release-notes/elastic-security/known-issues.md
natasha-moore-elastic: In my experience, we only do that when there's no real workaround on the affected version, so the only "workaround" is to upgrade to the fixed version. Otherwise, we keep the workaround for users who are still on the affected version, and just add when it was resolved.

PR #8034, opened 2026-08-20
ECH PrivateLink: warn that Cloud ID encodes non-VPCE URL

kunisen:
The Cloud ID always encodes the public (found.io-based) endpoint domain, not the PrivateLink-compatible elastic-cloud.com domain. Customers who derive their PrivateLink URL from the Cloud ID get a TLS cert mismatch.

- private-connectivity-aws.md: add note in Test the connection (ECH)
- find-cloud-id.md: add note warning PrivateLink users away from Cloud ID

Related: elastic/support-tech-lead#1921

## Generative AI disclosure

- [x] Yes  
- [ ] No  

Claude


deploy-manage/security/private-connectivity-aws.md
```diff
@@ -247,6 +247,10 @@ After you create your VPC endpoint and DNS entries, check that you are able to r
 ::::{applies-item} ess: ga
 :::{include} _snippets/private-url-struct.md
 :::
+
+:::{note}
+The Cloud ID for your deployment encodes the public (non-PrivateLink) endpoint domain. Do not use URLs derived from the Cloud ID for PrivateLink connections — they will not match the TLS certificate served at the PrivateLink endpoint and the connection will fail with a certificate hostname mismatch. Always construct your private URL manually as shown above. Review [our KB article](https://ela.st/avoid-using-cloudid-for-privatelink-deployment) for more guidance.
```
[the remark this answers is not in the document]
kunisen: no we may have to change the KB title later due to whatever reason so I want to keep things more generic and flexible.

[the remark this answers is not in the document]
kunisen: done in https://github.com/elastic/docs-content/pull/8034/changes/a23345a328bbbbd07fa4e2ee814c1345e3efaafb

deploy-manage/security/private-connectivity-aws.md
```diff
@@ -247,6 +247,10 @@ After you create your VPC endpoint and DNS entries, check that you are able to r
 ::::{applies-item} ess: ga
 :::{include} _snippets/private-url-struct.md
 :::
+
+:::{note}
+The Cloud ID for your deployment encodes the public (non-PrivateLink) endpoint domain. Do not use URLs derived from the Cloud ID for PrivateLink connections — they will not match the TLS certificate served at the PrivateLink endpoint and the connection will fail with a certificate hostname mismatch. Refer to the private URL structure in the preceding include. Review [our KB article](https://ela.st/avoid-using-cloudid-for-privatelink-deployment) for more guidance.
```
rahulranjan22: PrivateLink connections — they (We can remove dash)

rahulranjan22 (changes requested):
One minor comment regarding use of dash.

deploy-manage/security/private-connectivity-aws.md
kunisen:
```suggestion
The Cloud ID for your deployment encodes the public (non-PrivateLink) endpoint domain. Do not use URLs derived from the Cloud ID for PrivateLink connections, as they will not match the TLS certificate served at the PrivateLink endpoint and the connection will fail with a certificate hostname mismatch. Refer to the private URL structure in the preceding include. Review [our KB article](https://ela.st/avoid-using-cloudid-for-privatelink-deployment) for more guidance.
```

rahulranjan22 (approved):
Looking good.

deploy-manage/security/private-connectivity-aws.md
kunisen: Thanks @rahulranjan22 
I removed dash. 

Next time could we please let doc team review the doc related part? :) 
(The suggestion was made by AI and the doc AI in this PR didn't request me to change too, so IMHO this is not a wrong way of writing.)
My bad I should mention that I would love to have your review from tech perspective. 


Thanks anyways!

AlexP-Elastic (approved):
LGTM - thanks for making the change <3

(now we just have to remember to change it back when we finally do get round to fixing VPCE naming in cloud id :) )

cargious (approved):
LGTM, thank you for this improvement

deploy-manage/deploy/elastic-cloud/find-cloud-id.md
```diff
@@ -20,6 +20,10 @@ The Cloud ID works by assigning a unique ID to your hosted {{es}} cluster on {{e
 
 You include your Cloud ID along with your {{ecloud}} user credentials (defined in `cloud.auth`) when you run Beats or Logstash locally, and then let {{ecloud}} handle all of the remaining connection details to send the data to your hosted cluster on {{ecloud}} safely and securely.
 
+:::{note}
+The Cloud ID encodes the public (non-PrivateLink) endpoint domain for your deployment. If you are connecting over AWS PrivateLink, do not use the Cloud ID to derive your endpoint URL. Refer to [Private connectivity with AWS PrivateLink](/deploy-manage/security/private-connectivity-aws.md).
```
shainaraskas: If we want them to get more info from the AWS doc about this issue specifically, we should link them to the relevant section (see my other comment)

if it doesn't add a ton of value, we can also just link them directly to the kb from here so they don't have to hop around

deploy-manage/security/private-connectivity-aws.md
shainaraskas, line 79: here's where I think the limitation should move. we might want to put it in the snippet so it's visible on the main [limitations page](https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/8034/deploy-manage/deploy/elastic-cloud/restrictions-known-problems#ec-restrictions-network-security-kibana-sso) too. if we move it here we can link to the `#limitations` anchor from the cloud id page link

deploy-manage/security/private-connectivity-aws.md
```diff
@@ -247,6 +247,10 @@ After you create your VPC endpoint and DNS entries, check that you are able to r
 ::::{applies-item} ess: ga
 :::{include} _snippets/private-url-struct.md
 :::
+
+:::{note}
+The Cloud ID for your deployment encodes the public (non-PrivateLink) endpoint domain. Do not use URLs derived from the Cloud ID for PrivateLink connections, as they will not match the TLS certificate served at the PrivateLink endpoint and the connection will fail with a certificate hostname mismatch. Refer to the private URL structure in the preceding include. Review [our KB article](https://ela.st/avoid-using-cloudid-for-privatelink-deployment) for more guidance.
```
shainaraskas: an "include" is an internal docs concept we don't want to expose here

I think "urls derived from the cloud ID" is perhaps confusing ... unless you believe people are constructing urls with the cloud ID somehow, but we don't provide guidance on that in the docs - for example, [beats](https://www.elastic.co/docs/reference/beats/metricbeat/configure-cloud-id) just uses a flat cloud ID and afaik can't take a URL - what does the person do in that case? 

I think it would maybe make more sense to put this in the privatelink limitations unless there's a use case for constructing urls using cloud ID ... see other comment

shainaraskas (commented):
I've read through the linked issues and threads and am not finding anything that clarifies how people are actually trying to use cloud ID in this case, but given what we document about cloud ID I think this might need clarification (or I can just get clarification on the use case)

deploy-manage/deploy/elastic-cloud/find-cloud-id.md
kunisen: > how people are actually trying to use cloud ID in this case

The way is to "not use" cloud ID at all, but just use the deployment id (first part of cloud ID) and follow PrivateLink doc to make the correct URL for private connection. 
So in this Cloud ID page, in general we want to say "don't use it as is in PrivateLink use case".

KB is more for "error trace" where if user hit that, they can know how to handle. 
But here we probably are not stepping into the error yet, we want to raise awareness to user that they must not use Cloud ID. (and in PrivateLink doc we explain that more clearly)

---

However, if you think from a reader perspective, KB is easier to understand, then I am also fine with that opinion. It's more a "doc team" decision here :) Thanks!

deploy-manage/security/private-connectivity-aws.md
kunisen: Can we also add link to refer to the limitation page instead of simply removing it please?
It was there for years and if we simply remove that, and if user doesn't read limitation pages, they may run into issues.

As long as you agree with this point of view, I think we are fine with the update. 
May I trouble you to please make a suggestion about what in specific should I tweak?

Thank you!

deploy-manage/security/private-connectivity-aws.md
kunisen: `urls derived from the cloud ID` is related to `Cloud ID for your deployment encodes the public`, and we need to decode it with base64. 

So for example, in public doc - https://www.elastic.co/docs/reference/beats/metricbeat/configure-cloud-id
> `cloud.id: "staging:dXMtZWFzdC0xLmF3cy5mb3VuZC5pbyRjZWM2ZjI2MWE3NGJmMjRjZTMzYmI4ODExYjg0Mjk0ZiRjNmMyY2E2ZDA0MjI0OWFmMGNjN2Q3YTllOTYyNTc0Mw=="`

This could be decoded as per below:

---

image

---

`unless you believe people are constructing urls with the cloud ID somehow` in our case, it's indeed people are constructing URLs with the cloud ID somehow... (it was decoded and used as is, where it was decoded as `us-east-1.aws.found.io`, however, the real URL needs to be used in PrivateLink use case is different - `us-east-1.aws.elastic-cloud.com`

kunisen: > an "include" is an internal docs concept we don't want to expose here

Agreed. 

---

> what does the person do in that case?

So I'd think just saying "don't use cloud ID" would be the simplest way to go...
Or if you think we should add more details about how to decode, we can do that ... (out of this PR scope, so maybe a 2nd PR to follow that)

---

> I think it would maybe make more sense to put this in the privatelink limitations unless there's a use case for constructing urls using cloud ID ... see other comment

Thanks. As also commented in above, I think I am happy to follow doc team's suggestion for sure. 
But sorry that I am a bit unsure if I can get your point 100% clearly, so if possible, may I trouble @shainaraskas you to help me make the suggestion please?

Thank you again!

shainaraskas:
discussing enhancements with @kunisen in slack

shainaraskas:
@kunisen let me know what you think of this: https://github.com/elastic/docs-content/pull/8034/commits/b957a88558e80b4a2e0d998bdc998552151c05ff

companion PRs to beats and logstash: https://github.com/elastic/beats/pull/52845 / https://github.com/elastic/logstash/pull/19462

deploy-manage/security/private-connectivity-aws.md
shainaraskas: the limitation is now present on both the overall limitations page and the aws-specific page

kunisen:
@shainaraskas thank you it looks great!
The only thing I want to double check is I can see the KB was removed from public mentioning. 

Is that intentional and because of we want to add more details in public doc, and KB is no longer really helpful (because it only contains error trace and other than that, nothing beyond public docs, as well as it didn't mention beats and logstash, etc)?

If so, then I am good with we go with your suggestions.
Thank you!!

shainaraskas:
@kunisen I think that given what we have in the docs, the link will not be adding a lot more value to users - it's much more valuable in reverse (kb link to docs workaround).

shainaraskas (approved):

shainaraskas:
oops one fix coming in - forgot you can't just copy the private link from the admin console for ech yet

kunisen:
> @kunisen I think that given what we have in the docs, the link will not be adding a lot more value to users - it's much more valuable in reverse (kb link to docs workaround).

Noted thank you!

---

There's one remaining thing that Logstash page didn't get updated yet despite of the PR merge. I asked here - https://github.com/elastic/logstash/pull/19462#issuecomment-5433252303. But maybe it's just a matter of time, and thank you again @shainaraskas for all the great help! 🙏

PR #8037, opened 2026-08-20
Adapt docs-triage to the codex engine behind the OpenRouter proxy

theletterf:
docs-actions#291 switched the shared triage workflows from the copilot engine to codex, routed through the Elastic Docs OpenRouter proxy:

- A new `exchange` job trades the workflow's GitHub OIDC token for a short-lived proxy token (this repo is already on the proxy's allow-list). That's why `id-token: write` appears in permissions.
- The token is passed explicitly as `CODEX_API_KEY`. Explicit `secrets:` replaces `secrets: inherit` because job outputs can't ride `inherit` — the three optional gh-aw secrets are passed through by name so nothing that worked before stops working.

Validated end to end in docs-actions-sandbox: real `issue_comment`/`issues` triggers, green runs, labels and comments applied (elastic/docs-actions-sandbox#57–60).

### DIFF
--- a/.github/workflows/docs-triage.yml
+++ b/.github/workflows/docs-triage.yml
@@ -8,16 +8,37 @@ permissions:
   actions: write
   contents: read
   discussions: write
+  id-token: write
   issues: write
   pull-requests: write
   copilot-requests: write
 
 jobs:
-  run:
+  exchange:
     if: >-
       github.event_name == 'workflow_dispatch' ||
       startsWith(github.event.comment.body, '/triage')
-    uses: elastic/docs-actions/.github/workflows/gh-aw-issue-triage.lock.yml@v1
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+    outputs:
+      token: ${{ steps.get.outputs.token }}
+    steps:
+      - id: get
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-triage.yml
```diff
@@ -8,16 +8,37 @@ permissions:
...
+            "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=openrouter-proxy" | jq -r .value)
+          token=$(curl -sSf -X POST -H "Authorization: Bearer $oidc" "$PROXY_URL/token" | jq -r .token)
+          echo "token=$token" >> "$GITHUB_OUTPUT"
   run:
+    needs: exchange
     if: >-
       github.event_name == 'workflow_dispatch' ||
       startsWith(github.event.comment.body, '/triage')
-    uses: elastic/docs-actions/.github/workflows/gh-aw-issue-triage.lock.yml@v1
+    uses: elastic/docs-actions/.github/workflows/gh-aw-issue-triage.lock.yml@main
```
reakaleek: why `@main` ?

theletterf: Ah, silly Claude. Let me fix.

reakaleek (approved):

PR #8045, opened 2026-08-21
ci: validate changelog only when changelog files change

theletterf:
## Summary

- Removes `edited`, `labeled`, and `unlabeled` from the `pull_request` trigger types — these re-ran validation when nothing in the changelog changed.
- Adds a `paths` filter scoped to `docs/changelog/**` so the workflow only fires when a changelog file is actually added or modified.
- Keeps `opened`, `synchronize`, and `reopened` as the only trigger types.

## Test plan

- [ ] Open a PR that touches `docs/changelog/` — workflow should trigger.
- [ ] Open a PR with no changelog files — workflow should not appear in checks.
- [ ] Add a label to a PR with a changelog file — workflow should not re-trigger.

### DIFF
--- a/.github/workflows/changelog-validate.yml
+++ b/.github/workflows/changelog-validate.yml
@@ -2,7 +2,9 @@ name: changelog-validate
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, edited, labeled, unlabeled]
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'docs/changelog/**'
 
 permissions: {}
 


.github/workflows/changelog-validate.yml
```diff
@@ -2,7 +2,9 @@ name: changelog-validate
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, edited, labeled, unlabeled]
+    types: [opened, synchronize, reopened]
```
reakaleek, line 5: I'm not entirely sure how this works, but there is a connection between the changelog and labels. Hence, I think you always need to validate changelogs together with labels.

reakaleek: I think that was the reason for also validating on `labeled` and `unlabeled`

theletterf:
Closing — this change is incorrect. The `labeled`, `unlabeled`, and `edited` triggers are intentional: labels determine the changelog type and title changes affect the generated content. The `paths` filter would prevent validate from firing when no changelog file exists yet, which is exactly when it needs to run to kick off auto-generation via the downstream `changelog-submit` workflow_run trigger.

PR #8053, opened 2026-08-24
ECE users-permissions: clarify that ECE service containers run as root (UID 0) by design

kunisen:
## Summary

The page stated "You must not install ECE as the root user" without clarifying that the ECE service containers themselves run as root (UID 0) internally — leading customers to assume ECE is root-averse end-to-end. This PR adds a note drawing that distinction clearly.

Related: elastic/support-tech-lead#1924

## Generative AI disclosure

- [x] Yes
- [ ] No

Tool(s) and model(s) used: Claude (Anthropic) via Cowork


deploy-manage/deploy/cloud-enterprise/ece-users-permissions.md
```diff
@@ -30,4 +30,6 @@ In this example, the user `elastic` with a UID and GID of 1000 belongs to both t
 For ECE installation with Podman, the user does not need to be added to the `docker` group. Instead, the user must be added to the `podman` group.
 ::::
 
-
+::::{note}
+While the ECE installer user must be non-root (UID ≥ 1000), the [ECE service containers](ece-architecture.md#ece-service-containers) run as `root` (UID 0) internally. This is an architectural requirement. Elastic Stack containers run as UID 1000 (non-root) since {{es}} 8.0.
```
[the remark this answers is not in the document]
kunisen: I used admonition instead since it's not logically reasonable to put in plaintext, but it's indeed a note for heads-up purposes.

deploy-manage/deploy/cloud-enterprise/ece-users-permissions.md
```diff
@@ -26,8 +26,10 @@ uid=1000(elastic) gid=1000(elastic) groups=1000(elastic),
 ```
 In this example, the user `elastic` with a UID and GID of 1000 belongs to both the `sudo` and the `docker` groups.
 
+:::{admonition} ECE service containers run as root
+While the ECE installer user must be non-root (UID ≥ 1000), the [ECE service containers](ece-architecture.md#ece-service-containers) run as `root` (UID 0) internally. This is an architectural requirement. Elastic Stack containers run as UID 1000 (non-root) since {{es}} 8.0.
```
shainaraskas: we only support 8.0+ so we don't need to say "since 8.0"

```suggestion
While the ECE installer user must be non-root (UID ≥ 1000), the [ECE service containers](ece-architecture.md#ece-service-containers) run as `root` (UID 0) internally. This is an architectural requirement.
```

shainaraskas (commented):
I don't think this is the (only) right place for this info, because this fact is not strictly relevant to the user you need to set up. Can we possibly add it here as well? https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-architecture#ece-service-containers

What is the implication to the user of this information? What do they need to do or avoid when they set up their environment to allow these containers to run as root? e.g. avoid registry policies that prohibit this action. maybe that goes here, with private registries? https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-install-offline-with-registry

> If your registry or image scanner rejects images that run as the root user, add an exception for ECE platform images, including elastic-cloud-enterprise. Those images run as UID 0. ECE cannot run if the registry blocks them.

Might also be nice to explain the why: "because the allocator/runner containers interact directly with the Docker/Podman socket and perform privileged operations (cgroup management, resource allocation)"

deploy-manage/deploy/cloud-enterprise/ece-users-permissions.md
kunisen: @shainaraskas ++ on removing 8.0, but can we keep this sentence please? 
This is to make it clear that service containers running as root, but non service containers - stack containers run as non-root.

kunisen:
```suggestion
While the ECE installer user must be non-root (UID ≥ 1000), the [ECE service containers](ece-architecture.md#ece-service-containers) run as `root` (UID 0) internally. This is an architectural requirement. Elastic Stack containers run as UID 1000 (non-root)
```

kunisen: wdyt this? @shainaraskas

kunisen:
@shainaraskas thanks for the comments!

### [1]

> I don't think this is the (only) right place for this info, because this fact is not strictly relevant to the user you need to set up. Can we possibly add it here as well? https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-architecture#ece-service-containers

Yes, we can.
But that service container page only talks about functionalities and doesn't talk about permissions. 
My personal feeling is adding there might looks a bit unrelated.

However, if you feel from reader perspective, adding the piece there would clarify things better, I am ok with adding that.

---

### [2]

> What is the implication to the user of this information? What do they need to do or avoid when they set up their environment to allow these containers to run as root? e.g. avoid registry policies that prohibit this action. maybe that goes here, with private registries? https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-install-offline-with-registry

This is to reduce the confusion and tell user that, they need to understand that ECE needs to run those service containers as root.

So originally we say: `You must not install ECE as the root user.` [here](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-users-permissions#:~:text=You%20must%20not%20install%20ECE%20as%20the%20root%20user.) in public doc.
It can make wrong implication that ECE runs as non-root, however, in reality, it's not. 

Adding the piece is to make user able to get that implication correctly.
(Because sometimes user wants to get rid of installing services as root)

---

### [3]

> > If your registry or image scanner rejects images that run as the root user, add an exception for ECE platform images, including elastic-cloud-enterprise. Those images run as UID 0. ECE cannot run if the registry blocks them.
> 
> Might also be nice to explain the why: "because the allocator/runner containers interact directly with the Docker/Podman socket and perform privileged operations (cgroup management, resource allocation)"

I understand your suggestion but that's more technical ones, and I don't really know if that's the only reason of ECE needs service containers to be running as root.
If @shainaraskas you think we must add this, then I'd defer to @matt-elastic for this content approval.

However, I am a bit confused why we need this.
`add an exception` to allow ECE to run as root, is seemingly reasonable, however, if we add that, maybe customers will also chase for how to do it, etc, and then we also need to explain it's out of Elastic Support scope.
Then I feel it might be a bit too much for the scope of this doc PR.

@shainaraskas my personal vote is we can have a ticket 
* discussing the user permissions in details, including container side topic (you shared above) and scope (we only provide direction but we do not provide detailed steps)
* and we do the description in [user permission page](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-users-permissions) first
* and then we link this page in different relevant pages, such as the [service container page](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-architecture#ece-service-containers) and [air gap page](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-install-offline-with-registry).

I am a bit hesitant we discuss and address all these in this doc PR... 😅 

---

Hope that makes sense and thank you!

deploy-manage/deploy/cloud-enterprise/ece-users-permissions.md
shainaraskas: needs a period. otherwise lgtm 

```suggestion
While the ECE installer user must be non-root (UID ≥ 1000), the [ECE service containers](ece-architecture.md#ece-service-containers) run as `root` (UID 0) internally. This is an architectural requirement. Elastic Stack containers run as UID 1000 (non-root).
```

shainaraskas:
> ### [1]
> > ... Can we possibly add it here as well? [elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-architecture#ece-service-containers](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-architecture#ece-service-containers)
> 
> Yes, we can. But that service container page only talks about functionalities and doesn't talk about permissions. My personal feeling is adding there might looks a bit unrelated.

This is ultimately a function of the architecture, not permissions, so I think that this is the more accurate spot.

> 
> ### [2]
> > 
> 
> This is to reduce the confusion and tell user that, they need to understand that ECE needs to run those service containers as root.
> 

My question here is not about the implication of the PR, but instead: what does the user need to do based on this information? What action do they need to perform or avoid? that action seems to be "don't block this when you use a private container registry". I think that without the guidance, adding this information is not as helpful.

> ### [3]
> > > If your registry or image scanner rejects images that run as the root user, add an exception for ECE platform images, including elastic-cloud-enterprise. Those images run as UID 0. ECE cannot run if the registry blocks them.
> > 
> > 
> > Might also be nice to explain the why: "because the allocator/runner containers interact directly with the Docker/Podman socket and perform privileged operations (cgroup management, resource allocation)"
> 
> I understand your suggestion but that's more technical ones, and I don't really know if that's the only reason of ECE needs service containers to be running as root. If @shainaraskas you think we must add this, then I'd defer to @matt-elastic for this content approval.
> 
> However, I am a bit confused why we need this. `add an exception` to allow ECE to run as root, is seemingly reasonable, however, if we add that, maybe customers will also chase for how to do it, etc, and then we also need to explain it's out of Elastic Support scope. Then I feel it might be a bit too much for the scope of this doc PR.

I pulled this from the dev's reply on the SDH so it should already be accurate. I think it's nice when we say we must have elevated permissions that we say why we require those permissions. no issue with getting another approval.

For customers that have this limitation, it's their private registry causing the issue, so I think leaving the workaround to them is ok. We can cut it down to just connect this with the idea that a private registry block will cause issues and they'll need to not block these images somehow. I think this is likely an edge case too so we don't need to worry about elevated support volume (because this hasn't come up frequently in the past)
 
> @shainaraskas my personal vote is we can have a ticket
> 
> * discussing the user permissions in details, including container side topic (you shared above) and scope (we only provide direction but we do not provide detailed steps)
> * and we do the description in [user permission page](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-users-permissions) first
> * and then we link this page in different relevant pages, such as the [service container page](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-architecture#ece-service-containers) and [air gap page](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-install-offline-with-registry).
> 
> I am a bit hesitant we discuss and address all these in this doc PR... 😅
> 
> Hope that makes sense and thank you!

I don't agree that we can only ship the change to the user page. the minimum change, in my view, is to the service container and air gapped pages, and not as a link. the service container page is the canonical home, because these are not exactly user actors. The users + permissions page is only nice to have, in my view.

I think we can do a little less in this PR, e.g. leave out the why (which is only nice to have), but I don't think we can leave out the workaround (at least clearly stating the conflict lies at the private container registry level). I don't think this needs or should have another ticket. This PR does not do enough to address the issue unless this scope is addressed.

deploy-manage/deploy/cloud-enterprise/ece-users-permissions.md
kunisen: Thank you!

kunisen:
@shainaraskas thanks for the detailed explanation.
I now fully get your point and I fully agree with all the above pointers you mentioned. 

> My question here is not about the implication of the PR, but instead: what does the user need to do based on this information? What action do they need to perform or avoid? that action seems to be "don't block this when you use a private container registry". I think that without the guidance, adding this information is not as helpful.

Yes, you are right, not only give opinions, but also give call to actions. 

---

> I pulled this from the dev's reply on the SDH so it should already be accurate. I think it's nice when we say we must have elevated permissions that we say why we require those permissions. no issue with getting another approval.

Yes, you are right, we should do this to solve the real problems and make call to actions.

---

> I don't agree that we can only ship the change to the user page. the minimum change, in my view, is to the service container and air gapped pages, and not as a link. the service container page is the canonical home, because these are not exactly user actors. The users + permissions page is only nice to have, in my view.

Yes, you are right, now I understand that we should also make more clear statements in air-gapped page and service container pages to explain the importance of running as root and make heads-up for better risk prevention.

---

I will do updates based on your suggestion accordingly.

kunisen:
@shainaraskas thanks for being patient. I updated the PR to include all your suggestions. 

One more small thing, about Elastic Stack containers running as non-root user, I had a test that they may not be always running at UID 1000, so I used UID >= 1000 to loose the statement. We can also remove the UID mentioning at all if you think it may cause more confusion than it clarifies.

Below is the test log and they show 1024 as UID, for ES & Kibana containers.

```bash
# test log
#----------------------------
kuniyasu_sen@sl-kuniyasusen-7dcb6fd4-host1:~$ sudo su - 

root@sl-kuniyasusen-7dcb6fd4-host1:~# docker ps | grep elasticsearch
13bfdfc1afdc   docker.elastic.co/cloud-release/elasticsearch-cloud-ess:8.19.20     "/sbin/tini -- /app/…"   21 hours ago   Up 21 hours             9200/tcp, 0.0.0.0:18488->18488/tcp, 0.0.0.0:19495->19495/tcp, 0.0.0.0:20345->20345/tcp, 9300/tcp, 0.0.0.0:23530->23530/tcp                                                                                                 fac-29b16c6d37be4ff9a45c05498e4fc48d-instance-0000000000
11dd08ec67c4   docker.elastic.co/cloud-release/elasticsearch-cloud-ess:8.19.20     "/sbin/tini -- /app/…"   21 hours ago   Up 21 hours             9200/tcp, 0.0.0.0:18617->18617/tcp, 0.0.0.0:19398->19398/tcp, 0.0.0.0:20720->20720/tcp, 9300/tcp, 0.0.0.0:23546->23546/tcp                                                                                                 fac-29b16c6d37be4ff9a45c05498e4fc48d-instance-0000000001
992247263c7d   docker.elastic.co/cloud-release/elasticsearch-cloud-ess:8.19.20     "/sbin/tini -- /app/…"   21 hours ago   Up 21 hours             9200/tcp, 0.0.0.0:18056->18056/tcp, 0.0.0.0:19015->19015/tcp, 0.0.0.0:20609->20609/tcp, 9300/tcp, 0.0.0.0:23263->23263/tcp                                                                                                 fac-54f3af882c934fb4acb7fa277343cbee-instance-0000000000
6d78fff9d413   docker.elastic.co/cloud-release/elasticsearch-cloud-ess:8.19.20     "/sbin/tini -- /app/…"   21 hours ago   Up 21 hours             9200/tcp, 0.0.0.0:18636->18636/tcp, 0.0.0.0:19473->19473/tcp, 0.0.0.0:20536->20536/tcp, 9300/tcp, 0.0.0.0:23956->23956/tcp                                                                                                 fac-d318f7afe81e476e9d68a1872bf06758-instance-0000000000
6979bf81f62d   docker.elastic.co/cloud-release/elasticsearch-cloud-ess:8.19.20     "/sbin/tini -- /app/…"   21 hours ago   Up 21 hours             9200/tcp, 0.0.0.0:18032->18032/tcp, 0.0.0.0:19817->19817/tcp, 0.0.0.0:20305->20305/tcp, 9300/tcp, 0.0.0.0:23236->23236/tcp                                                                                                 fac-d318f7afe81e476e9d68a1872bf06758-instance-0000000001

root@sl-kuniyasusen-7dcb6fd4-host1:~# docker ps | grep kibana
905f682de799   docker.elastic.co/cloud-release/kibana-cloud:8.19.20                "/bin/tini -- /app/k…"   21 hours ago   Up 21 hours             5601/tcp, 0.0.0.0:18365->18365/tcp                                                                                                                                                                                         fac-c3ddcb1a063d48db913e2194a1104524-instance-0000000000

root@sl-kuniyasusen-7dcb6fd4-host1:~# docker exec 13bfdfc1afdc id
uid=1024 gid=1024 groups=0(root),1024

root@sl-kuniyasusen-7dcb6fd4-host1:~# docker exec 905f682de799 id
uid=1024 gid=1024 groups=0(root),1024


# additional information
# - for service containers, they run as root - UID 0 
# -------------------------------------
root@sl-kuniyasusen-7dcb6fd4-host1:~# docker ps | grep zook
6dc692eab55e   docker.elastic.co/cloud-enterprise/elastic-cloud-enterprise:4.2.0   "/usr/sbin/my_init"      21 hours ago   Up 21 hours (healthy)   0.0.0.0:2191->2191/tcp, 0.0.0.0:12191->12191/tcp, 0.0.0.0:12898->12898/tcp, 0.0.0.0:13898->13898/tcp                                                                                                                       frc-zookeeper-servers-zookeeper

root@sl-kuniyasusen-7dcb6fd4-host1:~# docker exec frc-zookeeper-servers-zookeeper id
uid=0(root) gid=0(root) groups=0(root)

root@sl-kuniyasusen-7dcb6fd4-host1:~# docker ps | grep forwarder
7176281a9a83   docker.elastic.co/cloud-enterprise/elastic-cloud-enterprise:4.2.0   "/usr/sbin/my_init"      21 hours ago   Up 21 hours (healthy)   0.0.0.0:9244->9244/tcp                                                                                                                                                                                                     frc-services-forwarders-services-forwarder
7ba60e0c7c36   docker.elastic.co/cloud-enterprise/elastic-cloud-enterprise:4.2.0   "/usr/sbin/my_init"      21 hours ago   Up 21 hours (healthy)                                                                                                                                                                                                                              frc-client-forwarders-client-forwarder

root@sl-kuniyasusen-7dcb6fd4-host1:~# docker exec frc-services-forwarders-services-forwarder id
uid=0(root) gid=0(root) groups=0(root)

root@sl-kuniyasusen-7dcb6fd4-host1:~# docker exec frc-client-forwarders-client-forwarder id
uid=0(root) gid=0(root) groups=0(root)
```

kunisen:
(hit an error, do we know how to fix please? https://github.com/elastic/docs-content/actions/runs/33136715201/job/98738239958/#step:7:1)

deploy-manage/deploy/cloud-enterprise/ece-architecture.md
```diff
@@ -107,3 +107,5 @@ Each {{ece}} service runs as a dedicated container. These containers are automat
 | `frc-zookeeper-servers-zookeeper`                 | Director      | Consistent distributed data store used to track ECE state and coordinate communication between services. |
 | `frc-proxies-proxyv2`                             | Proxy         | Routes user traffic to the {{stack}} deployments. |
 | `frc-proxies-route-server`                        | Proxy         | Manages the routing tables used by the proxy service. |
+
+ECE service containers run as `root` (UID 0) internally. This is an architectural requirement: the containers interact directly with the Docker or Podman socket and perform privileged operations, including cgroup management and resource allocation. {{stack}} containers, such as {{es}} and {{kib}}, run as a non-root user (UID ≥ 1000).
```
shainaraskas: we should move this above the big table
```suggestion
```

deploy-manage/deploy/cloud-enterprise/ece-architecture.md
shainaraskas: we can move this up here, and make it a note so it acts as supplementary 

we can drop "internally" in the note because it's in context

```suggestion
Each {{ece}} service runs as a dedicated container. These containers are automatically deployed based on the roles assigned to each ECE host. The following table lists the containers on ECE hosts, along with the host roles that include each container.

:::{note}
ECE service containers run as `root` (UID 0). This is an architectural requirement and is not configurable: the containers interact directly with the Docker or Podman socket and perform privileged operations, including cgroup management and resource allocation. {{stack}} containers, such as {{es}} and {{kib}}, run as a non-root user (UID ≥ 1000).
:::
```

deploy-manage/deploy/cloud-enterprise/ece-install-offline-with-registry.md
```diff
@@ -13,6 +13,10 @@ products:
 
 Installing ECE on multiple hosts with your own registry server is simpler, because you do not have to load the Docker images on each host.
 
+::::{admonition} ECE service containers run as root
+[ECE service containers](ece-architecture.md#ece-service-containers), run as `root` (UID 0) because they interact directly with the Docker or Podman socket and perform privileged operations such as cgroup management and resource allocation. If your private registry or image scanner rejects images that run as the root user, add an exception for ECE service images. ECE cannot start if the registry blocks them.
```
shainaraskas: given my understanding of the error, this would not likely result in ECE being unable to start, but instead the registry rejecting the image (so you can't `docker push`, and can never host the image [and have nothing to "start"]).

```suggestion
[ECE service containers](ece-architecture.md#ece-service-containers) run as `root` (UID 0) because they interact directly with the Docker or Podman socket and perform privileged operations such as cgroup management and resource allocation. If your private registry or image scanner rejects images that run as the root user, add an exception for `elastic-cloud-enterprise`.
```

shainaraskas (approved):
looks really good. just some polish up for visibility and a little accuracy suggestion based on my current understanding.

deploy-manage/deploy/cloud-enterprise/ece-install-offline-with-registry.md
kunisen: Thanks yes you are right, the containers won't be created at all, so the "not start" is not correct enough.

kunisen:
Thank you so much @shainaraskas for your detailed check and suggestion as always! 🙇

PR #8084, opened 2026-08-25
[Security] Entity store opt-in in all Kibana spaces

### LINKED ISSUE #8073: [Entity Analytics] Default Opt in Entity Store Setting
In the Entity Analytics documentation for Risk scoring, it states that in a default kibana space, risk scoring and entity store are auto enabled. This has changes to opt-in option in all spaces. 

**PR:**  https://github.com/elastic/kibana/pull/281432

**Documentation:** https://www.elastic.co/docs/solutions/security/advanced-entity-analytics/turn-on-risk-scoring-engine 

**Remove line:** "In the default Kibana space, both risk scoring and entity store are enabled automatically." 

**Change line:** "For non-default spaces, if you're enabling risk scoring for the first time:" 

**to** "For default and non-default spaces, if you're enabling risk scoring for the first time"

natasha-moore-elastic:
Documents that from 9.4.6 and 9.5.2, risk scoring and the entity store require opt-in in every Kibana space, including the default space. The previous default space auto-enable behavior is preserved for earlier 9.4 and 9.5 patches.

Minor content deduplication: Updates the Enable entity store section to link to Turn on risk scoring instead of repeating the enablement steps.

Resolves #8073


CAWilson94 (approved):

solutions/security/advanced-entity-analytics/entity-store.md
```diff
@@ -68,12 +68,7 @@ For each entity type (hosts, users, and services):
 ::::{applies-switch}
 
 :::{applies-item} { stack: ga 9.4+, serverless: ga }
-The entity store is automatically enabled when you turn on risk scoring. In the default {{kib}} space, both are enabled automatically. In non-default spaces, you must enable them manually:
-
-1. Go to the **Entity Analytics** management page. Accessing this page differs based on the [solution view](/deploy-manage/manage-spaces.md#spaces-managing) that you're using:
-    * **Security solution view**: Find **{{stack-manage-app}} → Entity Analytics** in the navigation menu.
-    * **Classic view**: Find **Manage → Entity Analytics** in the navigation menu.
-2. Turn the toggle on.
+The entity store is automatically enabled when you [turn on risk scoring](/solutions/security/advanced-entity-analytics/turn-on-risk-scoring-engine.md#_turn_on_the_latest_risk_engine).
```
vishaangelova: According to the changes in [turn-on-risk-scoring-engine.md‎](https://github.com/elastic/docs-content/pull/8084/changes#diff-7c090f5c4ad7831ec38a975b7526166d3ba1ea044f3145041c96d0be79212cb4) in this PR:

```
In {{stack}} 9.4.0–9.4.5 and 9.5.0–9.5.1, risk scoring and the entity store are enabled automatically in the default {{kib}} space. Starting in 9.4.6 and 9.5.2, you must enable them in every space, including the default space.
```

So I think we need to update the `applies-item` here as well, or better yet, keep the existing applies-item tab (revert the deletion), and add another one for the fact that users need to enable the entity store from 9.4.6 and 9.5.2 onwards?

I'm wondering if this use case is one of those cases where we need to explicitly state the patch in the tag?

natasha-moore-elastic: Hmm I think using the word "automatically" here may have caused confusion. These are two separate behaviors:

* From 9.4, the entity store is enabled "automatically" when users turn on risk scoring, meaning it's no longer enabled separately to risk scoring, which was the case before 9.4. So from 9.4, users just need to turn on risk scoring to get the entity store enabled. Currently, we're repeating those enablement steps (Go to the **Entity Analytics** management page., etc) on both pages. That’s why I deleted the duplicated enablement steps here and linked to Turn on risk scoring instead.

* From 9.4.6 and 9.5.2, users must enable risk scoring (and entity store gets enabled with it) manually everywhere, including the default space. They're no longer auto-on in the default space.


Would removing "automatically" from the sentence on line 71 make this clearer?

natasha-moore-elastic: > I'm wondering if this use case is one of those cases where we need to explicitly state the patch in the tag?

I was thinking the same, but in the end I decided to follow what's in the [official guidance](https://www.elastic.co/docs/contribute-docs/how-to/cumulative-docs/reference#version):

> When critical patch-level differences exist, use plain text descriptions alongside the badge rather than specifying patch versions.

I'm open to using the patch tag though, if you think that would be clearer.

vishaangelova: Ah got it—yes, I think if we removed "automatically", this would work.

And let's follow the official guidance about the `applies_to` :)

vishaangelova (approved):
LGTM (but could we remove the "automatically"?:)

solutions/security/advanced-entity-analytics/entity-store.md
natasha-moore-elastic:
```suggestion
The entity store is enabled when you [turn on risk scoring](/solutions/security/advanced-entity-analytics/turn-on-risk-scoring-engine.md#_turn_on_the_latest_risk_engine).
```

PR #8120, opened 2026-08-31
[ECE] Document component template method to customize logging-and-metrics retention

eedugon:
## Summary

- Documents the new component template method to customize data retention in the ECE `logging-and-metrics` cluster, available from ECE 4.1.1 (applies_to set to 4.2 to avoid confusion in 4.1.0).
- Adds a table of reserved `@custom` component template names with their scope (index-specific, all logs, all metrics, all data streams).
- Renames the existing cloning method section and adds a note directing ECE 4.1.1+ users to the new approach.
- Updates the page intro to describe the overall process and differentiate by version.

## Related

Closes https://github.com/elastic/docs-content-internal/issues/840

## Test plan

- [ ] Preview builds correctly and the new section renders with the `ga 4.2+` badge.
- [ ] Links to ILM policy, component templates, and data stream rollover docs resolve correctly.
- [ ] `applies_to` block on the new section displays the correct version badge.


gigerdo (approved):
LGTM, thank you for adding more docs on this 👍

deploy-manage/monitor/orchestrators/ece-monitoring-ece-set-retention.md
```diff
@@ -23,6 +23,8 @@ You might need to adjust the retention period for one of the following reasons:
 Before increasing retention, ensure the `logging-and-metrics` system deployment has sufficient resources and disk capacity. Longer retention increases storage usage and cluster workload, and can result in a busy or overloaded cluster if the deployment is not scaled appropriately. Refer to [ECE system deployments configuration](/deploy-manage/deploy/cloud-enterprise/system-deployments-configuration.md) for more information.
 ::::
 
+To customize the retention period, create a new {{ilm-init}} policy with the required settings and apply it to the target data streams. On ECE 4.1.1 or later, apply it through a [component template](#customize-retention-component-templates), which is the recommended approach. On earlier versions, you can [clone the relevant index template](#customize-retention-index-templates) and configure it to use your custom {{ilm-init}} policy, though this requires repeating the procedure after upgrades that change template names.
```
shainaraskas: if you expand the link target it will be clearer that you're linking down to a procedure specific to the retention period, rather than a definition of / more information on component templates generically

```suggestion
To customize the retention period, create a new {{ilm-init}} policy with the required settings and apply it to the target data streams. On ECE 4.1.1 or later, [apply it through a component template](#customize-retention-component-templates), which is the recommended approach. On earlier versions, you can [clone the relevant index template](#customize-retention-index-templates) and configure it to use your custom {{ilm-init}} policy, though this requires repeating the procedure after upgrades that change template names.
```

deploy-manage/monitor/orchestrators/ece-monitoring-ece-set-retention.md
```diff
@@ -39,13 +41,63 @@ The following list contains the names of the most relevant index templates and d
...
 ::::
 
-## Customize retention period
+## Customize retention using component templates [customize-retention-component-templates]
+```{applies_to}
+deployment:
+  ece: ga 4.2+
+```
+
+Starting from ECE 4.1.1, each index template in the `logging-and-metrics` cluster includes a `composed_of` array that references a set of reserved component template names. If you create a component template with one of these names, {{es}} merges its settings into new backing indices automatically. Because these names do not include a version tag, the customization persists across ECE upgrades without any additional action.
```
shainaraskas: I think that having 4.2 and 4.1.1 here is not ideal

would prefer either
* no version in body, stay with 4.2 applicability
* 4.1 applicability, specify 4.1 in body

current version makes one seem wrong, rather than looking like we're adding precision by specifying a patch

---

re: the content, can we make the benefit statement a little punchier? right now it reads as implementation details

maybe 



````suggestion
```{applies_to}
deployment:
  ece: ga 4.1+
```

Create a component template with a reserved name to apply custom retention that remains in place after ECE upgrades.

Each index template in the `logging-and-metrics` cluster includes a `composed_of` array that references these reserved names. If you create a matching component template, {{es}} merges its settings into new backing indices. The names do not include a version tag, so they continue to match after an ECE upgrade.

```{note}
This functionality is only available in ECE 4.1.1 and higher.
```

````

File: deploy-manage/monitor/orchestrators/ece-monitoring-ece-set-retention.md [hunk 3]
```diff
@@ -39,13 +41,63 @@ The following list contains the names of the most relevant index templates and d
...
+
+The following component template names are reserved for customization. Define only the one that matches the scope you need:
```
shainaraskas: I assume you can define as many as you want, right? this reads as you can only define one, but maybe a couple can coexist? 


```suggestion
The following component template names are reserved for customization. Define only the templates that match the scope you need:
```

shainaraskas (approved):
general lgtm, with some suggestions to consider

deploy-manage/monitor/orchestrators/ece-monitoring-ece-set-retention.md
eedugon: great catch

PR #8166, opened 2026-09-02
Mark BYOK for existing deployments as GA

alxchalkias:
## Summary

Encrypting an **existing** Elastic Cloud Hosted deployment with a customer-managed key (BYOK) is now generally available. This removes the tech preview labelling and the preview-era wording from [Use a customer-managed encryption key](https://www.elastic.co/docs/deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key).

Tracking issue: [elastic/roadmap#40](https://github.com/elastic/roadmap/issues/40)

## Changes

- Removed the section-level `applies_to: ess: preview` badge from **Encrypt an existing deployment with your key**. The section now inherits the page-level `ess: ga`, so no redundant GA badge is added.
- Reworded the three per-CSP tips (AWS, Azure, Google Cloud) that described creating a deployment from a snapshot as a *workaround* "until encrypting existing deployments with a customer-managed key is supported". That framing is obsolete now that the in-place flow is GA. The tips keep the snapshot-restore option and now cross-link to the existing-deployment section instead.

No other content changes. The remaining forward-looking statements on the page are about *changing or removing* a key, which is still unsupported, so they are left as-is.

## Test plan

- [ ] Confirm the **Encrypt an existing deployment with your key** section renders without a Preview badge
- [ ] Confirm the three reworded tips render and the anchor link `#ec_encrypt_an_existing_deployment_with_a_customer_managed_key` resolves

### DIFF
--- a/deploy-manage/deploy/elastic-cloud/azure-native-isv-service-byok.md
+++ b/deploy-manage/deploy/elastic-cloud/azure-native-isv-service-byok.md
@@ -1,7 +1,7 @@
 ---
 applies_to:
   deployment:
-    ess: preview
+    ess: ga
 products:
   - id: cloud-hosted
 navigation_title: Customer-managed encryption keys
--- a/deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key.md
+++ b/deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key.md
@@ -276,7 +276,7 @@ You can encrypt a deployment with your customer-managed key when you [create a n
             ```
 
             ::::{tip}
-            You can also create the deployment from a snapshot of a deployment that was initially not encrypted with a customer-managed key. You can use this as a workaround to encrypt existing data under new deployments using your key, until encrypting existing deployments with a customer-managed key is supported.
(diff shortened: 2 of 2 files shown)


eedugon (commented):
Changes look good!
But we also need to remove the tech preview badge at `deploy-manage/deploy/elastic-cloud/azure-native-isv-service-byok.md`.

alxchalkias:
Thanks @eedugon — good catch, that page was missed. Addressed in e6cefd8.

One note on the approach, since it differs slightly from the other page in this PR: on `azure-native-isv-service-byok.md` the `preview` marker is the **page-level** `applies_to`, so I changed it to `ess: ga` rather than deleting the block. Removing it outright would drop the ECH applicability badge from the page entirely. That now matches the frontmatter on `encrypt-deployment-with-customer-managed-encryption-key.md`, which is already `ess: ga`.

On that page the badge I deleted was a redundant *section-level* override, which is why it could be removed without a replacement.

The page marker made sense to update because the whole page exists for the existing-deployment path — "BYOK is not available as a configuration option when creating a deployment from the Azure portal. You can add a customer-managed key to your deployment after it is created" — so it was in preview alongside the flow that is now GA.

I also swept the repo for any other `ess: preview` on pages mentioning BYOK/CMEK and found none, so I believe these two are the complete set.

eedugon (approved):
LGTM!

Sorry I said before to remove the technical preview statement on the Azure dedicated page. I meant to update it to GA, as you have done. My bad =)

PR #8177, opened 2026-09-02
Remove admin and dev projects from add-to-board automation

shainaraskas:
## Summary
The core docs team uses [project-based workflows](https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/adding-items-automatically) to add items to our board from this repo, so this automation is not needed. It is also adding items to retired boards currently. Removed our projects, in case the others are still needed

## Generative AI disclosure

- [ ] Yes  
- [x] No

### DIFF
--- a/.github/workflows/add-to-board.yml
+++ b/.github/workflows/add-to-board.yml
@@ -29,8 +29,6 @@ jobs:
           "Team:Ingest":      "1034",
           "Team:Experience":  "1034",
           "Team:SKI":         "1034",
-          "Team:Admin":       "1927",
-          "Team:Developer":   "726",
           "Team:DocsEng":     "1625",
           "Team:Platform":    "1232",
           "Team:Projects":    "1415",


.github/workflows/add-to-board.yml
```diff
@@ -29,8 +29,6 @@ jobs:
           "Team:Ingest":      "1034",
           "Team:Experience":  "1034",
           "Team:SKI":         "1034",
-          "Team:Admin":       "1927",
-          "Team:Developer":   "726",
```
bmorelli25, line 33: Are you sure you don't want to keep those and just update the numbers to 1232 like we did for ingest and exp above (pointing to 1034 which is the ski board)?

shainaraskas: we use the board workflows to pull issues from the relevant projects, so this ends up being duplicative for us

image
image

bmorelli25: Ah so this whole workflow is duplicative for you then.

bmorelli25: I see. We can do the same on my end and then just remove this file

bmorelli25: I guess we'll still need the remove label action. I'll look into this in the future because we can definitely simplify. Thanks!

theletterf (approved):

PR #8208, opened 2026-09-04
Add documentation on scaling integration policies

### LINKED ISSUE #2166: [Internal]: Fleet Managed Integration Improvement to help with scale
### Description

What: Improved documentation to help customers scale without having hundreds of integration policies. 
When: As soon as it aligns with resourcing and upcoming Doc releases/updates. 
Why: The current documentation does not provide a clear directive on how users can consolidate/minimise the number of integration policies they have to create. 

For example, we have customers who wonder if they need to create 200 Oracle integrations to monitor their Oracle DB fleet, as they want to avoid the integration being unhealthy when they set multiple IPs, each of which can only reach itself. 

It would be good to have something like [this](https://support.elastic.dev/knowledge/view/2bae10d6) part of the documentation to help customers reduce the number of integrations.

[Internal Slack thread](https://elastic.slack.com/archives/C0JF80CJZ/p1752673722213789?thread_ts=1752669683.337189&cid=C0JF80CJZ) that kicked this off.

### Resources

KB on how to work around this
https://support.elastic.dev/knowledge/view/2bae10d6

### Feature differences

Affects Both Cloud (stateful + Serveless as users can still deploy their own agents) and On-Prem. 

N/A

### Serverless release

N/A

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @

**Stakeholders:**

vishaangelova:
## Summary

Customers who monitor many hosts running the same service had no documented way to avoid creating one integration policy per host. The issue cites a customer asking whether 200 Oracle databases require 200 Oracle integrations, and reports that when they instead list all 200 hosts in a single integration policy, the integration turns unhealthy. Nothing in the docs explained why that happens or what to do instead.

This PR adds a new doc, Scale integration policies across many hosts (`scale-integration-policies.md`), under **Fleet and Elastic Agent → Manage integrations**. It documents the technique from the Support knowledge base article linked in the issue (embedding a provider variable in the integration setting so that one policy serves every host), explains why a shared host list doesn't work, and covers the secondary consolidation options: sharing an integration policy across agent policies and grouping agents by role.

Closes #2166

### Other changes

* Folds `fleet-agent-environment-variables.md` into the new page and deletes it. The stub only said "use the env provider" and sat under *Agent policies*; the new page is the same topic with the motivation and procedure the stub lacked, so keeping both would split one topic across two pages. Nothing else in the repo linked to it. The published URL is preserved through `mapped_pages` and a `redirects.yml` entry that maps the old anchor to `#reuse-one-integration-policy-with-variables`.
* Adds entry points from the places readers hit this problem: both scaling-limit sections in `agent-policy.md`, the end of the add procedure in `add-integration-to-policy.md`, the `manage-integrations.md` table, and the Fleet-managed sections of `providers.md` and `env-provider.md`.

### Verification

The Support KB article referenced in the issue (`support.elastic.co/knowledge/2bae10d6`) is the source for the technique itself: the embedded `${env.ORACLE_DB_ADDRESS}` pattern in **Oracle DSN**, the systemd override with `Environment=` or `EnvironmentFile=`, and verifying provider values through the diagnostics bundle. The KB also describes the local provider on Fleet-managed agents; that's deliberately left out, see the note below. Product claims were then checked against source rather than the issue or KB text.

<details>
<summary>Source checks</summary>

* `title: Oracle DSN` on the `hosts` variable, `multi: true`, default `oracle://0.0.0.0:1521/ORCLCDB.localdomain?sysdba=1`; `password` is `type: password`, `secret: true` — `elastic/integrations`, `packages/oracle/manifest.yml`. The multi-value field is what makes the anti-pattern tempting, which is why the page leads with it, and the secret flag is what backs the guidance to keep credentials in the integration's own fields.
* Fleet passes `${env.X}` through to the agent unchanged. Integration templates are compiled with Handlebars `noEscape: true` and there is no special handling of `$` — `elastic/kibana`, `fleet/server/services/epm/agent/agent.ts`.
* **New hosts** / **Existing hosts** tabs — `fleet/public/.../steps/step_select_hosts.tsx`.
* Reusable integration policies gated on an Enterprise license — `fleet/public/hooks/use_multiple_agent_policies.ts`, `LICENCE_FOR_MULTIPLE_AGENT_POLICIES = 'enterprise'` in `fleet/common/constants/package_policy.ts`.
* **Actions** → **View policy** — agent policy `actions_menu.tsx`.
* `elastic-agent inspect --variables` — `elastic/elastic-agent`, `internal/pkg/agent/cmd/inspect.go`.
* Provider *settings* (as opposed to provider *variables*) can't be managed from Fleet. There's no UI field for them, the API-only agent policy `overrides` is documented as not for routine use (`fleet/server/types/models/agent_policy.ts`), and on the agent side the providers controller is built once from the local `elastic-agent.yml` (`internal/pkg/agent/application/application.go`, `composable.New`) with no reload path from the policy (`internal/pkg/composable/controller.go`). The only channel is editing `elastic-agent.yml` on each host, which is not a workflow to recommend to Fleet users, so the KB's local-provider technique is left out of the page. Environment variables cover the same per-host need.
* `variables.yaml` and `computed-config.yaml` in the diagnostics bundle — diagnostics hooks in `internal/pkg/agent/application/coordinator/coordinator.go`; both files are already listed in `agent-command-reference.md#elastic-agent-diagnostics-command`. The KB says `variables.yaml` sits in the agent install directory; source shows it's produced only as part of diagnostics, so the page describes it that way.
* Undefined variable removes the input, `|` fallback chaining, `$$` escaping — consistent with `dynamic-input-configuration.md`.
* Policy limits (1000 / 500 agent policies, 10,000 integration policies per agent policy) — consistent with `agent-policy.md`.

</details>

Vale is clean apart from the expected `Elastic.MappedPages` warning, which is there because the new page takes over the deleted page's legacy URL. `docs-builder --strict` reports no errors or hints for the page.

### Follow-up

Integration policy conditions (new in 9.5) are a fourth consolidation technique, but they're currently undocumented for Fleet users anywhere. Rather than version-gate part of an otherwise cumulative page, that work is left for a separate PR that fixes `dynamic-input-configuration.md#conditions` — which is where the product's own **Learn more** link points — and then adds the technique here.

## Generative AI disclosure

- [x] Yes
- [ ] No

2. Tool(s) and model(s) used: Cursor (Claude Opus 5, Claude Fable 5.1)


reference/fleet/providers.md
```diff
@@ -106,4 +106,6 @@ providers:
 
 On {{fleet}}-managed {{agent}}, you can use provider variables in integration policy settings (for example, `${host.name}`, `${env.foo}`, `${agent.id}`), but you cannot add a `providers` configuration block directly through the {{fleet}} UI.
```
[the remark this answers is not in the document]
vishaangelova: I used Claude to check how provider settings can reach a Fleet-managed agent. Here's what we found out:

- Kibana has no UI field for a `providers` block on agent policies (the **Advanced YAML configuration** field exists only on outputs). The API-only agent policy `overrides` is documented as not for routine use.
- On the agent side, the providers controller is built once at startup from the local `elastic-agent.yml` (`composable.New` in `application.go`) and has no reload path from the Fleet policy, so a `providers` block delivered via Fleet would be ignored anyway.

That leaves editing `elastic-agent.yml` on each host as the only option, which isn't a workflow we want to recommend to Fleet users, and environment variables already cover the per-host need on this page.

I removed the local-provider row and example from `scale-integration-policies.md` in https://github.com/elastic/docs-content/pull/8208/commits/807aeb478cb1e891b4ac22058c3098a5f32d6f3c. There should be no contradiction between the two pages now.

vishaangelova:
@bturquet, could you suggest an SME from your team who could review this PR? Thanks!

reference/fleet/scale-integration-policies.md
```diff
@@ -0,0 +1,158 @@
...
+* [Group agents by role](#group-agents-by-role), so that your policies match how your infrastructure is organized.
+
+
+## Why a shared host list doesn't work [why-a-shared-host-list-doesnt-work]
+
+Many integrations accept more than one value in their host field, so listing every host in a single integration policy looks like a shortcut. It isn't, because of how {{fleet}} distributes the policy.
+
+An integration policy applies to *every* {{agent}} enrolled in the {{agent}} policy that contains it. If you list 200 database hosts, each of the 200 agents receives the full list and tries to connect to all 200 databases, not only the one running on its own host.
+
+Agents that can't reach the other databases report failed connections, and the integration shows as unhealthy. Agents that can reach them collect the same data from every database, so you store 200 copies of every metric.
```
lucabelluccini, line 33: Nit: while it's true each EA would collect the same metrics, if the destination data streams are TSDS, most of the duplicate events (determined by the dimensions of the metric and the timestamp) will be rejected by Elasticsearch.

The waste of the resources to run 200 EA is real - as well as the resources on ES to reject the events.

It might be acceptable to have 2 EA pulling the same metrics as a poor attempt to implement HA.

reference/fleet/scale-integration-policies.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Reuse one integration policy with variables [reuse-one-integration-policy-with-variables]
+
+Instead of hardcoding a value in an integration policy, you can reference a variable that each {{agent}} resolves at runtime. {{agent}} [providers](/reference/fleet/providers.md) supply these variables, and they're available to {{fleet}}-managed agents without any setup.
+
+The most useful providers for this purpose are:
+
+| Provider | Example variables | Use it for |
+| --- | --- | --- |
+| [Env](/reference/fleet/env-provider.md) | `${env.VAR_NAME}` | Values you define per host as environment variables, such as a database address or a log directory. |
```
lucabelluccini, line 46: I would mention env vars will need to be deployed by the user via other means (EA cannot deploy env vars), like configuration management tools.

reference/fleet/scale-integration-policies.md
```diff
@@ -0,0 +1,158 @@
...
+5. Restart {{agent}} on the host so that it picks up the new environment variable.
+
+Each agent resolves `${env.ORACLE_DB_ADDRESS}` to its own database address, so a single integration policy covers all 200 hosts. To add another database, define the environment variable on the new host and enroll it in the same {{agent}} policy. The policy itself doesn't change.
+
+Keep credentials in the integration's own fields, such as **Username** and **Password**, rather than in a variable. {{fleet}} stores those fields as secrets, whereas an environment variable is readable by anyone who can inspect the service configuration or the process environment on the host.
+
+These steps use a connection string, but the same approach works for any integration setting that differs between hosts, and for any of the variables listed in the preceding table. For example, set a filestream path to `${env.APP_LOG_DIR}/app.log` when applications log to different directories, or to `/var/log/${host.name}/app.log` to build the path from the hostname itself.
+
+
+### Set a fallback value [set-a-fallback-value]
```
lucabelluccini, line 102: Good!

reference/fleet/scale-integration-policies.md
```diff
@@ -0,0 +1,158 @@
...
+description: Use one Fleet integration policy for many hosts by referencing Elastic Agent variables, sharing policies across agent policies, and grouping agents by role.
+applies_to:
+  stack: ga
+  serverless: ga
+products:
+  - id: fleet
+  - id: elastic-agent
+---
+
+# Scale integration policies across many hosts [scale-integration-policies]
```
lucabelluccini, line 14: All the content added is generally valid. 

We will need to be careful and address some more advanced concepts like Integrations which can be scaled horizontally. Example at https://github.com/elastic/package-spec/issues/842

bturquet:
@criamico @juliaElastic can you help @vishaangelova with the review ? thanks

juliaElastic:
The Fleet related content LGTM. About the Agent / envvar related parts I would defer to Agent team cc @ebeahan

reference/fleet/scale-integration-policies.md
vishaangelova: In https://github.com/elastic/docs-content/pull/8208/changes/ea28e93e78303082ce63b99732bf45ab60e020a3, I added a short note in the introduction that some integrations collect from a shared source and that adding agents increases throughput without duplicate data.

Documenting the scaling model for each integration is tracked in https://github.com/elastic/docs-content/issues/7761, the follow-up from elastic/package-spec#842. We'll cover that there.

reference/fleet/scale-integration-policies.md
vishaangelova: Addressed in https://github.com/elastic/docs-content/commit/ea28e93e78303082ce63b99732bf45ab60e020a3

reference/fleet/scale-integration-policies.md
vishaangelova: Addressed in https://github.com/elastic/docs-content/commit/ea28e93e78303082ce63b99732bf45ab60e020a3

leehinman (approved):
LGTM

PR #1903, opened 2025-06-24
Update data tier _tier_preference to remove "available" as ambiguous  

stefnestor:
👋 howdy! Per feedback in [elasticsearch#129812](https://github.com/elastic/elasticsearch/issues/129812#issuecomment-2997687822), updates [this section](https://www.elastic.co/docs/manage-data/lifecycle/data-tiers#data-tier-allocation) to remove ambiguity.

PR #8213, opened 2026-09-05
Add data tier removal guidance for self-managed and ECK

### LINKED ISSUE #6186: Add information about removing a data tier for ECK/self-managed
[Manage data tiers in self-managed and ECK](https://www.elastic.co/docs/manage-data/lifecycle/data-tiers/manage-data-tiers-self-managed-eck) could be improved with adding instructions on how to remove data tiers in these deployment types, similar to how we describe this task on the [Manage data tiers in ECH or ECE](https://www.elastic.co/docs/manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece#remove-data-tier-ech-ece) page.

For more information, see Stef's suggestion in https://github.com/elastic/docs-content/pull/6092#issuecomment-4315892256

eedugon:
## Summary

- Add the data tier management and removal procedure for self-managed deployments and ECK, based on the existing ECH/ECE guidance.
- Apply the shared improvements to both guides while preserving deployment-specific orchestration and UI steps.

## Main procedure improvements

- Rewrite the section introductions and preflight guidance to clearly distinguish regular indices, fully mounted snapshots, partially mounted snapshots, and mixed tiers.
- Clarify how `_tier_preference`, legacy `require` rules, and custom allocation filters affect relocation, including the difference between relocation triggered by index settings and by the deployment or orchestration plan.
- Make ILM handling safer by waiting for it to stop, reviewing policies before restart, and verifying its status and errors afterwards.
- Strengthen the searchable snapshot workflow with hidden-index discovery, support for non-standard names, snapshot and repository inventory, safer restore settings, restored-index verification, and optional source snapshot deletion with dependency warnings.
- Add clearer capacity, disk watermark, autoscaling, shard-monitoring, and final cluster-health checks.

Closes https://github.com/elastic/docs-content/issues/6186

## Validation

- `git diff --check`
- Vale (only existing or intentional Cloud terminology warnings)

### DIFF
--- a/manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
+++ b/manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
@@ -44,7 +44,9 @@ Review [{{es}} data tiers](/manage-data/lifecycle/data-tiers.md) so you choose t
 
 ## Remove a data tier [remove-data-tier-ech-ece]
 
-Follow this section when you need to remove a data tier from an {{ech}} or {{ece}} deployment. The steps differ depending on whether the tier holds [regular indices](#non-searchable-snapshot-data-tier) or [{{search-snap}}](#searchable-snapshot-data-tier) indices (typical for cold or frozen when using {{ilm}} ({{ilm-init}})).
+Follow this section when you need to remove the warm, cold, or frozen tier from an {{ech}} or {{ece}} deployment. The shared hot and content tier is required and cannot be removed.
+
+The steps differ depending on whether the tier holds [regular indices](#non-searchable-snapshot-data-tier) or [{{search-snap}}](#searchable-snapshot-data-tier) indices (typical for cold or frozen when using {{ilm}} ({{ilm-init}})).
 
 ### Before you remove a data tier [before-you-remove-a-data-tier]
 
@@ -56,30 +58,52 @@ To avoid this, especially for [production environments](/deploy-manage/productio
 * Review the disk size, CPU, JVM memory pressure, and other [performance metrics](/deploy-manage/monitor/access-performance-metrics-on-elastic-cloud.md) of your deployment **before** attempting to perform the scaling down action.
 * Make sure that you have enough resources and [availability zones](/deploy-manage/production-guidance/availability-and-resilience.md) to handle your workloads after scaling down.
 * Check that your [deployment hardware profile](/deploy-manage/deploy/elastic-cloud/ec-change-hardware-profile.md) (for {{ech}}) or [deployment template](/deploy-manage/deploy/cloud-enterprise/configure-deployment-templates.md) (for {{ece}}) is correct for your business use case. For example, if you need to scale due to CPU pressure increases and are using a *Storage Optimized* hardware profile, consider switching to a *CPU Optimized* configuration instead.
+* Review [disk watermarks](/troubleshoot/elasticsearch/fix-watermark-errors.md) and ensure the remaining nodes are not close to their limits.
 
 Read [https://www.elastic.co/cloud/shared-responsibility](https://www.elastic.co/cloud/shared-responsibility) for additional details.
 If in doubt, reach out to Support.
 :::
 
-* Know whether you are disabling a tier that stores regular indices or {{search-snaps}}. The frozen tier only holds [partially mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#partially-mounted) {{search-snaps}}. The cold tier can hold regular indices or [fully mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#fully-mounted) {{search-snaps}}. The hot tier usually holds regular indices, but an {{ilm-init}} policy can mount a fully mounted {{search-snap}} on the hot tier (for example, when the [searchable_snapshot](elasticsearch://reference/elasticsearch/index-lifecycle-actions/ilm-searchable-snapshot.md) action runs in the `hot` phase). Use these requests to check for {{search-snap}} indices on the tier you are removing:
+Check whether the tier you are removing holds regular indices, [{{search-snaps}}](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md), or both. Use the guidance for the tier you are removing:
+
+* **Warm tier:** This tier typically holds regular indices. Follow [Remove a tier with regular indices](#non-searchable-snapshot-data-tier) unless you have manually mounted {{search-snaps}} on the tier.
+* **Cold tier:** This tier can hold regular indices or [fully mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#fully-mounted) {{search-snaps}}. Check for standard {{ilm-init}}-managed {{search-snap}} indices:
 
     ```sh
-    # cold data tier: {{search-snap}} indices
-    GET /_cat/indices/restored-*
-
-    # frozen data tier: {{search-snap}} indices
-    GET /_cat/indices/partial-*
+    GET /_cat/indices/restored-*?expand_wildcards=all
     ```
 
-* To learn more about {{ilm-init}} or shard allocation filtering, see [Create your index lifecycle policy](/manage-data/lifecycle/index-lifecycle-management/configure-lifecycle-policy.md), [Managing the index lifecycle](/manage-data/lifecycle/index-lifecycle-management.md), and [Shard allocation filters](/deploy-manage/distributed-architecture/shard-allocation-relocation-recovery/index-level-shard-allocation.md).
+    For each returned index, check its [current data tier preference](/manage-data/lifecycle/data-tiers.md#data-tier-allocation-value) to determine whether it is on the tier you are removing.
+
+    Exclude any fully mounted indices associated with the hot tier from the removal inventory. The hot tier is required and is not removed by this procedure.
+
+* **Frozen tier:** This tier only holds [partially mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#partially-mounted) {{search-snaps}}. Check for standard {{ilm-init}}-managed indices:
+
+    ```sh
+    GET /_cat/indices/partial-*?expand_wildcards=all
+    ```
+
+:::{note}
+Manually mounted {{search-snaps}} might not use the standard `restored-*` or `partial-*` prefixes. If you mounted snapshots manually, adapt the index names or patterns in these requests to match your configuration.
+:::
+
+* If the tier does not contain any {{search-snap}} indices, follow [Remove a tier with regular indices](#non-searchable-snapshot-data-tier).
+* If the tier contains {{search-snap}} indices, review [Remove a tier with {{search-snaps}}](#searchable-snapshot-data-tier) and select the appropriate procedure based on how the indices are mounted. If regular indices also remain, restore or move the {{search-snap}} indices first, but do not disable the tier yet. Then return to the regular indices procedure.
+
+To learn more about {{ilm-init}} or shard allocation filtering, refer to [Create your index lifecycle policy](/manage-data/lifecycle/index-lifecycle-management/configure-lifecycle-policy.md), [Managing the index lifecycle](/manage-data/lifecycle/index-lifecycle-management.md), and [Shard allocation filters](/deploy-manage/distributed-architecture/shard-allocation-relocation-recovery/index-level-shard-allocation.md).
 
 ### Remove a tier with regular indices [non-searchable-snapshot-data-tier]
 
-The frozen tier only stores [partially mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#partially-mounted) {{search-snaps}}. [Fully mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#fully-mounted) {{search-snaps}} can be allocated to the hot or cold tier depending on the {{ilm-init}} phase, while the cold tier can also hold regular indices. Use the checks in [Before you remove a data tier](#before-you-remove-a-data-tier) if you are unsure what is on the tier.
+This section covers the removal of a tier that holds regular indices. The goal is to ensure all shard allocation rules allow the data to move to other tiers before you disable the tier. You also need to temporarily stop {{ilm-init}} to prevent new indices from being routed to the tier while you work.
 
-When you update the deployment, {{ech}} and {{ece}} try to move all data from the nodes that are removed. To disable a tier that holds only regular indices, make sure that all data on that tier can be re-allocated by reconfiguring the relevant shard allocation filters. You’ll also need to temporarily stop your {{ilm-init}} policies to prevent new indices from being moved to the data tier you want to disable.
+:::{note}
+If the tier also holds [fully mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#fully-mounted) {{search-snaps}}, you have two options:
 
-To make sure that all data can be migrated from the data tier you want to disable, follow these steps:
+* **To keep them as {{search-snaps}} on another tier**: apply the same steps in this section. Fully mounted {{search-snaps}} follow the same shard placement rules as regular indices and can be moved by updating their allocation settings.
+* **To restore them to regular indices on another tier**: follow [Remove a tier with {{search-snaps}}](#searchable-snapshot-data-tier) to restore the indices and delete the original {{search-snap}} indices, but do not disable the tier yet. Then return to this section to move any regular indices and disable the tier. You can optionally delete source snapshots that you no longer need.
+:::
+
+When you update the deployment, {{ech}} and {{ece}} try to move all data from the nodes that are removed. Before applying this change, make sure that the relevant shard allocation filters allow the data to move.
 
 1. Determine which nodes will be removed from the cluster.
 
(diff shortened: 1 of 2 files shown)


eedugon:
@stefnestor : I'm adding you here as a reviewer because you also collaborated when we created the ECH/ECE version of the procedure.

Note that I didn't merge all into a single document because the resulting document would need a lot of applies switches and part of the narrative should be modified due to operational differences between Elastic Cloud + ECE and self-managed systems and ECK, but this could be evaluated in the future.

Also note that I found a few logical and technical caveats in the original procedure and I've tried to improve them.

Hoping you like the result Stef and @yetanothertw

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -56,30 +58,52 @@ To avoid this, especially for [production environments](/deploy-manage/productio
...
 
 ### Remove a tier with regular indices [non-searchable-snapshot-data-tier]
 
-The frozen tier only stores [partially mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#partially-mounted) {{search-snaps}}. [Fully mounted](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#fully-mounted) {{search-snaps}} can be allocated to the hot or cold tier depending on the {{ilm-init}} phase, while the cold tier can also hold regular indices. Use the checks in [Before you remove a data tier](#before-you-remove-a-data-tier) if you are unsure what is on the tier.
+This section covers the removal of a tier that holds regular indices. The goal is to ensure all shard allocation rules allow the data to move to other tiers before you disable the tier. You also need to temporarily stop {{ilm-init}} to prevent new indices from being routed to the tier while you work.
```
stefnestor, line 97: We in Support do not disable ILM 👻. Especially where drain off time can be extensive & you might full disk hot nodes when you can't stop ingest.

Instead we normally recommend users exclude nodes, ex:
```
PUT _cluster/settings
{"transient": {"cluster": {"routing": {"allocation": {"exclude": {"_name": "instance-0000000043"} } } } } }
```

ref: https://www.elastic.co/docs/reference/elasticsearch/configuration-reference/cluster-level-shard-allocation-routing-settings#shards-rebalancing-heuristics 

Note: this will incur ILM errors (and if you're hot>cold>frozen and removing frozen then agreed you'd still fill up hot, which is why we tell users to disable tiers the same as you'd rolling restart: from lowest-to-highest temp).

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -137,33 +164,41 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
+                        "_tier_preference": "data_warm,data_hot" <1>
                     }
                 }
             }
         ...
         }
         ```
+        1. The example represents an index in the `warm` tier.
+
+        Before disabling the tier, update `_tier_preference` so that the tier where you want the data to move is the first available tier in the list. This allows {{es}} to begin relocating the shards before the deployment plan removes the tier.
```
stefnestor, line 194: TBC shards will not allocate to a fallback list-item unless/until _no_ nodes of preceding exist in cluster. So this sentence reads catch-22 IMO. You must update `_tier_preference` to migrate shards off current deprecating tier, it is not to "begin relocating" early, it is a required step.

ref: https://www.elastic.co/docs/manage-data/lifecycle/data-tiers#data-tier-allocation

"This setting also accepts multiple tiers in order of preference. This prevents indices from remaining unallocated if there are no nodes in the cluster for the preferred tier."

rel: https://github.com/elastic/docs-content/pull/1903

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -178,13 +213,15 @@ To make sure that all data can be migrated from the data tier you want to disabl
         }
         ```
 
-        1. If the cluster has no cold tier, use the lowest remaining tier in order of preference, for example `data_hot` when no `data_cold` nodes exist. The frozen tier is for partially mounted {{search-snaps}} only, not a destination for regular indices in this flow.
+        1. You can also use `data_cold,data_hot`. Both values move the data to cold, but omitting `data_warm` removes that tier from the fallback sequence.
 
-        Updating allocation inclusion rules will trigger a shard re-allocation, moving the shards from the nodes to be removed.
+        :::{note}
+        Do not use the frozen tier as a fallback for regular indices. It is reserved for partially mounted {{search-snaps}}.
```
stefnestor, line 219:
```suggestion
        Do not use the frozen tier as a fallback for regular nor fully mounted indices. It is reserved for partially mounted {{search-snaps}}.
```

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -178,13 +213,15 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
+        :::
 
-    2. Updating node attribute allocation requirement rules.
+    2. Update node attribute allocation requirement rules.
```
stefnestor, line 222: Potential noop comment : This is considered user customization responsible and not default anymore. This is why https://www.elastic.co/docs/manage-data/lifecycle/data-tiers#data-tier-allocation-troubleshooting was simplified down to only ~"by now you should have migrated to data tiers & ensure you don't induce allocation conflicts". From Support's side, we usually say, "update tier preferences, let shards drain off, check allocation explain for anybody stubborn not moving off to see why".

This is the first common example: https://www.elastic.co/docs/troubleshoot/elasticsearch/cluster-allocation-api-examples#conflicting-settings

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -232,25 +269,27 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
+    If your allocation setting changes start relocation, wait until [shard allocation and recovery](/deploy-manage/distributed-architecture/shard-allocation-relocation-recovery.md) finish. Use `GET /_cat/allocation?v=true&s=node` to monitor the nodes that the plan will remove. Shards might remain if you only removed a `require` rule because that change does not force them to move. The deployment plan relocates them when it disables the tier.
+
+    If shards that you expect to move remain on the original tier, use the [cluster allocation explain]({{es-apis}}operation/operation-cluster-allocation-explain) API to determine the cause. Common causes include [disk watermarks](/troubleshoot/elasticsearch/fix-watermark-errors.md) or the [`index.routing.allocation.total_shards_per_node`](elasticsearch://reference/elasticsearch/index-settings/total-shards-per-node.md#total-shards-per-node) limit on the destination nodes.
```
stefnestor, line 292: These are the common examples: https://www.elastic.co/docs/troubleshoot/elasticsearch/cluster-allocation-api-examples

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -259,131 +298,170 @@ To make sure that all data can be migrated from the data tier you want to disabl
 
     Any remaining shards on the tier being disabled are re-allocated across the remaining cluster nodes while applying the plan to disable the data tier. Monitor shard allocation during the data migration phase to ensure all allocation rules have been correctly updated. If the plan fails to migrate data away from the data tier, then re-examine the allocation rules for the indices remaining on that data tier.
 
-6. Once the plan change completes, confirm that there are no remaining nodes associated with the disabled tier and that `GET _cluster/health` reports `green`. If this is the case, re-enable {{ilm-init}}.
+6. Once the plan change completes, confirm that `GET /_cat/nodes?v` shows no nodes associated with the disabled tier and that `GET /_cluster/health` reports `green`.
+
+7. Review your {{ilm-init}} policies and consider removing references to the disabled tier to keep them consistent with the deployment topology. This is especially important in older deployments where {{ilm-init}} uses node-attribute-based allocation, as those policies cannot run phases that target nodes that no longer exist.
```
stefnestor, line 303: FYI: Should probably do this before you start. 

👻 Also need to review/modify index+legacy+component templates

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -259,131 +298,170 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
     ```
 
+9. Verify that {{ilm-init}} is running and that no indices report errors related to the disabled tier:
+
+    ```sh
+    GET /_ilm/status
+    GET /_all/_ilm/explain?human=true&expand_wildcards=all&only_errors=true
+    ```
+
+    Confirm that `operation_mode` is `RUNNING`. Investigate any reported errors and verify that no policy still attempts to allocate data to the disabled tier.
```
stefnestor, line 320: Ref : https://www.elastic.co/docs/troubleshoot/elasticsearch/index-lifecycle-management-errors#ilm-steps-errors 

If you just ran a plan, you likely just start by doing a global Retry API

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -259,131 +298,170 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
+
+* **[Partially mounted {{search-snaps}}](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#partially-mounted) on the frozen tier:** The only way to keep the data available as indices when removing the frozen tier is to restore all partially mounted indices as regular indices on another tier. Follow the steps in this section to restore the indices and remove the original {{search-snap}} indices.
+* **[Fully mounted {{search-snaps}}](/deploy-manage/tools/snapshot-and-restore/searchable-snapshots.md#fully-mounted) on the cold tier:** To keep the indices as {{search-snaps}}, move them to another tier by following [Remove a tier with regular indices](#non-searchable-snapshot-data-tier). This works because fully mounted indices follow the same shard placement rules as regular indices. Alternatively, follow the steps in this section to restore them as regular indices on another tier.
+
+If you do not need to preserve the data, delete the {{search-snap}} indices and continue from the shard verification before disabling the tier.
```
stefnestor, line 329: You'd also delete the index-backing searchable snapshot. If you only delete the index, you'll leave stray data in your repository (which adds to your bill/disk).

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -259,131 +298,170 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
 
-
-    In the example we are removing the alias for the `frozen-index-1` index.
-
     :::{image} /manage-data/images/cloud-enterprise-ec-ce-remove-tier-remove-alias.png
     :alt: A screenshot showing the process of removing a {{search-snap}} index alias
     :::
 
-5. Restore indices from the {{search-snaps}}.
+5. Restore each index in the inventory from its source snapshot.
```
stefnestor, line 384: I'm a bit lost by this; it doesn't seem updated. This is the way old steps Support had raised concern on. Can we cross-compare https://support.elastic.dev/knowledge/view/39606517 instead.  (Ex: you don't add `rollover_alias` you set `indexing_complete` otherwise you'll ILM error 😕)

stefnestor:
👋 @eedugon, thanks for this! It looks great. Added notes about technical truths of certain lines and/or more links to share.

Housekeeping: Apologies. As of the top of this month, my GitHub notification flow has gotten broken. So please feel free to ping me if you don't hear back as soon as you're expecting. Or TBH, if I probably missed anywhere else you or your team pinged me. Sorry! 🙈

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
eedugon: This is interesting because in the original procedure we were also suggesting to stop ILM, so this isn't a change I've introduced.

But I get your point. Stopping ILM for long time can be critical and fill up the hot tier in very active clusters causing serious and unneeded damage.

That's a very good point, I'll find a way to cover it.

This could also be related with the ILM policies review (that I have at the end of the procedure atm). Maybe if the user first removes the tier from the ILM policies first there's no need to stop ILM, and would be the smoothest way to approach this.

So if a user is going to remove let's say the cold tier, I'd suggest to first update the ILM policies to not use cold anymore and then proceed with the procedure without stopping ILM.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
eedugon: @stefnestor , when you say:

> You must update _tier_preference... it is a required step.

Yep, this is a **required*** step, not optional, and we say:

> Before disabling the tier, update `_tier_preference` so that the tier where you want the data to move is the first available tier in the list.

We tell users to ensure the destination tier is the first in the list, and then we give an example of moving data from warm to cold before removing the warm data.

So, I don't get your comment, as I agree with it but I think the text already states it properly. If you want us to rephrase please share some options.

On the other hand:

> shards will not allocate to a fallback list-item unless/until no nodes of preceding exist in cluster. 

That's totally accurate, I was actually thinking on including that type of sentence in a note so people understands how tier_preference really works.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
```diff
@@ -137,33 +164,41 @@ To make sure that all data can be migrated from the data tier you want to disabl
...
         ```
+        1. The example represents an index in the `warm` tier.
+
+        Before disabling the tier, update `_tier_preference` so that the tier where you want the data to move is the first available tier in the list. This allows {{es}} to begin relocating the shards before the deployment plan removes the tier.
 
-        You must remove the relevant tier from the inclusion rules. For example, to disable the warm tier, remove the `data_warm` parameter and set `_tier_preference` to a tier you are keeping. Prefer promoting data through the lifecycle, for example from warm to cold, not back to hot, unless the cluster has no colder tier to accept the data:
+        Update the setting based on where you want to move the data:
+
+        * To move the data to an existing fallback tier, remove the tier being disabled from the list. For example, when disabling the warm tier, change `data_warm,data_hot` to `data_hot`.
+        * To move the data to a later lifecycle tier, add that tier before the tier being disabled. For example, when disabling the warm tier, change `data_warm,data_hot` to `data_cold,data_warm,data_hot`.
```
eedugon: @stefnestor , please read that block. As you can see we don't say this is optional. We tell the user to update the tier_preference setting, and we tell them how exactly in both cases:

- if they want to remove warm and put data in hot.
- if they want to remove warm and put data in cold.

With a clear example IMO.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
eedugon: Anyway I'm updating this section to be clearer and avoid the "this allows {{es}} begin relocating" statement for a clearer one.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
eedugon: Yes, I understand node attribute based allocation shouldn't be in place anymore by ILM, but it might be a user customization so it's worthy to mention it here.

That's why we treat this as possible "custom configuration"

I'll check if it's worthy merging this section with the next step, which is for "other custom allocation rules" (they are both related actually).

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
eedugon: yes, moved to the beginning. I think the approach is much more robust this way, first the user should ensure no future data is going to reach the tier to remove, then they can vacate existing data in the tier and remove the nodes / tier.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
eedugon: Good point, added.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
eedugon: @stefnestor , that article doesn't look aligned with this topic. Maybe wrong paste?

eedugon: I think I found it: https://support.elastic.dev/knowledge/view/68395ecc

From that article I like the call to `_ilm/remove` (we didn't have that) together with the fact that in the article the ILM policy association + rollover_alias are removed from the index during the restore (what we are already doing). I'll be taking that from the article.

But the KB seems to mix two different approaches. I'll try to explain why but let me know if you feel I'm wrong.

If we restore the index without `index.lifecycle.name` and `index.lifecycle.rollover_alias` (both set to `null` as we are already doing), and then call `_ilm/remove` to clear its ILM execution state, `indexing_complete` is not needed at all. And that's the most efficient procedure from my point of view.

`indexing_complete` would be relevant when retaining or reapplying a rollover policy and intentionally marking the restored index as no longer requiring rollover, but it's not the case here. Even then, the rollover alias must be configured or `indexing_complete` without a rollover alias would end up in an error state (per [this doc](https://www.elastic.co/docs/manage-data/lifecycle/index-lifecycle-management/skip-rollover#skip-rollover-manually)).

Anyway, the proposal I'll share for this `rehydrate` would be:

- Restore the index with no ILM association, and clean (`_ilm/remove`) the ILM state on the restored index (almost the same as in your KB).
- Then tell the user to apply the policy they want, recommending to use a policy without rollover as the nature of searchable snapshots data is data that was already rolled over, so there's no need to roll it over again.

I'll let you know when I have the final version so you can review. Thanks!

eedugon:
Making this a draft again until we merge https://github.com/elastic/docs-content/pull/8327
Then we will point to that document for the searchable snapshots part.

manage-data/lifecycle/data-tiers/manage-data-tiers-ech-ece.md
stefnestor: Yeah, sorry! In retrospect, I think I was held up more on "this allows Elasticsearch to begin" as a form of phrasing as performance-helpful but not process-necessary, which is why I was stuck on it's required or migrations won't happen. But 100% that's me shadowing-in how phrases like this get misinterpreted on Support cases 🙂.

PR #8239, opened 2026-09-08
Document ECE RHEL 10 host prep

gigerdo:
## Summary

ECE 4.2 already supports RHEL 10 + Podman 5 on the [support matrix](https://www.elastic.co/support/matrix#elastic-cloud-enterprise), but the host-prep docs still listed only RHEL 8/9. Following the RHEL 9 steps on RHEL 10 fails: `containernetworking-plugins` is not available, and ECE keeps Podman's default Netavark backend instead of switching to CNI.

Updates:

- [Configure a RHEL host](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/configure-host-rhel): add RHEL 10; Podman 5 only; keep Netavark
- [Install ECE](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/install): include RHEL 10 in the OS list
- [Migrate ECE to Podman hosts](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/migrate-ece-to-podman-hosts): same host-prep differences (also drops the stale RHEL `<9` restriction)
- [Set up IPv6 for ECE on AWS](https://www.elastic.co/docs/deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup): RHEL 10 AMI and do not pin `network_backend=cni`

Host-prep differences are based on the ECE Ansible role RHEL 10 path (`tasks/base/RedHat-10`).

## Generative AI disclosure
- [x] Yes
- [ ] No

Tool(s) and model(s) used: Cursor (Grok)

<small>*(AI-authored via Cursor 🤖)*</small>

### DIFF
--- a/deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
+++ b/deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
@@ -12,14 +12,16 @@ products:
 
 # Configure a RHEL host [ece-configure-hosts-rhel-centos]
 
-Red Hat Enterprise Linux 8 and 9, along with Rocky Linux 8 and 9, run {{ece}} (ECE) on Podman rather than Docker. Use these steps to install and configure Podman, and to apply the SELinux, XFS, and kernel tuning that ECE expects on RHEL-compatible distributions.
+Red Hat Enterprise Linux 8, 9, and 10, along with Rocky Linux 8 and 9, run {{ece}} (ECE) on Podman rather than Docker. Use these steps to install and configure Podman, and to apply the SELinux, XFS, and kernel tuning that ECE expects on RHEL-compatible distributions.
 
 * [Prerequisites](#ece-prerequisites-rhel8)
 * [Install Podman and configure the host](#ece-configure-hosts-rhel8-podman)
 
 ## Prerequisites [ece-prerequisites-rhel8]
 
-* Follow your internal guidelines to create a RHEL 8 (the version must be >= 8.5), RHEL 9, Rocky Linux 8, or Rocky Linux 9 server or VM in your environment.
+* Follow your internal guidelines to create a RHEL 8 (the version must be >= 8.5), RHEL 9, RHEL 10, Rocky Linux 8, or Rocky Linux 9 server or VM in your environment.
+
+    {applies_to}`ece: ga 4.2` RHEL 10 is supported with Podman 5 only. Do not install Podman 4 on RHEL 10, and do not follow the RHEL 9 CNI networking steps on RHEL 10.
 
 * Verify that required traffic is allowed. Check the [Networking prerequisites](ece-networking-prereq.md) for a list of ports that need to be open. The technical configuration depends on the underlying infrastructure. For example, for AWS, allowing traffic between hosts is implemented using security groups.
 
@@ -46,6 +48,7 @@ Red Hat Enterprise Linux 8 and 9, along with Rocky Linux 8 and 9, run {{ece}} (E
     sudo dnf -y install containernetworking-plugins
     ```
 
+    Do not install `containernetworking-plugins` on RHEL 10. That package is not available, and ECE on RHEL 10 uses Podman's default Netavark backend instead of CNI.
     ::::
 
 2. Remove Docker and previously installed Podman packages (if previously installed).
@@ -74,6 +77,11 @@ Red Hat Enterprise Linux 8 and 9, along with Rocky Linux 8 and 9, run {{ece}} (E
         ```
 
 4. Install Podman:
+
+    ::::{note}
+    :applies_to: ece: ga 4.2
+    RHEL 10 supports Podman 5 only. Use the Podman 5 steps in this section. Podman 4 is not a supported combination on RHEL 10.
+    ::::
 
     * For Podman 4:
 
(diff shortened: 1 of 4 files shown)


deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
```diff
@@ -12,14 +12,16 @@ products:
...
 
 * [Prerequisites](#ece-prerequisites-rhel8)
 * [Install Podman and configure the host](#ece-configure-hosts-rhel8-podman)
 
 ## Prerequisites [ece-prerequisites-rhel8]
 
-* Follow your internal guidelines to create a RHEL 8 (the version must be >= 8.5), RHEL 9, Rocky Linux 8, or Rocky Linux 9 server or VM in your environment.
+* Follow your internal guidelines to create a RHEL 8 (the version must be >= 8.5), RHEL 9, RHEL 10, Rocky Linux 8, or Rocky Linux 9 server or VM in your environment.
+
+    {applies_to}`ece: ga 4.2` RHEL 10 is supported with Podman 5 only. Do not install Podman 4 on RHEL 10, and do not follow the RHEL 9 CNI networking steps on RHEL 10.
```
eedugon, line 24: I'm proposing a small change on this line that you are adding. Considering how we have evolved from RHEL 8 to RHEL 10 and that we still support 2 major versions of Podman and 2 network backends, I think it's a good time to introduce a short summary that will feel visually simple, not only for RHEL 10.

```suggestion
    The following Podman versions and network backends are supported:
    
    * RHEL 8 and Rocky Linux 8: Podman 4 with CNI
    * RHEL 9: Podman 4 or Podman 5 with CNI
    * Rocky Linux 9: Podman 5 with CNI
    * {applies_to}`ece: ga 4.2` RHEL 10: Podman 5 with Netavark
```

The previous, together with the important note where we tell the users to ensure they use a supported combination of OS and Podman based on the support matrix should be enough here.

deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
```diff
@@ -46,6 +48,7 @@ Red Hat Enterprise Linux 8 and 9, along with Rocky Linux 8 and 9, run {{ece}} (E
     sudo dnf -y install containernetworking-plugins
     ```
 
+    Do not install `containernetworking-plugins` on RHEL 10. That package is not available, and ECE on RHEL 10 uses Podman's default Netavark backend instead of CNI.
```
eedugon, line 51: If the package is not available on RHEL 10, and if the note says explicitly that it's only for RHEL 9 and Rocky Linux 9 (I suspect it's probably needed for RHEL 8 but that's a different story), then I'd say this sentence doesn't add much value here. I'd suggest to remove it.

```suggestion
```

Later in the network backend configuration we tell the readers to keep the default backend.

deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
```diff
@@ -413,6 +428,8 @@ Red Hat Enterprise Linux 8 and 9, along with Rocky Linux 8 and 9, run {{ece}} (E
 
         On RHEL 9 and Rocky Linux 9, merge this setting with the existing `network_backend="cni"` configuration rather than creating a duplicate `[network]` section.
 
+        {applies_to}`ece: ga 4.2` On RHEL 10, do not add `network_backend`. Leave that setting unset so Podman continues to use Netavark.
```
eedugon, line 431: I think we can remove these two paragraphs. network_backend shouldn't be mentioned here in the dual-stack section.

```suggestion
```

The only message for any OS (RHEL 8,9 or 10) would be to not create a duplicate `network` section, and I'd say that's already covered in the step description.

Mentioning here again network_backend and then mentioning the RHEL 10 difference makes the reading flow unexpected in the dual-stack section.

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
```diff
@@ -624,7 +624,7 @@ To support IPv6 egress in an existing IPv4 ECE environment, you must update both
       ece-network
     ```
 
-    Then open `/etc/containers/containers.conf` and, in the `[network]` section, set `default_network`. If the file or section does not exist yet, create it. On RHEL 9 and Rocky Linux 9, merge this setting with the existing `network_backend="cni"` configuration rather than creating a duplicate `[network]` section:
+    Then open `/etc/containers/containers.conf` and, in the `[network]` section, set `default_network`. If the file or section does not exist yet, create it. On RHEL 9 and Rocky Linux 9, merge this setting with the existing `network_backend="cni"` configuration rather than creating a duplicate `[network]` section. {applies_to}`ece: ga 4.2` On RHEL 10, do not add `network_backend`. Leave that setting unset so Podman continues to use Netavark:
```
eedugon, line 627:
```suggestion
    Then open `/etc/containers/containers.conf` and, in the `[network]` section, set `default_network`. If the file or section does not exist yet, create it. Make sure not to create a duplicate `[network]` section if one already exists.
```

As before, I think the best solution is to not talk about the network backend in this specific doc, as it's not necessary.

deploy-manage/deploy/cloud-enterprise/migrate-ece-to-podman-hosts.md
```diff
@@ -153,20 +161,27 @@ Using Docker or Podman as container runtime is a configuration local to the host
...
             ```
 
-5. [This step is for RHEL 9 and Rocky Linux 9 only] Switch the network stack from Netavark to CNI:
+5. Configure the Podman network backend. The required backend depends on the OS version:
 
-    1. If the */etc/containers/containers.conf* file does not exist, copy the */usr/share/containers/containers.conf* file to the */etc/containers/* directory (for example, using `cp /usr/share/containers/containers.conf /etc/containers/`).
-    2. Open the */etc/containers/containers.conf* file. Navigate to the **network** section and make sure that the **network_backend** setting is set to `cni`.
-    3. Reboot the system (`reboot`).
-    4. Check that the network stack has changed to `cni`: <br>
+    * For RHEL 9 and Rocky Linux 9 only: Switch the network stack from Netavark to CNI.
```
eedugon, line 166: Ideally we should probably include a bullet for RHEL 8 / Rocky Linux 8.

Originally this step said for RHEL 9 only because on RHEL 9 we had to update the default backend to CNI, while in RHEL 8 I think the default backend is CNI already.

Now that we are converting the step to "configure network backend" and mentioning RHEL 9 and RHEL 10 in different bullets I'd say it's worthy a bullet for RHEL 8 saying something like:

> For RHEL 8, keep the default CNI backend.

Or even...

> For RHEL 8, ensure that Podman uses the CNI network backend. Although CNI is the default, configure `network_backend="cni"` explicitly to avoid depending on the existing host configuration.

That would be aligned with:

https://github.com/elastic/ansible-elastic-cloud-enterprise/blob/master/tasks/base/RedHat-8/configure_cni_network.yml

eedugon (commented):
Thanks for applying these changes @gigerdo !

Shared some comments for your consideration.

deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
gigerdo: Applied this summary as suggested. Same list is now on the migrate page so the two host-prep copies stay aligned.

<small>*(AI-authored via Cursor 🤖)*</small>

deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
gigerdo: Removed. The note is already scoped to RHEL 9 / Rocky Linux 9, and the RHEL 10 backend step covers keeping Netavark.

<small>*(AI-authored via Cursor 🤖)*</small>

deploy-manage/deploy/cloud-enterprise/configure-host-rhel.md
gigerdo: Removed both paragraphs. Dual-stack now only sets `default_network` and says not to create a duplicate `[network]` section.

<small>*(AI-authored via Cursor 🤖)*</small>

deploy-manage/deploy/cloud-enterprise/ece-ipv6-aws-setup.md
gigerdo: Applied. This page no longer mentions `network_backend`.

<small>*(AI-authored via Cursor 🤖)*</small>

deploy-manage/deploy/cloud-enterprise/migrate-ece-to-podman-hosts.md
gigerdo: Added a RHEL 8 / Rocky Linux 8 bullet that pins `network_backend="cni"` explicitly, matching the Ansible RHEL 8 task. Same steps are on the RHEL host-prep page.

<small>*(AI-authored via Cursor 🤖)*</small>

eedugon (approved):
LGTM! I think in some of the sections of the RHEL instructions we could use tabs for different OS versions instead of bullets, as that would make the doc visually nicer.

But as I'm planning to work with all installation guides (including Ubuntu & SuSE) soon, I can do that in a separate PR in the near future.

gigerdo:
> LGTM! I think in some of the sections of the RHEL instructions we could use tabs for different OS versions instead of bullets, as that would make the doc visually nicer.

Yes I think that makes sense :+1: we can keep iterating on this

PR #8243, opened 2026-09-08
On the EIS "Region and hosting" page, do not hyphenate 'Cloud Connect'; also, link it

acrewdson:
* 'Cloud Connect' is not hyphenated
* we can link to the 'Cloud Connect' docs


seanhandley (approved):

leemthompo (approved):
thanks @acrewdson !

PR #8251, opened 2026-09-09
Adding the new project type

yetanothertw:
## Summary

Updating the [Create a serverless project](https://www.elastic.co/docs/deploy-manage/deploy/elastic-cloud/create-serverless-project) page to include the new project type.

Related to https://github.com/elastic/docs-content-internal/issues/1780

## Details

I've tried to add a new line to the table, but it didn't look right. The padding around the Security logo looks off already and when the Vector Database one was added, that one looked even more off:
image

I've tried using the new table directive with smaller icons, and here's what I got. It still looks off as we can't control the padding in the cells.
image

Final option I settled on (stacked list with small icons):
image

I've also explored just adding `applies_to` tags instead of logos, but decided against it. Project types lose their immediately recognisable logos and imo just blend into one. 

## Generative AI disclosure

- [x] Yes  
- [ ] No


eedugon (approved):
I'm ok with the format change, although I feel this document is wrong in nature.

The user journey started with this doc never ends when they open any of the 3 links. Maybe we just need a sentence like, select your type of project, give it a name and click on "create". For more information about different projects and solutions, refer to.... (and the three links).

yetanothertw:
> I'm ok with the format change, although I feel this document is wrong in nature.
> 
> The user journey started with this doc never ends when they open any of the 3 links. Maybe we just need a sentence like, select your type of project, give it a name and click on "create". For more information about different projects and solutions, refer to.... (and the three links).

You're right, this page's problematic. I opened https://github.com/elastic/docs-content-internal/issues/1781 to figure out the best way to address this.

PR #8261, opened 2026-09-10
Fix tip formatting issue on ES vector page

florent-leborgne:
What the title says, otherwise the tip is actually the whole page :) 

AI disclosure: N/A - I could manage to do this without using AI 😌


leemthompo (approved):
That's a big tip!

PR #8289, opened 2026-09-14
docs(eis): split rate limit RPM column into search vs ingest

seanhandley:
## Summary

The previous rate limits page had a single table mixing chat, embedding, and reranking models. This didn't work well because different model types have fundamentally different limit dimensions — LLMs don't have token limits, rerankers have no ingest path, and only embedding models need separate search vs. ingest columns.

Changes:
- Split into three focused tables: **Chat**, **Embeddings**, **Rerankers**
- Embedding table has separate search and ingest columns for both RPM and TPM
- Corrects Jina v5 Nano and Small ingest RPM: 60,000 → **30,000** (sourced from `eis-gateway/config/default.yaml`)
- Removes duplicate Jina v5 rows (parenthetical name variants were duplicates)
- Chat table: RPM only, no token column (there is no token limit)
- Reranker table: single RPM and TPM (search-only, no ingest)

## Test plan

- [ ] Verify all three tables render correctly in docs preview
- [ ] Confirm Jina v5 search=60,000 and ingest=30,000 match gateway config
- [ ] Confirm section anchors (`[eis-rate-limits-chat]` etc.) are valid


leemthompo (approved):
👍

PR #8291, opened 2026-09-14
Document the API path for encrypting an existing deployment with a customer-managed key

### LINKED ISSUE #8236: Document the API path for encrypting an existing deployment with a customer-managed key
## Page

[Use a customer-managed encryption key](https://www.elastic.co/docs/deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key) (`deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key.md`)

## What's missing

The **Encrypt an existing deployment with your key** section documents only the Elastic Cloud console flow (deployment **Security** page > **Encryption at rest** > **Manage encryption key**). Encrypting an existing deployment is also supported through the Elastic Cloud API, but the API path is not documented anywhere on the page.

Today the `settings.byok.key_resource_path` example appears only under **Create a deployment encrypted with your key**, which reads as though the API supports setting a key at creation time only.

## Why it matters

Customers adopting BYOK on an existing footprint typically have many deployments, so the API is the path they will actually use rather than clicking through each deployment in the console. The console-only documentation understates what the feature can do for exactly the audience most likely to need it.

An upcoming Elastic blog post announcing GA of BYOK for existing deployments references the API path, so readers who follow the docs link will not find it covered.

## Suggested change

Add an API subsection under **Encrypt an existing deployment with your key**, mirroring the structure already used in **Create a deployment encrypted with your key**:

- the endpoint and method used to set a customer-managed key on an existing deployment
- the required API key permissions
- a worked `curl` example showing `settings.byok.key_resource_path`
- a note on how to track the resulting plan change

The existing note that a customer-managed key cannot be edited or removed once set should apply to the API path too, so it is worth restating or cross-linking there.

## Related

- Docs PR marking this capability GA: elastic/docs-content#8166
- Release note: elastic/cloud#158683

alxchalkias:
Closes #8236

## What this changes

Encrypting an existing deployment with a customer-managed key is documented for the Console only. This adds the API path, which is what customers rolling BYOK out across a fleet will actually use.

Additions to **Encrypt an existing deployment with your key**:

- The dedicated `POST /api/v1/deployments/{deployment_id}/byok-migration` endpoint, with a worked `curl` example. The existing Console steps are now labelled so the two paths sit side by side.
- The roles the endpoint accepts. These differ from create-time BYOK: **Editor** on the deployment is sufficient, which matters for anyone using scoped API keys and would otherwise over-provision to **Admin**.
- Its five error responses, including the pending-plan conflict and the ILM policy precondition.
- A new **Monitor re-encryption progress** subsection, leading with the Console **Overview** banner and covering the **Activity** page and the API equivalent.
- A **Use a highly available deployment** consideration.

A new Troubleshooting entry covers the conflict error, since Console users hit the pending-plan case too and it currently surfaces as an unexplained failure.

## Changes to existing text

Most of the diff is additions, but two things on the page were wrong and are corrected here.

**"This plan change happens without downtime"** is not supportable for every deployment shape, so it now reads that each instance is replaced with an encrypted one and the deployment stays accessible while re-encryption runs. Nothing in the migration path enforces or checks high availability, and `ByokMigrationSpec` exercises `zone_count = 1`, so non-HA deployments will run this. Re-encryption is a sequential grow-then-shrink, one instance at a time, so on a single-node deployment the sole master is replaced. `ElasticsearchClusterMigrateToByokPlanHandler.preStopInstancesPreAddInstances` documents that case directly: without the voting-config handling, "every subsequent HTTP call to the cluster fails with `master_not_discovered_exception`". The voting exclusions make that survivable rather than invisible. The replacement wording matches what the Console already tells users during the operation.

**High availability was absent from the page entirely**, while the Console flyout warns users to be "set up for high availability to prevent any risk of data loss" before they add a key. Added as a recommendation rather than a requirement, since nothing enforces it, and framed around both the data-loss risk and the single-AZ interruption above.

## Why the API shape looks different from the create-time example

Worth calling out for reviewers, because the obvious guess is wrong. Adding a key to an existing deployment is **not** `PUT /api/v1/deployments/{id}` with `settings.byok`. `DeploymentUpdateSettings` has no `byok` field, so that request is simply not the right path; `settings.byok` exists only on create. The operation has its own endpoint and a flat request body.

## Where the technical claims come from

Every technical assertion here was taken from source in `elastic/cloud` and `elastic/cloud-ui`, not from a published spec, so it deserves a technical review as well as a docs one:

| Claim | Source |
| --- | --- |
| Endpoint path and `update` permission | `DeploymentsByokMigrationRoutes.scala` |
| Request and response body | `ByokMigrationModels.scala`; `ByokMigrationRequest` / `ByokMigrationResponse` in cloud-ui `modules/cloud-api/v1/types.ts` |
| Error responses | `DeploymentsByokMigrationController.scala` |
| Role matrix | `DeploymentsByokMigrationRoutesAuthzSpec.scala` |
| `metadata.byok_migration_in_progress`, `settings.byok.key_resource_path` | `DeploymentGetResponse.scala`, `DeploymentToApiUtils.scala` |
| Full request-to-completion flow | `ByokMigrationSpec.scala` |
| Field naming in the public contract | cloud-ui `public/actions/byokMigration.ts` sends `key_resource_path` |
| Sequential grow-then-shrink, one instance at a time | `MasterAwareRollingGrowShrinkScheduler.scala`, `RollingGrowAndShrink.scala` |
| No HA or availability-zone precondition | `ByokMigrationValidations.scala`, `ValidatePlanPrerequisites.scala` |

One deliberate omission:

- Snapshot behavior for deployments **with** a cold or frozen tier. `ByokBucketMigrationSequence` preserves the old bucket read-only in `additional_*_resources` for those deployments, and no Elasticsearch repository is registered against it, so pre-migration snapshots are retained in storage but aren't reachable through `found-snapshots`. That is a retention promise either way, and there's an open question about searchable snapshot indices mounted outside ILM, so it's excluded here and will follow separately once confirmed.

## Related

`POST .../byok-migration` isn't published in the [Elastic Cloud API reference](https://www.elastic.co/docs/api/doc/cloud/group/endpoint-deployments) and `operation-migrate-deployment-byok` returns 404, although the operation does appear in backend OpenAPI generation and in cloud-ui swagger metadata. That's tracked separately against `elastic/cloud` and isn't a blocker here, but it's why this page carries a full `curl` example rather than linking to the operation.


alxchalkias:
Pushed a correction to the progress-tracking wording after verifying against a live migration, since this was the one claim in the original commit that rested on inference rather than source.

What the Console actually does:

- The deployment **Overview** page carries a dedicated banner: "This deployment is being re-encrypted with a customer-managed key (3/8 instances complete, 5/8 remaining). The time re-encryption takes depends on the size of your data and the stack version you're using." It includes a **View activity** link, and individual instances pick up an **Encrypted** marker as they finish.
- The **Activity** page shows the underlying plan change as a generic **Grow and shrink** entry with the usual plan steps, and nothing identifies it as a key migration.

My first pass sent readers to the Activity page, which is the weaker of the two screens: you would see a plan running with no indication it was the encryption. The section now leads with the Overview banner and keeps Activity for step-level detail, with an explicit note that the change isn't labeled there.

That was the last unsourced assertion in this PR. Everything else still traces to the files listed in the description.

eedugon:
@alxchalkias , this look great!

What do you think of adding this API to this list https://docs-v3-preview.elastic.dev/elastic/docs-content/pull/8291/deploy-manage/deploy/elastic-cloud/manage-deployments-using-elastic-cloud-api#ec_common_operations_with_dedicated_endpoints?

I think it would be a great addition to the list of "common operations with dedicated endpoints" in the doc where we talk about managing deployments with the API.

You can include a link to the procedure for the example and extra background.

alxchalkias:
@eedugon this operations should not be too "common" (it's mostly a one-off), but I don't mind the addition if you think it could be useful.

eedugon:
@alxchalkias : you are right on the "common" side, definitely.

The benefit would be to show that we have a dedicated API for this use case (BYOK) in the document where we consolidate the majority of knowledge about handling deployments with the API (besides the API docs).

But whatever you feel is better, no big deal from my side.

deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key.md
```diff
@@ -454,16 +454,72 @@ The deployment is now created and encrypted using the specified key. Future snap
 
 ### Encrypt an existing deployment with your key [ec_encrypt_an_existing_deployment_with_a_customer_managed_key]
 
+You can add a customer-managed key to a deployment that is already running, from the [{{ecloud}} Console](https://cloud.elastic.co?page=docs&placement=docs-body) or from the {{ecloud}} API.
```
eedugon, line 457:
```suggestion
You can add a customer-managed key to a running deployment from the [{{ecloud}} Console](https://cloud.elastic.co?page=docs&placement=docs-body) or the {{ecloud}} API.
```

alxchalkias:
@eedugon SGTM, let's add it.

deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key.md
```diff
@@ -454,16 +454,72 @@ The deployment is now created and encrypted using the specified key. Future snap
...
+
+**Before you begin**
+
+Check your index lifecycle management policies. If any ILM policy in the deployment uses the Elastic-managed `found-snapshots` repository for a `searchable_snapshot` action in its cold or frozen phase, encryption is rejected before it starts.
```
eedugon, line 461:
```suggestion
Check your index lifecycle management policies. If any ILM policy in the deployment uses the Elastic-managed `found-snapshots` repository for a `searchable_snapshot` action in its cold or frozen phase, the request to enable encryption is rejected.
```

deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key.md
```diff
@@ -454,16 +454,72 @@ The deployment is now created and encrypted using the specified key. Future snap
...
+
+    A successful request returns `"accepted": true`.
+
+If you send the same key again for a deployment that is already configured with that key, the request succeeds but no new plan change starts.
+
+The endpoint returns an error in the following cases:
+
+* `409` if the deployment is already encrypted with a different customer-managed key.
+* `409` if another plan is already pending on the deployment. Wait for that plan to finish, then retry.
+* `400` if the key can't be reached, or isn't valid for the deployment's cloud provider.
```
eedugon, line 498:
```suggestion
* `400` if the key can't be reached or isn't valid for the deployment's cloud provider.
```

deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key.md
```diff
@@ -454,16 +454,72 @@ The deployment is now created and encrypted using the specified key. Future snap
...
+* `400` with `byok_migration.ilm_policy_conflict` if an ILM policy uses the `found-snapshots` repository in a cold or frozen phase. Update those policies, then retry.
```
eedugon, line 499: > Update those policies, then retry.

When we say `update those policies` (here and in the introduction) what do we mean? Would it make sense to be more specific with something like this?

> Remove the `searchable_snapshot` action or configure it to use a custom snapshot repository, then retry.

And one extra question.... if a user does that, could they re-configure the ILM policies and use `found-snapshots` repository with searchable snapshots after having configured BYOK? Or is it something not supported from that moment onwards?

That would be better to clarify as from the doc we won't know if this is a temporary workaround / configuration for the BYOK migration to work and complete, or if it is something that has to be permanent.

deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key.md
```diff
@@ -454,16 +454,72 @@ The deployment is now created and encrypted using the specified key. Future snap
...
+
+From the API, get the deployment and check `metadata.byok_migration_in_progress`:
+
+```bash
+curl -XGET \
+-H "Authorization: ApiKey <replace with encoded API key>" \
+"https://api.elastic-cloud.com/api/v1/deployments/<replace with deployment ID>?show_plans=true&show_settings=true"
+```
+
+While encryption is running, `metadata.byok_migration_in_progress` is `true`. When it finishes, that field is `false` and `settings.byok.key_resource_path` holds your key identifier.
```
eedugon, line 521:
```suggestion
While re-encryption is running, `metadata.byok_migration_in_progress` is `true`. When it finishes, that field is `false` and `settings.byok.key_resource_path` holds your key identifier.
```

deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key.md
```diff
@@ -454,16 +454,72 @@ The deployment is now created and encrypted using the specified key. Future snap
...
+* `400` if the key can't be reached, or isn't valid for the deployment's cloud provider.
+* `400` with `byok_migration.ilm_policy_conflict` if an ILM policy uses the `found-snapshots` repository in a cold or frozen phase. Update those policies, then retry.
+
 {{ecloud}} then applies a plan change to encrypt your deployment's data and snapshots with your key. This plan change happens without downtime.
 
 ::::{note}
 Once you set a customer-managed key on a deployment, you cannot edit or remove it. Once encryption begins, you cannot undo it or switch to a different key. The ability to change or remove a customer-managed key will be supported on {{ech}} in the future.
 ::::
 
+#### Track the encryption progress
```
eedugon, line 507:
```suggestion
#### Monitor re-encryption progress
```

For your consideration!

eedugon (commented):
A few small comments for your consideration, all looks very good.

deploy-manage/security/encrypt-deployment-with-customer-managed-encryption-key.md
alxchalkias: @matt-elastic @UncertainCat can you help with answers here?

matt-elastic: @gigerdo Can you please comment on this topic?
I think there is no easy/practical workaround in this context that allows the customer to keep the data.

UncertainCat: "Update those policies" means removing the `searchable_snapshot` action from the cold and frozen phases, or pointing it at a repository the customer manages. Edu's wording works.

On switching back afterwards: nothing blocks re-adding `found-snapshots` after migration, and for new data it should work. For indices already mounted before migration, I suspect ILM would error, since they still reference snapshots in the old bucket.

gigerdo: Yes that sounds correct. (Though disclaimer I never actually went through those steps, if we want to be sure it works we should reproduce and test the remediation steps).

PR #8319, opened 2026-09-16
Update data views String field formatters options indent

dadoonet:
In 9.x the problem described in https://github.com/elastic/kibana/pull/181558 is still there. Less obvious but [here](https://www.elastic.co/docs/explore-analyze/find-and-organize/data-views#string-field-formatters).

Screenshot 2026-09-16 at 09-07-42 —


florent-leborgne:
@dadoonet Thanks for the suggestion. Believe it or not, after not touching to these docs for quite some time, I pushed a small refactor earlier today that reorganizes a little bit the data views docs. I think it already addresses your point: https://www.elastic.co/docs/explore-analyze/find-and-organize/data-views/field-formatters#string-transformations

Do you mind taking a look and letting me know?

dadoonet:
Lovely! I can close my PR then 😉

PR #8402, opened 2026-09-21
[Agent Builder] Split action-policy-management from rule-management

charlotte-hoblik:
Kibana [#281663](https://github.com/elastic/kibana/pull/281663) split `action-policy-management` out of the `rule-management` Agent Builder skill in 9.6. The skills reference still bundled them.

- Narrows the `rule-management` entry to rules, and scopes `manage_action_policy` with `removed 9.6` so 9.5 stays correct.
- Adds an `action-policy-management` entry.
- Fixes the same bundling in the inline tools list on the built-in tools reference.
- Adds the missing prerequisites lines. Both skills need `alerting:v2:enabled` and the Agent Builder experimental features setting.

A TODO comment tracks the third gate for `action-policy-management`, which lands with Kibana [#292162](https://github.com/elastic/kibana/pull/292162).

Closes https://github.com/elastic/docs-content-internal/issues/1612


nastasha-solomon (approved):
Changes to the alerting v2 docs look good - thanks!

PR #8436, opened 2026-09-21
Document SIGHUP reload of Kibana logging config (self-managed)

gsoldevila:
## Summary

Documents that self-managed _Kibana_ re-reads `kibana.yml` and re-applies its **logging** configuration when it receives a `SIGHUP` signal, **without a restart**. This is the reliable no-restart path for temporarily raising verbosity while troubleshooting (adjust a logger level or meta filters, reload, capture logs, revert, reload again).

Adds a "Reload the logging configuration without restarting" section to the self-managed **Configure Kibana** page, covering:

- The `kill -HUP <pid>` workflow and the Docker `docker kill --signal=HUP` variant.
- The multi-worker case (signal each process).
- How to confirm the reload (`Reloaded logging configuration due to SIGHUP` log line).
- The Windows limitation (`SIGHUP` unavailable → restart instead).

Scoped to logging on purpose: `SIGHUP` re-reads the whole `kibana.yml`, but logging is the configuration that Kibana deliberately re-applies live.

Relates to elastic/kibana#271682 (parent epic elastic/kibana#257240).


deploy-manage/deploy/self-managed/configure-kibana.md
```diff
@@ -29,6 +29,35 @@ In this file, you can also enable SSL and set a variety of other options.
 
 Environment variables can be injected into configuration using `${MY_ENV_VAR}` syntax. By default, configuration validation will fail if an environment variable used in the config file is not present when {{kib}} starts. This behavior can be changed by using a default value for the environment variable, using the `${MY_ENV_VAR:defaultValue}` syntax.
 
+## Reload the logging configuration without restarting [reload-logging-configuration]
+
+Most {{kib}} settings are read only at startup, so changing them requires a restart. The **logging** configuration is an exception: {{kib}} re-reads `kibana.yml` and re-applies its logging settings when the process receives a `SIGHUP` signal, without restarting.
+
+This is useful to temporarily raise verbosity while troubleshooting: adjust [`logging.root.level`](kibana://reference/configuration-reference/logging-settings.md), a specific [dedicated logger](/deploy-manage/monitor/logging-configuration/kib-advanced-logging.md), or [meta filters](kibana://reference/configuration-reference/logging-settings.md) in `kibana.yml`, reload, capture the logs you need, then revert and reload again.
```
eedugon, line 36:
```suggestion
## Reload configuration without restarting [reload-configuration]

Most {{kib}} settings are read only at startup, so changing them requires a restart. However, you can update some settings without restarting {{kib}} by sending a `SIGHUP` signal to the running process. The settings described in this section support this type of reload on Unix-like systems. On Windows, you must restart {{kib}} to apply changes made in `kibana.yml`.

:::{important}
If {{kib}} cannot parse the updated `kibana.yml` or validate the settings being reloaded, it shuts down. Check your changes before sending `SIGHUP`.
:::

### Reload logging settings [reload-logging-settings]

You can reload [{{kib}} logging settings](/deploy-manage/monitor/logging-configuration/kibana-logging.md) without restarting {{kib}}. This is useful when you need to temporarily increase logging verbosity while troubleshooting. You can change `logging.root.level`, configure a [dedicated logger](/deploy-manage/monitor/logging-configuration/kib-advanced-logging.md#dedicated-loggers), or increase verbosity for log records that match specific metadata. Reload the configuration and collect the detailed logs you need. When you finish, restore the original settings and reload the configuration.

For all available logging settings, refer to the [{{kib}} logging configuration reference](kibana://reference/configuration-reference/logging-settings.md).
```

deploy-manage/deploy/self-managed/configure-kibana.md
```diff
@@ -29,6 +29,35 @@ In this file, you can also enable SSL and set a variety of other options.
...
+
+To reload the logging configuration:
+
+1. Edit the `logging` settings in `kibana.yml`.
```
eedugon, line 40:
```suggestion
1. Update the `logging` settings in `kibana.yml` to configure the log levels, output, or filtering you need.
```

deploy-manage/deploy/self-managed/configure-kibana.md
```diff
@@ -29,6 +29,35 @@ In this file, you can also enable SSL and set a variety of other options.
...
+2. Send `SIGHUP` to the {{kib}} process:
```
eedugon, line 41:
```suggestion
2. Send a `SIGHUP` signal to the running {{kib}} process to reload the configuration:
```

deploy-manage/deploy/self-managed/configure-kibana.md
```diff
@@ -29,6 +29,35 @@ In this file, you can also enable SSL and set a variety of other options.
...
+
+    On [Docker](/deploy-manage/deploy/self-managed/install-kibana-with-docker.md), signal the container instead:
+
+    ```sh
+    docker kill --signal=HUP <container_id>
+    ```
+
+    If you run {{kib}} with multiple worker processes, send the signal to each process.
+
+3. Confirm the reload in the {{kib}} logs. You should see a message such as `Reloaded logging configuration due to SIGHUP`.
```
eedugon, line 55:
````suggestion
3. Confirm the reload in the {{kib}} logs. A successful reload produces a message similar to:

    ```text
    Reloaded Kibana configuration (reason: SIGHUP signal received).
    ```

4. Verify that the updated logging settings produce the expected output.

5. If the logging changes are temporary, restore the original settings and send another `SIGHUP` signal to reload the configuration.
````

File: deploy-manage/deploy/self-managed/configure-kibana.md [hunk 5]
```diff
@@ -29,6 +29,35 @@ In this file, you can also enable SSL and set a variety of other options.
...
+
+::::{note}
+`SIGHUP` is not available on Windows. On Windows, restart {{kib}} to apply logging configuration changes.
+::::
```
eedugon, line 59:
```suggestion
```

We can remove the note as now we are mentioning it in the intro.

eedugon (commented):
Thanks for sharing this @gsoldevila !

I left inline suggestions to make the section extensible to other settings that might support reload in the future, while keeping logging as the only documented case for now.

A few follow-ups:

- `server.ssl.*` appears to support dynamic TLS reload when TLS is already enabled ([implementation](https://github.com/elastic/kibana/blob/main/src/core/packages/http/server-internal/src/http_server.ts#L254-L275)). Is reloading these settings through `SIGHUP` intended to be a supported user-facing workflow? If so, it could be documented as another subsection. If not, we should keep the scope limited to logging today.

- Could we add a link to this section from the [Kibana logging](https://www.elastic.co/docs/deploy-manage/monitor/logging-configuration/kibana-logging) page in this PR? That is probably where users looking to change logging configuration will start, so telling them that logging settings can be reloaded without restart would be welcome.

- The Kibana configuration reference could also link to this section from the logging settings page, saying that they support reload (if not included already). Because that content lives in the Kibana repository, I’m happy to handle that in a follow-up PR if you like.

- I also suggested an `important` callout explaining that invalid YAML or an invalid logging configuration can cause Kibana to shut down during reload. Could you confirm that this accurately describes the expected behavior?

- I also updated the log message used to confirm a successful reload based on the current implementation. Could you confirm that it matches what {{kib}} emits?

shainaraskas:
this new section is very prominent. it should probably sit _below_ available settings, because that link is more important/generic

PR #8439, opened 2026-09-21
add multi-fields example to DLS/FLS docs

richard-dennehy:
## Summary
Some customers have reported confusion with how FLS rules apply to multi-fields. Add an example of how this should be configured, to clarify

## Generative AI disclosure
- [x] Yes  
- [ ] No  
2. If you answered "Yes" to the previous question, please specify the tool(s) and model(s) used (e.g., Google Gemini, OpenAI ChatGPT-4, etc.).

Tool(s) and model(s) used:
GPT 5.6 Sol Medium


deploy-manage/users-roles/cluster-or-deployment-auth/controlling-access-at-document-field-level.md
```diff
@@ -608,7 +608,42 @@ The resulting permission amounts to granted access to all `a.*` fields except th
 }
 ```
 
+### Configure access to multi-fields [field-level-security-multi-fields]
 
+To restrict access to a field and all of its [multi-fields](elasticsearch://reference/elasticsearch/mapping-reference/multi-fields.md), specify both the field and a wildcard that matches its multi-fields. For example, the following configurations restrict access to the `city` field and multi-fields such as `city.raw`, while granting read access to all other fields.
```
shainaraskas: We should say why this is needed.

also it's just one example configured through two surfaces


```suggestion
[Multi-fields](elasticsearch://reference/elasticsearch/mapping-reference/multi-fields.md) are stored as independent fields in the index, so excluding a parent field does not automatically exclude its multi-fields. To restrict access to a field and all of its multi-fields, specify both the field and a wildcard that matches its multi-fields. The following example restricts access to the `city` field and multi-fields such as `city.raw`, while granting read access to all other fields.
```

deploy-manage/users-roles/cluster-or-deployment-auth/controlling-access-at-document-field-level.md
```diff
@@ -608,7 +608,42 @@ The resulting permission amounts to granted access to all `a.*` fields except th
...
+1. In the **Index privileges** area, specify `*` as the data stream or index pattern and `read` as the privilege.
+1. Enable **Grant access to specific fields**.
+1. Add `*` to the **Granted fields** list.
+1. Add `city` and `city.*` as separate entries in the **Denied fields** list.
+
+::::
+:::::
 
 :::{note}
 Field-level security should not be set on [`alias`](elasticsearch://reference/elasticsearch/mapping-reference/field-alias.md) fields. To secure a concrete field, its field name must be used directly.
```
shainaraskas: now that this example has its own heading, we should float this note higher so it doesn't get lost. either above the heading, or closer to the opening of this section.

shainaraskas (approved):
this section is getting a little long, but your addition is a good one so I won't fuss with getting a perfect structure today. couple small comments, then good to go

deploy-manage/users-roles/cluster-or-deployment-auth/controlling-access-at-document-field-level.md
richard-dennehy: I've moved this after the first paragraph of the FLS section; as you say, it's buried otherwise

